PR #8, opened 2018-02-20
Dev/raquela/rebranding

raquelsa:
Re-branding: change from Library Installer to Library Manager


src/LibraryManager/Manifest.cs
```diff
@@ -228,15 +228,18 @@ private int DeleteLibraryFiles(ILibraryInstallationState state, Action<string> d
         {
             int filesDeleted = 0;
 
-            foreach (string file in state.Files)
+            if (state.Files != null)
```
alexgav: This is not a part of this PR, right? Also we need a bit more complicated fix here. Rather than doing nothing, we need to get full file list from the catalog. "null" here means "all files".

LibraryManager.sln
```diff
@@ -18,19 +18,19 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
...
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Web.LibraryInstaller.Vsix", "src\LibraryInstaller.Vsix\Microsoft.Web.LibraryInstaller.Vsix.csproj", "{EDA2179C-D952-449F-9A3D-8B3A152D6E5A}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Web.LibraryManager.Vsix", "src\LibraryManager.Vsix\Microsoft.Web.LibraryManager.Vsix.csproj", "{EDA2179C-D952-449F-9A3D-8B3A152D6E5A}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Src", "Src", "{A72570B1-99EF-4BDD-B629-321CD56C9F7F}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Test", "Test", "{FFCD12F4-5CE2-4CC2-A2C4-EACC8F387D7A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Web.LibraryInstaller.Build", "src\LibraryInstaller.Build\Microsoft.Web.LibraryInstaller.Build.csproj", "{7536DF26-B808-4AC2-A9DA-58E408A1E05B}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Web.LibraryManager.Build", "src\LibraryManager.Build\Microsoft.Web.LibraryManager.Build.csproj", "{7536DF26-B808-4AC2-A9DA-58E408A1E05B}"
```
alexgav: Why are we changing the project type guid?

raquelsa: It got changed during project rename. Not sure why that happened. I will revert...

jimmylewis: This change is from LegacyProjectSystemCSharp to ManagedProjectSystemCSharp, which seems appropriate since this is a netstandard project.  Not sure why it would have used the old GUID before...

src/LibraryManager/Manifest.cs
raquelsa: Good catch... I had it locally to avoid crashes while I was testing... :)

PR #11, opened 2018-03-01
Dev/raquela/disable commands

raquelsa:
Disabling commands during build and debug.

### DIFF
--- a/src/LibraryManager.Vsix/Commands/ManageLibrariesCommand.cs
+++ b/src/LibraryManager.Vsix/Commands/ManageLibrariesCommand.cs
@@ -32,6 +32,13 @@ namespace Microsoft.Web.LibraryManager.Vsix
             Instance = new ManageLibrariesCommand(package, commandService);
         }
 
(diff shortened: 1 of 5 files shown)


src/LibraryManager.Vsix/Commands/ManageLibrariesCommand.cs
```diff
@@ -32,6 +32,13 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
             Instance = new ManageLibrariesCommand(package, commandService);
         }
 
+        private void BeforeQueryStatus(object sender, EventArgs e)
+        {
+            var button = (OleMenuCommand)sender;
+
+            button.Visible = true;
```
alexgav: It's still going to be invisible if not in a web project though, right? It will call into QueryStatus only if visibility constrains are satisfied otherwise?

raquelsa: Yes, that's correct

PR #12, opened 2018-03-01
Adding localization files and targets. Using Multilingual App Toolkit…

alexgav:
… (MAT) for RESX file and custom tool VsctToXliff (sources added) for VSCT. XLF files are the loc files that vendors localized for us. We translated from RESX and VSCT to XLF, then vendors localize, then we tranlsate it back from localized XLF files to localized RESX and VSCT.


alexgav:
Yep, we will add vsixlangpack after I come back.

src/LibraryManager.Vsix/Commands/CommandTable/VSCommandTable.en.vsct
```diff
@@ -12,9 +12,7 @@
 
         <Buttons>
             <!-- Folder -->
-            <!-- Package Manager UI disabled.-->
-            <!-- To enable remove Condition="Defined(UI_ENABLED)" or define UI_ENABLED -->
-            <Button guid="guidLibraryManagerPackageCmdSet" id="InstallPackage" priority="0x0100" type="Button" Condition="Defined(UI_ENABLED)">
```
jodavis: Is this supposed to be removed as part of this change? Looks like Raquel added it last week to disable the UI, and this could re-enable it.

raquelsa: Yes, this will re-enable the UI. All VSCommandTable.*.vsct files need to be updated.

alexgav: Thanks guys, that wasn't intentional - looks like older copy of VSCT got used. I pushed a fix (the other seemingly million commits are my inept yet finally successful attempt to resolve merge conflict between my branch and master). You can look at the last commit only. Please let me know when you guys are done reviewing.

alexgav:
Thanks all!

PR #13, opened 2018-03-02
Fixing file lock bug on save. 

raquelsa:
We no longer read the manifest from disk on save but from the the document TextBuffer.

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -107,22 +107,27 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
         private void OnFileSaved(object sender, TextDocumentFileActionEventArgs e)
         {
-            if (e.FileActionType == FileActionTypes.ContentSavedToDisk)
+            var textDocument = (ITextDocument) sender;
+
+            if (e.FileActionType == FileActionTypes.ContentSavedToDisk && textDocument != null)
             {
                 Task.Run(async () =>
                 {
                     try
                     {
-                        Manifest newManifest = await Manifest.FromFileAsync(e.FilePath, _dependencies, CancellationToken.None).ConfigureAwait(false);
+                        Manifest newManifest = Manifest.FromJson(textDocument.TextBuffer.CurrentSnapshot.GetText(), _dependencies);
                         await RemoveFilesAsync(newManifest).ConfigureAwait(false);
 
                         _manifest = newManifest;
 
-                        await LibraryHelpers.RestoreAsync(e.FilePath, CancellationToken.None).ConfigureAwait(false);
+                        await LibraryHelpers.RestoreAsync(textDocument.FilePath, _manifest, CancellationToken.None).ConfigureAwait(false);
                         Telemetry.TrackOperation("restoresave");
                     }
                     catch (Exception ex)
                     {
+                        string textMessage = string.Concat(Environment.NewLine, LibraryManager.Resources.Text.RestoreHasErrors, Environment.NewLine);
+
+                        Logger.LogEvent(textMessage, LogLevel.Task);
                         Logger.LogEvent(ex.ToString(), LogLevel.Error);
                         Telemetry.TrackException("restoresavefailed", ex);
                     }
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -107,7 +107,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
         private void OnFileSaved(object sender, TextDocumentFileActionEventArgs e)
         {
-            var textDocument = (ITextDocument) sender;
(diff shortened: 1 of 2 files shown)


src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -107,22 +107,27 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
...
 
                         _manifest = newManifest;
 
-                        await LibraryHelpers.RestoreAsync(e.FilePath, CancellationToken.None).ConfigureAwait(false);
+                        await LibraryHelpers.RestoreAsync(textDocument.FilePath, _manifest, CancellationToken.None).ConfigureAwait(false);
                         Telemetry.TrackOperation("restoresave");
                     }
                     catch (Exception ex)
                     {
+                        string textMessage = string.Concat(Environment.NewLine, LibraryManager.Resources.Text.RestoreHasErrors, Environment.NewLine);
```
raquelsa: Adding logging to Output pane and status bar in case of failure. This is the same behavior we have for other restore operations invoked by UI.

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -48,13 +71,14 @@ public static bool IsSupported(this Project project)
             bool hasErrors = false;
             var telResult = new Dictionary<string, double>();
 
-            foreach (string configFilePath in configFilePaths)
+            foreach (var manifest in manifests)
```
jimmylewis: style: this should use the type instead of `var`.

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -35,10 +35,33 @@ public static bool IsSupported(this Project project)
...
+                var dependencies = Dependencies.FromConfigFile(configFilePath);
+                Manifest manifest = await Manifest.FromFileAsync(configFilePath, dependencies, cancellationToken).ConfigureAwait(false);
+
+                manifests.Add(configFilePath, manifest);
+            }
+
+            await RestoreAsync(manifests, cancellationToken).ConfigureAwait(false);
+        }
+
+        private static async Task RestoreAsync(IDictionary<string, Manifest> manifests,CancellationToken cancellationToken = default(CancellationToken))
```
jimmylewis: nit: missing a space after `manifests,`

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -35,10 +35,33 @@ public static bool IsSupported(this Project project)
...
+            await RestoreAsync(new Dictionary<string, Manifest>() { [configFilePath] = manifest } , cancellationToken).ConfigureAwait(false);
         }
 
         public static async Task RestoreAsync(IEnumerable<string> configFilePaths, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            Dictionary<string, Manifest> manifests = new Dictionary<string, Manifest>();
+
+            foreach (string configFilePath in configFilePaths)
+            {
+                var dependencies = Dependencies.FromConfigFile(configFilePath);
```
jimmylewis: style: `var` should be changed to type.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -107,22 +107,27 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
 
         private void OnFileSaved(object sender, TextDocumentFileActionEventArgs e)
         {
-            if (e.FileActionType == FileActionTypes.ContentSavedToDisk)
+            var textDocument = (ITextDocument) sender;
```
jimmylewis: if sender isn't `ITextDocument` this will throw an InvalidCastException.  Did you want `sender as ITextDocument`?

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -48,13 +71,14 @@ public static bool IsSupported(this Project project)
             bool hasErrors = false;
             var telResult = new Dictionary<string, double>();
 
-            foreach (string configFilePath in configFilePaths)
+            foreach (KeyValuePair<string, Manifest> manifest in manifests)
```
alexgav: If we end up doing this in more places, we should probably make path a non-serialized field of Manifest.

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -35,7 +35,7 @@ public static bool IsSupported(this Project project)
 
         public static async Task RestoreAsync(string configFilePath, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var dependencies = Dependencies.FromConfigFile(configFilePath);
+            Dependencies dependencies = Dependencies.FromConfigFile(configFilePath);
```
alexgav: +1 on not using the var in LibMan codebase (I know we have a million of them).

PR #14, opened 2018-03-03
Fixing bug 568284 (nullref on Clean command).

alexgav:
It's a bit ugly but I think it's OK for preview 3. Not sure if we should propagate the update failure out and fail the entire Clean operation (may have cleaned some but not all libraries - success or failure?).

We should test the FileSystem provider though as well. The code says Files should be required, not sure if it's enforced.

### DIFF
--- a/src/LibraryManager.Contracts/IProvider.cs
+++ b/src/LibraryManager.Contracts/IProvider.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Web.LibraryManager.Contracts
 
         /// <summary>
         /// Installs a library as specified in the <paramref name="desiredState"/> parameter.
-        /// </summary>  
+        /// </summary>
         /// <param name="desiredState">The details about the library to install.</param>
         /// <param name="cancellationToken">A token that allows for the operation to be cancelled.</param>
         /// <returns>The <see cref="ILibraryInstallationResult"/> from the installation process.</returns>
--- a/src/LibraryManager.Contracts/IProvider.cs
+++ b/src/LibraryManager.Contracts/IProvider.cs
@@ -31,11 +31,19 @@ namespace Microsoft.Web.LibraryManager.Contracts
 
         /// <summary>
         /// Installs a library as specified in the <paramref name="desiredState"/> parameter.
-        /// </summary>
+        /// </summary>  
         /// <param name="desiredState">The details about the library to install.</param>
         /// <param name="cancellationToken">A token that allows for the operation to be cancelled.</param>
         /// <returns>The <see cref="ILibraryInstallationResult"/> from the installation process.</returns>
         Task<ILibraryInstallationResult> InstallAsync(ILibraryInstallationState desiredState, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Updates library state using catalog if needed
+        /// </summary>
+        /// <param name="desiredState"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task<ILibraryInstallationResult> UpdateStateAsync(ILibraryInstallationState desiredState, CancellationToken cancellationToken);
 
         /// <summary>
         /// Gets the <see cref="ILibraryCatalog"/> for the <see cref="IProvider"/>. May be <code>null</code> if no catalog is supported.
--- a/src/LibraryManager.Vsix/ErrorList/ErrorList.cs
+++ b/src/LibraryManager.Vsix/ErrorList/ErrorList.cs
@@ -1,10 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.Web.LibraryManager.Contracts;
 using System.Collections.Generic;
-using System.Linq;
 using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.Telemetry;
+using Microsoft.Web.LibraryManager.Contracts;
 
 namespace Microsoft.Web.LibraryManager.Vsix
 {
@@ -37,7 +38,7 @@ namespace Microsoft.Web.LibraryManager.Vsix
                     foreach (IError error in result.Errors)
                     {
                         Logger.LogEvent(error.Message, LogLevel.Operation);
-                        Telemetry.TrackOperation("error", new KeyValuePair<string, object>("code", error.Code));
+                        Telemetry.TrackOperation("error", TelemetryResult.Failure, new KeyValuePair<string, object>("code", error.Code));
                     }
                 }
             }
--- a/src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
+++ b/src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
@@ -128,12 +128,20 @@ namespace Microsoft.Web.LibraryManager.Vsix
             Manifest manifest = await Manifest.FromFileAsync(configFileName, dependencies, CancellationToken.None).ConfigureAwait(false);
             var hostInteraction = dependencies.GetHostInteractions() as HostInteraction;
 
-            int? filesDeleted = manifest?.Clean((file) => hostInteraction.DeleteFiles(file));
+            IEnumerable<ILibraryInstallationResult> results = manifest?.Clean((file) => hostInteraction.DeleteFiles(file));
 
-            Logger.LogEvent(Resources.Text.CleanLibrariesSucceeded + Environment.NewLine, LogLevel.Task);
+            if (results != null && results.All(r => r.Success))
+            {
+                Logger.LogEvent(Resources.Text.CleanLibrariesSucceeded + Environment.NewLine, LogLevel.Task);
+            }
+            else
+            {
+                Logger.LogEvent(Resources.Text.CleanLibrariesFailed + Environment.NewLine, LogLevel.Task);
+            }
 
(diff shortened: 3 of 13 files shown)


src/LibraryManager.Contracts/IProvider.cs
```diff
@@ -31,12 +31,20 @@ public interface IProvider
 
         /// <summary>
         /// Installs a library as specified in the <paramref name="desiredState"/> parameter.
-        /// </summary>
+        /// </summary>
```
raquelsa: nit: remove extra space

src/LibraryManager/Manifest.cs
```diff
@@ -218,7 +219,7 @@ public int Clean(Action<string> deleteFileAction)
 
             foreach (ILibraryInstallationState state in Libraries)
             {
-                filesDeleted += DeleteLibraryFiles(state, deleteFileAction);
+                    filesDeleted += DeleteLibraryFiles(state, deleteFileAction);
```
raquelsa: nit: remove extra space

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -139,6 +122,56 @@ public ILibraryCatalog GetCatalog()
...
+
+                desiredState = new LibraryInstallationState
+                {
+                    ProviderId = Id,
+                    LibraryId = desiredState.LibraryId,
+                    DestinationPath = desiredState.DestinationPath,
+                    Files = library.Files.Keys.ToList(),
+                };
+            }
+            catch (Exception ex) when (ex is InvalidLibraryException || ex.InnerException is InvalidLibraryException)
```
raquelsa: Should we remove the try/catch for (InvalidLibraryException || ex.InnerException is InvalidLibraryException)  for the callers of UpdateStateAsync ?

src/LibraryManager/Manifest.cs
```diff
@@ -228,16 +229,34 @@ private int DeleteLibraryFiles(ILibraryInstallationState state, Action<string> d
...
+                    {
+                        string relativePath = Path.Combine(state.DestinationPath, file).Replace('\\', '/');
+                        deleteFileAction?.Invoke(relativePath);
+                        filesDeleted++;
+                    }
+                }
+            }
+            else
+            {
+                _hostInteraction.Logger.Log(string.Format(Resources.Text.LibraryInfoUpdateFailed, state.LibraryId), LogLevel.Error);
```
raquelsa: Instead of logging the errors here, we could return ILibraryInstallationResult updatedStateResult. I log it at the callers level. The updatedStateResult could have errors for different reasons other than  "Resources.Text.LibraryInfoUpdateFailed".

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -79,30 +79,13 @@ public ILibraryCatalog GetCatalog()
...
-                    {
-                        ProviderId = Id,
-                        LibraryId = desiredState.LibraryId,
-                        DestinationPath = desiredState.DestinationPath,
-                        Files = library.Files.Keys.ToList(),
-                    };
+                    return result;
                 }
 
                 foreach (string file in desiredState.Files)
```
raquelsa: This should be foreach (string file in result.InstallationState.Files).

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -79,30 +79,13 @@ public ILibraryCatalog GetCatalog()
...
-                {
-                    desiredState = new LibraryInstallationState
-                    {
-                        ProviderId = Id,
-                        LibraryId = desiredState.LibraryId,
-                        DestinationPath = desiredState.DestinationPath,
-                        Files = library.Files.Keys.ToList(),
-                    };
+                    return result;
                 }
```
raquelsa: We are missing: desiredState = result.InstallationState;

src/LibraryManager/Manifest.cs
```diff
@@ -228,16 +229,34 @@ private int DeleteLibraryFiles(ILibraryInstallationState state, Action<string> d
         {
             int filesDeleted = 0;
 
-            foreach (string file in state.Files)
+            IProvider provider = _dependencies.GetProvider(state.ProviderId);
+            ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, new CancellationToken()).Result;
```
jimmylewis: `new CancellationToken()` is not great.  If you want to be able to cancel the operation, you need a `CancellationTokenSource`.  If you don't, then you should use `CancellationToken.None` to indicate the intent.

src/LibraryManager/Manifest.cs
```diff
@@ -228,16 +229,34 @@ private int DeleteLibraryFiles(ILibraryInstallationState state, Action<string> d
...
+                foreach (IError error in updatedStateResult.Errors)
                 {
-                    string relativePath = Path.Combine(state.DestinationPath, file).Replace('\\', '/');
-                    deleteFileAction?.Invoke(relativePath);
-                    filesDeleted++;
+                    // TODO: {alexgav} The logger should have a method to log an IError
+                    _hostInteraction.Logger.Log(error.Message, LogLevel.Error);
                 }
+
+                // TODO: {alexgav} Should we fail the entire Clean operation?
```
jimmylewis: IMO we should indicate failure if Clean fails.  The project is in a bad state.

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -99,6 +99,17 @@ public ILibraryCatalog GetCatalog()
             return LibraryInstallationResult.FromSuccess(desiredState);
         }
 
+        /// <summary>
+        /// No-op for FileSystemProvider
+        /// </summary>
+        /// <param name="desiredState"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public Task<ILibraryInstallationResult> UpdateStateAsync(ILibraryInstallationState desiredState, CancellationToken cancellationToken)
```
raquelsa: We need to implement this for filesystem as well. Restore will fail if we don't specify the files collection as well. Null ref exception. We need to be able to populate the Files collection based on the "library" specified. 
If path is to a file, add that file to the Files collection. If path is to a folder, add all the files from that folder.

raquelsa: There is another bug tracking this work item

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -128,12 +128,20 @@ public static bool IsSupported(this Project project)
...
+                Logger.LogEvent(Resources.Text.CleanLibrariesSucceeded + Environment.NewLine, LogLevel.Task);
+            }
+            else
+            {
+                Logger.LogEvent(Resources.Text.CleanLibrariesFailed + Environment.NewLine, LogLevel.Task);
+            }
 
-            TelemetryResult result = configProjectItem != null ? TelemetryResult.Success : TelemetryResult.Failure;
-            Telemetry.TrackUserTask("clean", new KeyValuePair<string, object>("filesdeleted", filesDeleted));
+            // Do we need this? We could log the individual errors for each library instead
```
raquelsa: Design question: Please comment!

src/LibraryManager/Manifest.cs
```diff
@@ -245,21 +247,14 @@ private int DeleteLibraryFiles(ILibraryInstallationState state, Action<string> d
                         filesDeleted++;
                     }
                 }
-            }
-            else
-            {
-                _hostInteraction.Logger.Log(string.Format(Resources.Text.LibraryInfoUpdateFailed, state.LibraryId), LogLevel.Error);
-                foreach (IError error in updatedStateResult.Errors)
+
+                if (filesDeleted == state.Files.Count())
```
raquelsa: Check if Files can be null here after calling UpdateStateAsync

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
jimmylewis: We should keep the telemetry here.  We also need to make sure that the `TelemetryResult` is getting hooked up correctly.  It looks like `TrackUserTask` needs an overload to accommodate the result.

jimmylewis: Can we also include error codes encountered in the failing cases?

src/LibraryManager.Vsix/ErrorList/ErrorList.cs
```diff
@@ -37,7 +38,7 @@ public bool HandleErrors(IEnumerable<ILibraryInstallationResult> results)
                     foreach (IError error in result.Errors)
                     {
                         Logger.LogEvent(error.Message, LogLevel.Operation);
-                        Telemetry.TrackOperation("error", new KeyValuePair<string, object>("code", error.Code));
+                        Telemetry.TrackOperation("error", TelemetryResult.Failure, new KeyValuePair<string, object>("code", error.Code));
```
jimmylewis: Not sure if the operation name "error" is best, as it doesn't say which operation failed.  Can we know which operation it was?

raquelsa: Yes, I agree with you but did not want to change what was in place without understanding more about how we should be logging telemetry events in general for this feature. I will log a bug to review Telemetry in general for LibMan.

PR #15, opened 2018-03-03
Fix for bug 568371 (changes in library version are ignored)

alexgav:
Code wans't using library version number when doing comparison of files from the old and new libraries, only relative installation path (which has no version info in it). Adding version info temporarily for comparison (then removing prior to installation). A bit dirty. Please comment if you think we want a cleaner fix (use a class / tuple with file path and version number as fields rather than just a path?)

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -61,9 +61,9 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
         private async Task RemoveFilesAsync(Manifest newManifest)
         {
-            IEnumerable<string> prevFiles = await GetAllManifestFilesAsync(_manifest).ConfigureAwait(false);
-            IEnumerable<string> newFiles = await GetAllManifestFilesAsync(newManifest).ConfigureAwait(false);
-            IEnumerable<string> filesToRemove = prevFiles.Where(f => !newFiles.Contains(f));
+            IEnumerable<string> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
+            IEnumerable<string> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


jimmylewis:
I'm in favor of a tuple/type to store the information separately.  Also, does the FileSystem provider have version information?  Can it determine when the files need to be updated?

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -61,9 +61,9 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
 
         private async Task RemoveFilesAsync(Manifest newManifest)
         {
-            IEnumerable<string> prevFiles = await GetAllManifestFilesAsync(_manifest).ConfigureAwait(false);
-            IEnumerable<string> newFiles = await GetAllManifestFilesAsync(newManifest).ConfigureAwait(false);
-            IEnumerable<string> filesToRemove = prevFiles.Where(f => !newFiles.Contains(f));
+            IEnumerable<string> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
+            IEnumerable<string> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest).ConfigureAwait(false);
+            IEnumerable<string> filesToRemove = prevFiles.Where(f => !newFiles.Contains(f)).Select(f => f.Substring(0, f.LastIndexOf(Path.DirectorySeparatorChar)));
```
raquelsa: This can throw System.ArgumentOutOfRangeException if LastIndex returns -1. Maybe the tuple option Jimmy suggested could be a better option here.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -89,7 +98,7 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
             return files;
         }
 
-        private async Task<IEnumerable<string>> GetFilesAsync(ILibraryInstallationState state)
+        private async Task<IEnumerable<string>> GetFilesWithVersionsAsync(ILibraryInstallationState state)
         {
             if (state.Files == null)
```
raquelsa: If state has files (the case where you specify the files you want from a library in the manifest), then we won't append the version causing failures.

PR #16, opened 2018-03-03
Fixing multiple output windows bug

alexgav:
… was already static. However, all of the web editors code checks for pane existence prior to creating one. Not sure how it can exist if our static is null, but it appears to be possible. So checking for existence before creating seems to fix that.

I still went ahead and added (ProjectName) suffix to messages. However, the current logging design is kinf of iffy. Existing Log was a essentially a static class, but wasn't declared as such because Mads wanted it to implement ILogger interface. So we were actually creating multiple instances of that essentially static class... Really weird. Anyway, I created PerProjectLogger that has instance data now, and converted Logger into the actual static class it was really supposed to be. So PerProjectLogger suffixes messages with project name, while static messages of the Logger don't... It doesn't look horrible, but a bit weird. We can discuss the design more. Below is example output restoring jquery in two different projects in the same solution.

Restoring jquery@3.3.0... (WebApplication1)

1 libraries restored in 0.22 seconds

Restoring jquery@3.3.0... (WebApplication3)

1 libraries restored in 0.07 seconds

### DIFF
--- a/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
+++ b/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnvDTE;
 using Microsoft.Web.LibraryManager.Contracts;
+using Microsoft.Web.LibraryManager.Vsix.Contracts;
 
 namespace Microsoft.Web.LibraryManager.Vsix
 {
@@ -16,11 +17,12 @@ namespace Microsoft.Web.LibraryManager.Vsix
         {
             string cwd = Path.GetDirectoryName(configFilePath);
             WorkingDirectory = cwd;
+            Logger = new PerProjectLogger(VsHelpers.GetProjectName(configFilePath));
         }
 
         public string WorkingDirectory { get; }
         public string CacheDirectory => Constants.CacheFolder;
-        public ILogger Logger { get; } = new Logger();
+        public ILogger Logger { get; } 
 
         public async Task<bool> WriteFileAsync(string path, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken)
         {
--- a/src/LibraryManager.Vsix/Contracts/Logger.cs
+++ b/src/LibraryManager.Vsix/Contracts/Logger.cs
@@ -4,17 +4,20 @@
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.VisualStudio.Shell.Interop;
 using System;
+using Microsoft.VisualStudio;
 
 namespace Microsoft.Web.LibraryManager.Vsix
 {
-    public class Logger : ILogger
+    public static class Logger
     {
+        private static Guid _outputPaneGuid = new Guid("cce35aef-ace6-4371-b1e1-8efa3cdc8324");
         private static IVsOutputWindowPane _pane;
         private static readonly IVsOutputWindow _output = VsHelpers.GetService<SVsOutputWindow, IVsOutputWindow>();
         private static readonly IVsActivityLog _activityLog = VsHelpers.GetService<SVsActivityLog, IVsActivityLog>();
         private static readonly IVsStatusbar _statusbar = VsHelpers.GetService<SVsStatusbar, IVsStatusbar>();
 
-        public void Log(string message, LogLevel level)
+
+        public static void Log(string message, LogLevel level)
         {
             LogEvent(message, level);
         }
@@ -75,9 +78,17 @@ namespace Microsoft.Web.LibraryManager.Vsix
         {
             if (_pane == null)
             {
-                var guid = Guid.NewGuid();
-                _output.CreatePane(ref guid, Vsix.Name, 1, 1);
-                _output.GetPane(ref guid, out _pane);
+                if (_output != null)
+                {
+                    if (ErrorHandler.Failed(_output.GetPane(ref _outputPaneGuid, out _pane)) &&
+                        ErrorHandler.Succeeded(_output.CreatePane(ref _outputPaneGuid, Vsix.Name, 0, 0)))
+                    {
+                        if (ErrorHandler.Succeeded(_output.GetPane(ref _outputPaneGuid, out _pane)))
+                        {
+                            _pane.Activate();
+                        }
+                    }
+                }
             }
 
             return _pane != null;
--- a/src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
+++ b/src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Web.LibraryManager.Contracts;
+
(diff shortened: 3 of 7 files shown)


src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+        private string _projectName;
+
+        public PerProjectLogger(string projectName)
+        {
+            _projectName = projectName;
+        }
+
+        public void Log(string message, LogLevel level)
+        {
+            Logger.Log($"{message} ({_projectName})", level);
```
raquelsa: _projectName can be null here if we fail to get projectName

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -3,6 +3,7 @@
 
 using EnvDTE;
```
jimmylewis: nit: usings aren't sorted correctly (System first)

src/LibraryManager.Vsix/Contracts/Logger.cs
```diff
@@ -4,17 +4,20 @@
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.VisualStudio.Shell.Interop;
 using System;
+using Microsoft.VisualStudio;
 
 namespace Microsoft.Web.LibraryManager.Vsix
 {
-    public class Logger : ILogger
+    public static class Logger
```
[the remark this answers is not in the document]
jimmylewis: It also feels weird that `Logger` doesn't implement `ILogger`, but another type does and all it does is call methods on `Logger`.  That seems like it ought to be using an inherited class instead.

src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Web.LibraryManager.Contracts;
+
+namespace Microsoft.Web.LibraryManager.Vsix.Contracts
+{
+    internal class PerProjectLogger : ILogger
```
jimmylewis: As commented on `Logger`, it feels like this ought to be extending the `Logger` class (before it was static).  `HostInteraction` would get an instance of this type through DI.

src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
jimmylewis: Should we have a fallback string for a default if we can't resolve a project name?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -236,6 +237,60 @@ public static Project GetDTEProject(IVsHierarchy hierarchy)
...
+        public static bool TryGetHierarchy(string filePath, out IVsHierarchy vsHierarchy, out uint vsItemId)
+        {
+            bool result = true;
+
+            vsHierarchy = null;
+            vsItemId = (uint)VSConstants.VSITEMID.Nil;
+
+            IVsUIShellOpenDocument vsUIShellOpenDocument = ServiceProvider.GlobalProvider.GetService(typeof(SVsUIShellOpenDocument)) as IVsUIShellOpenDocument;
+
+            IOleServiceProvider serviceProviderUnused = null;
```
jimmylewis: nit: you don't need this, you can just use `out _` in the method call and it will be discarded.

jimmylewis: https://docs.microsoft.com/en-us/dotnet/csharp/discards

jimmylewis: same with `docInProject`

src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
alexgav: Not sure what the default project name would be. I think if it's null, we should just omit it. Anything else would look bad to the users. I'm not sure in which case project name would be null though - do you guys know of such cases? 

Anyway - I added null check and will be using empty string for now if there is no project name - so the output would be as before this change.

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -16,11 +17,12 @@ public HostInteraction(string configFilePath)
         {
             string cwd = Path.GetDirectoryName(configFilePath);
             WorkingDirectory = cwd;
+            Logger = new PerProjectLogger(VsHelpers.GetProjectName(configFilePath));
```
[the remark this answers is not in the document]
alexgav: Sure, taking it as a parameter now and added internal setter as well.

src/LibraryManager.Vsix/Contracts/Logger.cs
```diff
@@ -4,17 +4,20 @@
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.VisualStudio.Shell.Interop;
 using System;
+using Microsoft.VisualStudio;
```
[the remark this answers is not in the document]
alexgav: Cleaned up.

src/LibraryManager.Vsix/Contracts/Logger.cs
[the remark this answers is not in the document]
alexgav: I don't think inheritance between PerProjectLogger and Logger would be appropriate here as "Logger" is essentially a VS Log Sink, just a wrapper around Status Bar & OutputWindow, and there will always be just one of them. PerProjectLogger gets created by the project, use by external "clients", and sends output to the VS Log Sink. So I think the composition is more appropriate here than inheritance.

As far as static objects vs singletons and making things testable, I totally agree in general. In this particular case, Logger is simply a very thin wrapper around VS Output Window and Status Bar, so I don't think it makes sense to test it via unit tests. I'll leave it as static for now, and we can chat more about it if you want.

src/LibraryManager.Vsix/Contracts/Logger.cs
```diff
@@ -4,17 +4,20 @@
...
 
 namespace Microsoft.Web.LibraryManager.Vsix
 {
-    public class Logger : ILogger
+    public static class Logger
     {
+        private static Guid _outputPaneGuid = new Guid("cce35aef-ace6-4371-b1e1-8efa3cdc8324");
```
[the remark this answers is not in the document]
alexgav: Fixed

src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
alexgav: See my comments above. We can chat more about it in person.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
alexgav: Nice, didn't know about these, thanks!

src/LibraryManager.Vsix/Contracts/Logger.cs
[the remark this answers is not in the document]
alexgav: Just kidding, can't be, passed by ref to a few methods below (even though they don't change it).

PR #19, opened 2018-03-08
Fixed Czech translations to make sense.

ridercz:


raquelsa:
Thanks for the contribution. The strings were reviewing and updated on #39 . Let us know if you still see any translation issues.

PR #21, opened 2018-03-09
Renaming the .vsix.

raquelsa:


src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -25,6 +25,7 @@
     <AssemblyName>Microsoft.Web.LibraryManager.Vsix</AssemblyName>
     <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <GeneratePkgDefFile>true</GeneratePkgDefFile>
+    <TargetVsixContainerName>Microsoft.Web.LibraryManager.Vsix</TargetVsixContainerName>
```
jimmylewis: nit: the file extension should be lower case for consistency

raquelsa: Missed that...

joeloff:
I can't leave comments in the files, so here goes:

<ExtensionInstallationFolder>Microsoft\Web Tools Azure\WebJobs</ExtensionInstallationFolder> This should be a folder for the Library installer. This folder you have is used by WTA extensions, it was just an example.

Also, just be aware that once IsProductComponent is set, people can't build and install the VSIX locally, because it will expect to have a matching component wrapper in the VS setup manifest - this won't happen until we do the actually insertion and ship VS

joeloff:
For the versioning <Identity Id="Microsoft.Web.LibraryManager" Version="15.0.0" Language="en-US" Publisher="Microsoft" /> Do you want to manually change it or do it automatically?

You can inject properties, assuming you have a way to generate a build number on the fly. I can share some examples.

appveyor.yml
```diff
@@ -18,4 +18,7 @@ artifacts:
     name: NuGet package
 
   - path: src\**\*.vsix
-    name: Visual Studio Extension
\ No newline at end of file
+    name: Visual Studio Extension
+
+    - path: src\**\Microsoft.Web.LibraryManager.json
+    name: Visual Studio Extension manifest
```
jimmylewis: nit: always put a newline at the end of the file.

PR #22, opened 2018-03-13
Addressing code review feedback

raquelsa:
- Using a Tuple<string, string> for <filepath, version> instead of a single path with the version appended at the end of path.

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -61,9 +61,9 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
         private async Task RemoveFilesAsync(Manifest newManifest)
         {
-            IEnumerable<string> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
-            IEnumerable<string> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest).ConfigureAwait(false);
-            IEnumerable<string> filesToRemove = prevFiles.Where(f => !newFiles.Contains(f)).Select(f => f.Substring(0, f.LastIndexOf(Path.DirectorySeparatorChar)));
+            IEnumerable<Tuple<string, string>> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
+            IEnumerable<Tuple<string, string>> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -61,9 +61,9 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
 
         private async Task RemoveFilesAsync(Manifest newManifest)
         {
-            IEnumerable<string> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
-            IEnumerable<string> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest).ConfigureAwait(false);
-            IEnumerable<string> filesToRemove = prevFiles.Where(f => !newFiles.Contains(f)).Select(f => f.Substring(0, f.LastIndexOf(Path.DirectorySeparatorChar)));
+            IEnumerable<Tuple<string, string>> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
```
jimmylewis: Use ValueTuples instead of Tuple.  It's so much more readable.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -79,39 +79,37 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
...
-                    return library?.Files.Keys.Select((f) => Path.Combine(f, library.Version)).ToList();
-                }
+            if (catalog != null)
+            {
+                ILibrary library = await catalog?.GetLibraryAsync(state.LibraryId, CancellationToken.None);
+                return library?.Files.Select(f => new Tuple<string, string>(Path.Combine(state.DestinationPath, f.Key), library.Version)).ToList();
             }
 
-            return state.Files.Distinct();
+            return null;
```
jimmylewis: I hate returning null, can we return an empty collection instead?

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
jimmylewis: (This goes for every other Tuple in this PR)

raquelsa: Ok, never used it.. Let me read about it

raquelsa: Not available on .NET Framework 4.6 and netstandard1.5 which are the ones we are targeting. I could add the Nuget package but decided to replace this Tuple<string, string> for a Type instead.

PR #25, opened 2018-03-16
Setting IsProductComponent to false.

raquelsa:


src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -27,7 +27,7 @@
     <TargetVsixContainerName>Microsoft.Web.LibraryManager.vsix</TargetVsixContainerName>
     <ExtensionInstallationRoot>Extensions</ExtensionInstallationRoot>
     <ExtensionInstallationFolder>Microsoft\Web Tools\Library Manager</ExtensionInstallationFolder>
-    <IsProductComponent>true</IsProductComponent>
+    <IsProductComponent>false</IsProductComponent>
```
jimmylewis: @joeloff should we change the CI configuration to set this to true for the signed builds that get inserted into VS?

joeloff: Yes. To do that this needs to be a conditioned: <IsProductComponent Condition="'$(IsProductComponent)'==''>false</IsProductComponent> This way, if it's undefined it's false, otherwise we can override it from things like the CI

joeloff: Corrected a typo: `<IsProductComponent Condition="'$(IsProductComponent)' == ''">false</IsProductComponent>`

PR #26, opened 2018-03-16
Fixing restore bug: 578106

raquelsa:
- When a libman.json file fails to parse initially, we get a null manifest leading to exceptions

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
             if (manifest != null)
             {
-                foreach (ILibraryInstallationState state in manifest.Libraries.Where(l => l.IsValid(out var errors)))
+                foreach (ILibraryInstallationState state in manifest.Libraries.Where(l => l.IsValid(out IEnumerable<IError> errors)))
                 {
                     IEnumerable<FileIdentifier> stateFiles = await GetFilesWithVersionsAsync(state).ConfigureAwait(false);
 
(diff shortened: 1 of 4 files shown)


src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -76,19 +76,23 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
         {
             var files = new List<FileIdentifier>();
 
-            foreach (ILibraryInstallationState state in manifest.Libraries.Where(l => l.IsValid(out var errors)))
+            if (manifest != null)
             {
-                IEnumerable<FileIdentifier> stateFiles = await GetFilesWithVersionsAsync(state).ConfigureAwait(false);
-
-                foreach (FileIdentifier fileIdentifier in stateFiles)
+                foreach (ILibraryInstallationState state in manifest.Libraries.Where(l => l.IsValid(out var errors)))
```
jimmylewis: nit: var?

jimmylewis: For that matter, if you're not using errors, just do `out _` instead.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -76,19 +76,23 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
...
-                        files.Add(fileIdentifier);
+                        if (!files.Contains(fileIdentifier))
+                        {
+                            files.Add(fileIdentifier);
+                        }
                     }
                 }
             }
 
+
```
jimmylewis: nit: remove redundant empty line.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
raquelsa: This code was there before. We have a bug logged to review use of var throughout LibMan code.

PR #27, opened 2018-03-16
Set EmbedInteropType to true for some of the VS binaries.

raquelsa:
After we switched to PackageReference for VS binaries we forgot to set EmbedInteropType to true for some binaries.

### DIFF
--- a/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
+++ b/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
@@ -411,6 +411,28 @@
       </VSIXSourceItem>
     </ItemGroup>
   </Target>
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -411,6 +411,28 @@
       </VSIXSourceItem>
     </ItemGroup>
   </Target>
+  <!-- This is a copy of the Microsoft.VisualStudio.SDK.EmbedInteropTypes NuGet package, but only the list of
+       assemblies that we need. The package includes things like EnvDTE which are reasonable for consumers, but
```
jimmylewis: The 2nd sentence ("The package includes things like EnvDTE...") doesn't seem appropriate to our project.

raquelsa:
Added the same changes on another pull request

PR #29, opened 2018-03-20
Dev/raquela/bug fixes

raquelsa:
578139: LibMan: restore fails on a valid manifest
586286: LibMan: library files are not properly deleted when updating a library to have Files property

To summarize these changes, basically we need to populate the “Files” property (call UpdateStateAsync) when deleting the files before restoring on “File Save”. We were only doing that during install which was causing unwanted files left behind after restore.
We also needed to implement the UpdateStateAsync for FileSystem so we can also populate the Files property when user specifies a folder as  a library.

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -1,13 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.VisualStudio.Editor;
-using Microsoft.VisualStudio.Language.Intellisense;
-using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.Text.Editor;
-using Microsoft.VisualStudio.TextManager.Interop;
-using Microsoft.VisualStudio.Utilities;
-using Microsoft.Web.LibraryManager.Contracts;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
@@ -15,6 +8,13 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Editor;
+using Microsoft.VisualStudio.Language.Intellisense;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+using Microsoft.VisualStudio.TextManager.Interop;
+using Microsoft.VisualStudio.Utilities;
+using Microsoft.Web.LibraryManager.Contracts;
 
 namespace Microsoft.Web.LibraryManager.Vsix.Json
 {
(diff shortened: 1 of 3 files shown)


src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -100,14 +110,56 @@ public ILibraryCatalog GetCatalog()
...
+            {
+                var catalog = (FileSystemCatalog)GetCatalog();
+                ILibrary library = await catalog.GetLibraryAsync(desiredState.LibraryId, cancellationToken).ConfigureAwait(false);
+
+                if (library == null)
+                {
+                    throw new InvalidLibraryException(desiredState.LibraryId, Id);
+                }
+
+                if (desiredState.Files != null && desiredState.Files.Count > 0)
```
raquelsa: What if desiredState.LibraryId is a file path and desiredState.Files contains different file paths. This currently implementation will use the desiredState.Files instead of desiredState.LibraryId. We should error if desiredState.LibraryId is a file path and Files is specified and is different from desiredState.LibraryId.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -103,7 +109,11 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
             if (catalog != null)
             {
                 ILibrary library = await catalog?.GetLibraryAsync(state.LibraryId, CancellationToken.None);
-                filesWithVersions = library?.Files.Select(f => new FileIdentifier(Path.Combine(state.DestinationPath, f.Key), library.Version));
+                IEnumerable<string> libraryStateFiles = state?.Files?.Where(f => library.Files.Keys.Contains(f));
```
raquelsa: Rename libraryStateFiles to desiredStateFiles

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
alexgav: I would think Files should be relative paths under LibraryId if LibraryId is a directory. If LibraryId is a file, then Files should be empty.

alexgav:
Is there a description somewhere on which bug(s) this is fixing?

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
raquelsa: Jimmy told me that the original implementation had a design to copy the LibraryId file contents to the destination with the Files: property names. Basically, a way to rename files. Not sure this is a clean design and that the outcome of this operation is what the user expects. E.g.: Library 
{
"library": "path/to/file.txt"
"destination": "folder1",
"provider": "filesystem",
"files" : ["newFileName.txt"]
}

This would lead to a new file being created at .\folder1\newFileName.txt with the path/to/file.txt contents. This is what the current implementation does and I could leave as it for now.

raquelsa:
Update the pull request with the bug descriptions.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -103,7 +109,11 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
             if (catalog != null)
             {
                 ILibrary library = await catalog?.GetLibraryAsync(state.LibraryId, CancellationToken.None);
```
raquelsa: Redundant null check?

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -100,14 +110,56 @@ public ILibraryCatalog GetCatalog()
...
         {
-            return Task.FromResult<ILibraryInstallationResult>(LibraryInstallationResult.FromSuccess(desiredState));
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return LibraryInstallationResult.FromCancelled(desiredState);
+            }
+
+            try
+            {
+                var catalog = (FileSystemCatalog)GetCatalog();
```
bhsubra: Can we use "as" instead of casting to avoid exception here and check if catalog is null?

raquelsa: Actually GetCatalog() returns a new instance of FileSystemCatalog so it is ok.

alexgav: Do we really need to cast it here? GetCatalog() returns ILibraryCatalog, which has GetLibraryAsync() on it already. 

In general though, I agree with Bhavya here. Even if it's not needed in the current implementation, it's future-proof and cleaner. 

Anyway - the right change here seems to be to remove the cast and simply use ILibraryCatalog as the variable type (and not "var" :) ).

PR #30, opened 2018-03-20
Fixing correct version of Microsoft.VisualStudio.Shell.Interop.14.0.D…

raquelsa:
…esignTime.

### DIFF
--- a/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
+++ b/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
@@ -425,7 +425,7 @@
     </ItemGroup>
     <ItemGroup>
       <ReferencePath Condition="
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -425,7 +425,7 @@
     </ItemGroup>
     <ItemGroup>
       <ReferencePath Condition="
-              '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.15.0.DesignTime'
+              '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime'
               or '%(FileName)' == 'Microsoft.Internal.VisualStudio.Shell.Interop.14.0.DesignTime'
```
jimmylewis: I think the MS.Internal.VS.Shell.Interop.14.0.DesignTime name is bogus, can you remove it?

PR #32, opened 2018-03-22
Fixing FileNotFoundException when trying to load Nuget.VisualStudio.

raquelsa:
Fixing FileNotFoundException when trying to load Nuget.VisualStudio. 
This failure started happening after we moved to packageReferences. Updating the version of the package fixes the issue.

### DIFF
--- a/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
+++ b/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
@@ -426,7 +426,9 @@
     <ItemGroup>
       <ReferencePath Condition="
               '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime'
+              or '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.12.0'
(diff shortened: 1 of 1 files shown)


jimmylewis:
I believe the root cause of this is that we used to embed interop types.  That regressed in my change where I moved to PackageReference.  Can you go through and make sure our EIT workaround is applied to the same set of assemblies as it was before that change?

raquelsa:
Updated all remaining references that needed EmbedInteropTypes set to true.

src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -426,7 +426,9 @@
     <ItemGroup>
       <ReferencePath Condition="
               '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime'
+              or '%(FileName)' == 'Microsoft.VisualStudio.Shell.Interop.12.0'
```
jimmylewis: Nit: we should sort these for easier maintenance in the future

jimmylewis: I.e. MS.VS.Imaging comes before MS.VS.Shell, and 12 comes before 14.

PR #35, opened 2018-03-22
Fixing unecessary check for validation of InstallationState

raquelsa:
Bug 588017: LibMan: library property completion does not work if there is no defaultPath specified
This is a regression introduced during the schema change PR.


alexgav:
The review comment doesn't match the change. The comment says that "defaultProvider" shouldn't be checked, while the code change removes the check for DesinationPath. Is the DestinationPath optional now?

raquelsa:
Just updated the pull request description. There is a typo as you pointed out. Thanks!

PR #38, opened 2018-03-28
Private assets

jimmylewis:
Set PrivateAssets on references to prevent them from leaking into the consuming project for the Build nupkg.

### DIFF
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
@@ -42,7 +42,7 @@
     </ItemGroup>
 
     <ItemGroup>
-      <ProjectReference Include="..\LibraryManager\Microsoft.Web.LibraryManager.csproj" />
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
```diff
@@ -42,7 +42,7 @@
     </ItemGroup>
 
     <ItemGroup>
-      <ProjectReference Include="..\LibraryManager\Microsoft.Web.LibraryManager.csproj" />
+      <ProjectReference Include="..\LibraryManager\Microsoft.Web.LibraryManager.csproj" PrivateAssets="all" />
```
alexgav: This assembly shouldn't be included / is not needed by the package?

jimmylewis: The assembly is packaged into the .nupkg below.  The PrivateAssets here prevents the Build.nupkg to pull in the core package as a dependency, which would add it to the consuming project as well.

PR #40, opened 2018-03-28
Dev/raquela/project capabilities

raquelsa:
Fixing project type GUIDs and using capabilities for identifying Core Web apps

### DIFF
--- a/src/LibraryManager.Vsix/Constants.cs
+++ b/src/LibraryManager.Vsix/Constants.cs
@@ -12,6 +12,9 @@ namespace Microsoft.Web.LibraryManager.Vsix
         public const string TelemetryNamespace = "webtools/librarymanager/";
         public const string MainNuGetPackageId = "Microsoft.Web.LibraryManager.Build";
         public const string ErrorCodeLink = "https://github.com/aspnet/LibraryManager/wiki/Error-codes#{0}";
+        public const string WAP = "{349C5851-65DF-11DA-9384-00065B846F21}";
+        public const string WEBSITE_PROJECT = "{E24C65DC-7377-472B-9ABA-BC803B73C61A}";
+        public const string DotNetCoreWebCapability = "DotNetCoreWeb";
 
         public static string CacheFolder
         {
@@ -27,5 +30,7 @@ namespace Microsoft.Web.LibraryManager.Vsix
                 return Path.Combine(Environment.ExpandEnvironmentVariables(envVar), ".librarymanager");
             }
         }
+
+        
     }
 }
--- a/src/LibraryManager.Vsix/Constants.cs
+++ b/src/LibraryManager.Vsix/Constants.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Web.LibraryManager.Vsix
         public const string MainNuGetPackageId = "Microsoft.Web.LibraryManager.Build";
         public const string ErrorCodeLink = "https://github.com/aspnet/LibraryManager/wiki/Error-codes#{0}";
         public const string WAP = "{349C5851-65DF-11DA-9384-00065B846F21}";
-        public const string WEBSITE_PROJECT = "{E24C65DC-7377-472B-9ABA-BC803B73C61A}";
+        public const string WebsiteProject = "{E24C65DC-7377-472B-9ABA-BC803B73C61A}";
         public const string DotNetCoreWebCapability = "DotNetCoreWeb";
 
         public static string CacheFolder
--- a/src/LibraryManager.Vsix/Constants.cs
+++ b/src/LibraryManager.Vsix/Constants.cs
@@ -30,7 +30,5 @@ namespace Microsoft.Web.LibraryManager.Vsix
                 return Path.Combine(Environment.ExpandEnvironmentVariables(envVar), ".librarymanager");
             }
         }
-
-        
     }
 }
--- a/src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
+++ b/src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
@@ -16,22 +16,6 @@ namespace Microsoft.Web.LibraryManager.Vsix
 {
     internal static class LibraryHelpers
     {
-        public static bool IsSupported(this Project project)
-        {
-            if (project.IsKind(ProjectTypes.DOTNET_Core, ProjectTypes.WEBSITE_PROJECT))
-                return true;
-
-            try
-            {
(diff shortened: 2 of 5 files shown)


src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -16,22 +16,6 @@ namespace Microsoft.Web.LibraryManager.Vsix
 {
     internal static class LibraryHelpers
     {
-        public static bool IsSupported(this Project project)
```
raquelsa: Not used anywhere....

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -54,7 +54,7 @@ public static void CheckFileOutOfSourceControl(string file)
 
         public static void AddFileToProject(this Project project, string file, string itemType = null)
         {
-            if (project.IsKind(ProjectTypes.ASPNET_5, ProjectTypes.DOTNET_Core, ProjectTypes.SSDT))
```
raquelsa: Not sure what this SSDT project type is, but we only support DotNetCoreWeb, WAPs and Web Sites.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -63,9 +63,7 @@ public static void AddFileToProject(this Project project, string file, string it
                 {
                     ProjectItem item = project.ProjectItems.AddFromFile(file);
 
-                    if (string.IsNullOrEmpty(itemType)
-                        || project.IsKind(ProjectTypes.WEBSITE_PROJECT)
-                        || project.IsKind(ProjectTypes.UNIVERSAL_APP))
```
raquelsa: We don't support Universal apps

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -236,18 +234,29 @@ public static Project GetDTEProject(IVsHierarchy hierarchy)
 
             return null;
         }
-    }
 
-    public static class ProjectTypes
```
raquelsa: Moving the ones we need to the Constants file. We had the wrong WAP GUID here. Fixed it at Constants file.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -236,18 +234,29 @@ public static Project GetDTEProject(IVsHierarchy hierarchy)
...
-    {
-        public const string ASPNET_5 = "{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}";
-        public const string DOTNET_Core = "{9A19103F-16F7-4668-BE54-9A1E7A4F7556}";
-        public const string MISC = "{66A2671D-8FB5-11D2-AA7E-00C04F688DDE}";
-        public const string NODE_JS = "{9092AA53-FB77-4645-B42D-1CCCA6BD08BD}";
-        public const string SOLUTION_FOLDER = "{66A26720-8FB5-11D2-AA7E-00C04F688DDE}";
-        public const string SSDT = "{00d1a9c2-b5f0-4af3-8072-f6c62b433612}";
-        public const string UNIVERSAL_APP = "{262852C6-CD72-467D-83FE-5EEB1973A190}";
-        public const string WAP = "{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}";
```
raquelsa: Wrong WAP GUID. This one is for C# projects

src/LibraryManager.Vsix/Constants.cs
```diff
@@ -12,6 +12,9 @@ public static class Constants
         public const string TelemetryNamespace = "webtools/librarymanager/";
         public const string MainNuGetPackageId = "Microsoft.Web.LibraryManager.Build";
         public const string ErrorCodeLink = "https://github.com/aspnet/LibraryManager/wiki/Error-codes#{0}";
+        public const string WAP = "{349C5851-65DF-11DA-9384-00065B846F21}";
+        public const string WEBSITE_PROJECT = "{E24C65DC-7377-472B-9ABA-BC803B73C61A}";
```
alexgav: Nit: Should we make the constant name mixed case to match the rest of the constants? I.e. WebsiteProject rather than WEBSITE_PROJECT?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -54,7 +54,7 @@ public static void CheckFileOutOfSourceControl(string file)
 
         public static void AddFileToProject(this Project project, string file, string itemType = null)
         {
-            if (project.IsKind(ProjectTypes.ASPNET_5, ProjectTypes.DOTNET_Core, ProjectTypes.SSDT))
+            if (IsCapabilityMatch(project, Constants.DotNetCoreWebCapability))
```
alexgav: Wait, is this right? We are *not* going to add libman.json to .Net Core projects? I think we were excluding only whatever SSDT subtype was. Now we are excluding *all* .Net Core web projects, right? I think we need to double-check with Mads on what he was trying to exclude. I think after this change you won't be able to do "Manage Client Libaries" (which should libman.json) in all projects.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -82,10 +80,10 @@ public static void AddFileToProject(this Project project, string file, string it
 
         public static void AddFilesToProject(this Project project, IEnumerable<string> files)
         {
-            if (project == null || project.IsKind(ProjectTypes.ASPNET_5, ProjectTypes.DOTNET_Core, ProjectTypes.SSDT))
+            if (project == null || IsCapabilityMatch(project, Constants.DotNetCoreWebCapability))
```
alexgav: See my comment above. This doesn't seem right to me, as it will exclude *all* .Net Core Web Projects, right?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
alexgav: Nevermind, I'm being slow, .net core adds everything to the project automatically, so it's a no-op for .net core.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
alexgav: Added automatically to .net core, my bad!

PR #44, opened 2018-03-28
Fixes InstallationTarget for 15.6

jimmylewis:
Even though VS has started using the Minor version value, VSIXes still need to use 15.0.<build> to target a specific release.


jimmylewis:
We may want to update this closer to release time to specify a 15.7 build, since that it technically the first version we want to support.  However, since 15.7 hasn't shipped yet, we don't have a build number to use for that.

PR #46, opened 2018-03-29
Adding error checks for calls to Microsoft.VisualStudio.Shell.Interop methods.

raquelsa:


src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -98,21 +98,23 @@ public static void AddFilesToProject(this Project project, IEnumerable<string> f
...
-            solutionService?.GetProjectOfUniqueName(project.UniqueName, out hierarchy);
-
-            if (hierarchy == null)
-                return;
-
-            var ip = (IVsProject)hierarchy;
-            VSADDRESULT[] result = new VSADDRESULT[files.Count()];
+            if (solutionService != null && ErrorHandler.Succeeded(solutionService.GetProjectOfUniqueName(project.UniqueName, out hierarchy)))
+            {
+                if (hierarchy == null)
```
jimmylewis: nit: curly braces

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -98,21 +98,23 @@ public static void AddFilesToProject(this Project project, IEnumerable<string> f
...
+                    return;
+
+                var ip = (IVsProject)hierarchy;
```
jimmylewis: the name `ip` seems misleading.  I know it was pre-existing, but renaming it would make this code more understandable IMO.

PR #48, opened 2018-04-04
Dev/raquela/log to error list

raquelsa:
Adding error to error list when there are manifest parser errors.


src/LibraryManager.Vsix/Contracts/Logger.cs
```diff
@@ -19,6 +20,18 @@ public void Log(string message, LogLevel level)
             LogEvent(message, level);
         }
 
+        public static void LogToErrorList(IError error, string projectName, string configPath)
```
raquelsa: Rename to LogError

raquelsa:
Sending a new one with different solution.

PR #49, opened 2018-04-05
Fixing null ref exception and logging proper error message for empty …

raquelsa:
Fixing null ref exception and logging proper error message for empty destination file names.
Fixes Bug 587187.

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -81,17 +81,21 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
                 foreach (ILibraryInstallationState state in manifest.Libraries.Where(l => l.IsValid(out IEnumerable<IError> errors)))
                 {
                     IProvider provider = _dependencies.GetProvider(state.ProviderId);
-                    ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, CancellationToken.None).Result;
 
-                    if (updatedStateResult.Success)
+                    if (provider != null)
                     {
-                        IEnumerable<FileIdentifier> stateFiles = await GetFilesWithVersionsAsync(updatedStateResult.InstallationState).ConfigureAwait(false);
+                        ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, CancellationToken.None).Result;
 
-                        foreach (FileIdentifier fileIdentifier in stateFiles)
+                        if (updatedStateResult.Success)
                         {
-                            if (!files.Contains(fileIdentifier))
+                            IEnumerable<FileIdentifier> stateFiles = await GetFilesWithVersionsAsync(updatedStateResult.InstallationState).ConfigureAwait(false);
+
+                            foreach (FileIdentifier fileIdentifier in stateFiles)
                             {
-                                files.Add(fileIdentifier);
+                                if (!files.Contains(fileIdentifier))
+                                {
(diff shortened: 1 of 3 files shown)


src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -97,6 +97,11 @@ public ILibraryCatalog GetCatalog()
                         return LibraryInstallationResult.FromCancelled(desiredState);
                     }
 
+                    if (string.IsNullOrEmpty(file))
+                    {
+                        return new LibraryInstallationResult(desiredState, PredefinedErrors.CouldNotWriteFile(file));
```
alexgav: Would this error look weird when there is no file name? Perhaps a different error would be better - like "Could not write file because file name is missing" or something like that?

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -86,6 +86,11 @@ public ILibraryCatalog GetCatalog()
                         return LibraryInstallationResult.FromCancelled(desiredState);
                     }
 
+                    if (string.IsNullOrEmpty(file))
+                    {
+                        return new LibraryInstallationResult(desiredState, PredefinedErrors.CouldNotWriteFile(file));
```
alexgav: See comment above - I think this would look a bit strange with missing file name.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -81,17 +81,21 @@ public void VsTextViewCreated(IVsTextView textViewAdapter)
                 foreach (ILibraryInstallationState state in manifest.Libraries.Where(l => l.IsValid(out IEnumerable<IError> errors)))
                 {
                     IProvider provider = _dependencies.GetProvider(state.ProviderId);
-                    ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, CancellationToken.None).Result;
 
-                    if (updatedStateResult.Success)
+                    if (provider != null)
```
alexgav: Should we give a "Bad provider id" error or something like that? If the user accidentally mistypes the id, it would be good for them to know what's wrong.

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
raquelsa: The error is: {0} could not be written to disk. Make sure the file name is correct. I could change this to: File "{0}" could not be written to disk. Make sure the file name is correct.

raquelsa: This will require new loc changes at this point. I can log a bug for post preview to get these fixed.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
raquelsa: Yes, we do give a proper error message. This check is only during file deletion before restore. At restore time, we give the proper validation errors.

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
jimmylewis: I think the existing message seems fine.

raquelsa: Logged a bug to review error message. Will push changes as is for preview4.

PR #50, opened 2018-04-05
Adding error to error list when there are manifest parser errors

raquelsa:


src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -291,5 +291,23 @@ public static IVsHierarchy GetHierarchy(Project project)
...
+            {
+                ProjectItem projectItem = DTE.Solution.FindProjectItem(file);
+                if (projectItem != null)
+                {
+                    return projectItem.ContainingProject;
+                }
+            }
+            catch (Exception ex)
+            {
+                System.Diagnostics.Debug.Write(ex);
```
alexgav: Should we log this somewhere? Log, Telemetry, etc.?

src/LibraryManager.Vsix/ErrorList/ErrorList.cs
```diff
@@ -47,6 +47,16 @@ public bool HandleErrors(IEnumerable<ILibraryInstallationResult> results)
             return Errors.Count > 0;
         }
 
+        public void HandleError(IError error)
+        {
+            Errors.Add(new DisplayError(error));
```
alexgav: In general, we should probably try to figure out the line/column of the error, otherwise it will always be (0,0). The way the code currently does it in HandleErrors and AddLineAndColumn it pretty horrible though. We don't need it for the particular error that this fix is adding, but should clean-up this code in the next release.

In general, I am not a fan of how this class is implemented. We create and throw away instance of it all the time, It seems like there should be just one instance, or at least one per project. Also I haven't looked at how the errors get cleared, I am assuming it works somehow since this mostly follows what HandleErrors method does for multiple errors. Might be good to check though.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -162,5 +177,15 @@ private void OnViewClosed(object sender, EventArgs e)
                 doc.FileActionOccurred -= OnFileSaved;
             }
         }
+
+        private void AddErrorToList(IError error)
+        {
+            if (_errorList == null)
+            {
+                _errorList = new ErrorList(_project.Name, _manifestPath);
```
alexgav: I believe we were talking about _project.Name possibly being null and using a default project name in Output Window review? Would it apply here as well?

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -25,6 +26,9 @@ public class TextviewCreationListener : IVsTextViewCreationListener
     {
         private Manifest _manifest;
         private Dependencies _dependencies;
+        private Project _project;
+        private ErrorList _errorList;
```
raquelsa: Clear _erroList on OnViewClosed

src/LibraryManager.Vsix/ErrorList/ErrorList.cs
raquelsa: Yes, I agree. I will open a bug for us to revisit the ErrorList class design.

raquelsa: They were not getting cleared when closing the file. I have a fix for that on another PR once this one goes in.

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
raquelsa: Fixed

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
raquelsa: Fixed.

PR #51, opened 2018-04-05
Version validation and error handling.

raquelsa:
This fixes bug 568399.

### DIFF
--- a/src/LibraryManager/LibraryInstallationResult.cs
+++ b/src/LibraryManager/LibraryInstallationResult.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Web.LibraryManager
         }
 
         /// <summary>Internal use only</summary>
-        public LibraryInstallationResult(IError[] errors)
+        public LibraryInstallationResult(IEnumerable<IError> errors)
         {
             Errors = new List<IError>(errors);
         }
@@ -74,7 +74,7 @@ namespace Microsoft.Web.LibraryManager
         }
 
         /// <summary>Internal use only</summary>
-        public static LibraryInstallationResult FromError(IError[] errors)
+        public static LibraryInstallationResult FromErrors(IEnumerable<IError> errors)
         {
             return new LibraryInstallationResult(errors);
         }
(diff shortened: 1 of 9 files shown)


src/LibraryManager/LibraryInstallationResult.cs
```diff
@@ -67,5 +72,11 @@ public static LibraryInstallationResult FromCancelled(ILibraryInstallationState
                 Cancelled = true
             };
         }
+
+        /// <summary>Internal use only</summary>
+        public static LibraryInstallationResult FromError(IError[] errors)
```
alexgav: Nit: I'd call it FromErrors (plural) since it takes an array. Also I'd make argument type IEnumerable<IError> so make it more general (and you can still pass in an array).

src/LibraryManager/Manifest.cs
```diff
@@ -20,6 +20,7 @@ namespace Microsoft.Web.LibraryManager
     /// </summary>
     public class Manifest
     {
+        private static readonly string[] _supportedVersions = { "1.0" };
```
alexgav: I think this is good for now, though maybe we should think of using a single Version object (rather than a string) and using the value as "highest supported version". I assume we will always support lower than current versions? Something we should probably discuss, but again - I'm totally fine with the string array for now.

raquelsa:
With these changes, version is required. Let me know if you think it should not be.

alexgav:
No strong opinion :) Do we give schema error when it's not specified?


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

src/LibraryManager/Manifest.cs
raquelsa: Yes, we should discuss the compatibility model going forward. I just implemented the simplest support for now.

raquelsa:
Yes, we do have a schema error if not specified.

alexgav:
I think it's fine then.


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

src/LibraryManager/LibraryInstallationResult.cs
raquelsa: Fixed

PR #54, opened 2018-04-05
Fixing incorrect library name IntelliSense

alexgav:
Fixing bug 568369 - [LibMan] IntelliSense too slow in library.json

Library name completion is slow because we are getting library names from a network request. We should look into caching them both to speed up completion and to enable offline scenarios (at least somewhat). For now the fix is scoped to avoiding the wrong completion of library name. It was possible to get completely wrong library name completion because we were leaving completion list up while the user was typing new characters of library name. The list wasn't getting either dismissed or filtered. In general, CompletionController needs design review and will possibly be re-written. For now I'm adding a workaround to make it work similar to NPM package.json library name completion, where we immediately dimiss completion list as soon as the next library name character is typed. Once the user stops typing, we make a network request and bring up completions matching the part of the library name typed in so far.

I'm scoping the fix down to the library names only for now, as this is what the bug pointed out.


alexgav:
I'm going to go ahead and merge this optimistically, we can code review in person tomorrow.

PR #71, opened 2018-04-26
Made change to get the JSONEditorDocuemnt from text buffer

bhsubra:
**What was the change?**
Made a small change to get the JSONEditorDocument from TextBufer instead of TextView

**Why?**
FromTextView(..) was made internal recently. Using public API instead - FromTextBuffer(..) instead

**How was it tested?**
Verified that completion works for the below:
- Files
- Library
- Path
- Provider

### DIFF
--- a/src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
+++ b/src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Web.LibraryManager.Vsix
                 return;
             }
 
-            var doc = JSONEditorDocument.FromTextView(_textView);
+            var doc = JSONEditorDocument.FromTextBuffer(_textView.TextBuffer);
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
```diff
@@ -122,7 +122,7 @@ private void OnCommitted(object sender, EventArgs e)
                 return;
             }
 
-            var doc = JSONEditorDocument.FromTextView(_textView);
+            var doc = JSONEditorDocument.FromTextBuffer(_textView.TextBuffer);
```
jimmylewis: `FromTextView` is implemented by using `textView.TextDataModel.DocumentBuffer`.  Maybe we should do the same here?  I think `_textView.TextBuffer` is not always the same content as the document buffer (maybe for diff?).

bhsubra:
Merged via https://github.com/aspnet/LibraryManager/commit/766487cc55554028eec5a1d97c3e8732b40e56d4

PR #75, opened 2018-05-02
[Do not merge] Global Tool for Libman

prafullbhosale:
Changes include:
- A Global tool project `dotnet-libman`
- Build changes to support building the global tool
  - Refactored package versions to a props files (`build/dependencies.props` & `build/testdependencies.props`).
  - Refactored common properties to a props file (`dir.props`).
- Command implementations (`init`, `restore`, `clean`, `install`, `uninstall`, `cache`)
- Tests for the global tool

Pending changes:
- Add target to run tests from command line.


src/dotnet-libman/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,186 @@
...
+                help.Append($"{Environment.NewLine}{Resources.ExamplesHeader}{Environment.NewLine}");
+                help.Append($"{Examples}{Environment.NewLine}");
+            }
+
+            return help.ToString();
+        }
+
+
+
+        protected void PrintOptionsAndArguments()
```
prafullbhosale: Will remove this method and all its references.

jimmylewis:
I really dislike .msbuild files.  They're a non-standard hack that we use internally and I wish we would be rid of them (I also don't like MSBuild as a build orchestrator, it better serves lower level tasks IMO).  Can this be some other more logical (and maybe even recognized) file type or build system?  We can look at doing this later too if it's too big for now, but I'd like to move away from that in general.

src/dotnet-libman/Commands/LibmanApp.cs
```diff
@@ -0,0 +1,44 @@
...
+            RootDir.ShowInHelpText = false;
+            Verbosity.ShowInHelpText = false;
+
+            return this;
+        }
+
+        private string GetVersion()
+        {
+            FileVersionInfo info = FileVersionInfo.GetVersionInfo(GetType().Assembly.Location);
+            return $"{info.ProductMajorPart}.{info.ProductMinorPart}.{info.ProductBuildPart}.{info.ProductPrivatePart}";
```
jimmylewis: `info.ProductVersion` will get you the version as a string, including any tags that get generated, e.g. `1.0.20+g3a85c785fc`

src/dotnet-libman/Commands/UninstallCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            List<string> errors = new List<string>();
+
+            if (string.IsNullOrWhiteSpace(LibraryId.Value))
+            {
+                errors.Add(Resources.LibraryIdRequiredForUnInstall);
+            }
+            else if (!Provider.HasValue())
+            {
+                if (manifest.Libraries.Count(l => l.LibraryId == LibraryId.Value) > 1)
```
jimmylewis: Should we check here also that there is a match?  Or do we rely on manifest.Uninstall to report the error if the library is not installed?

src/dotnet-libman/Contracts/Dependencies.cs
```diff
@@ -0,0 +1,82 @@
...
+            //TODO: This will scan all dependencies of the tool.
+            // Need to figure out how to handle when extensions are installed.
+            _assemblyPaths = Directory.EnumerateFiles(toolInstallationDir, "*.dll");
+
+            Initialize();
+        }
+
+        public IHostInteraction GetHostInteractions() => _hostInteraction;
+
+
```
jimmylewis: nit: extra line

src/dotnet-libman/Contracts/Dependencies.cs
```diff
@@ -0,0 +1,82 @@
...
+            foreach (IProviderFactory factory in factories)
+            {
+                if (factory != null)
+                {
+                    _providers.Add(factory.CreateProvider(_hostInteraction));
+                }
+            }
+        }
+
+        private IEnumerable<IProviderFactory> GetProvidersFromReflection()
```
jimmylewis: Can we use MEF instead?  Since it would be similar to what we do in the VSIX...

jimmylewis: I see we do it this way in the Build task too.  I wonder if we could consolidate these (CLI, MSBuild task, VSIX) into a single behavior...

src/dotnet-libman/Contracts/HostInteraction.cs
```diff
@@ -0,0 +1,95 @@
...
+        public string WorkingDirectory { get; private set; }
+        public string CacheDirectory { get; }
+        public ILogger Logger { get; }
+
+        public async Task<bool> WriteFileAsync(string path, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken)
+        {
+            var absolutePath = new FileInfo(Path.Combine(WorkingDirectory, path));
+
+            if (absolutePath.Exists)
+                return true;
```
jimmylewis: nit: add braces

src/dotnet-libman/Contracts/HostInteraction.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            if (!absolutePath.FullName.StartsWith(WorkingDirectory))
+                throw new UnauthorizedAccessException();
```
jimmylewis: nit: add braces

src/dotnet-libman/Contracts/HostInteraction.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                return true;
+            }
+
+            absolutePath.Directory.Create();
+
+            using (Stream stream = content.Invoke())
+            {
+                if (stream == null)
+                    return false;
```
jimmylewis: nit: add braces

src/dotnet-libman/Contracts/HostInteraction.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            var absolutePath = new FileInfo(Path.Combine(WorkingDirectory, path));
+
+            if (absolutePath.Exists)
+                return true;
+
+            if (!absolutePath.FullName.StartsWith(WorkingDirectory))
+                throw new UnauthorizedAccessException();
+
+            if (absolutePath.Exists && (absolutePath.Attributes & FileAttributes.ReadOnly) != 0)
```
jimmylewis: Don't we already return true above when `absolutePath.Exists`?

src/dotnet-libman/Contracts/HostInteraction.cs
jimmylewis: It's also a little odd that we return true when we don't write anything to the file because it already exists.  I know this is an existing implementation, but I question whether it's correct...

src/dotnet-libman/HostEnvironment.cs
```diff
@@ -0,0 +1,50 @@
...
+        public EnvironmentSettings EnvironmentSettings { get; }
+        public string ToolInstallationDir { get; } = Path.GetDirectoryName(typeof(HostEnvironment).Assembly.Location);
+
+
+        static object _syncObj = new object();
+        public static HostEnvironment Initialize(EnvironmentSettings settings)
+        {
+            lock (_syncObj)
+            {
+                Instance = new HostEnvironment(settings);
```
jimmylewis: Is this meant to be called multiple times?  Or should it have null checks on `Instance`?

src/dotnet-libman/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+    class Program
+    {
+        static void Main(string[] args)
+        {
+#if DEBUG
+            if (args.Contains("--debug"))
+            {
+                args = args.SkipLast(1).ToArray();
+                Console.WriteLine($"Attach Debugger to process: {System.Diagnostics.Process.GetCurrentProcess().Id}");
+                while (!System.Diagnostics.Debugger.IsAttached);
```
jimmylewis: Why not `Debugger.Launch`?

src/dotnet-libman/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Microsoft.Web.LibraryManager.Tools
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+#if DEBUG
+            if (args.Contains("--debug"))
+            {
+                args = args.SkipLast(1).ToArray();
```
jimmylewis: what if `--debug` is not the last one?

src/dotnet-libman/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                app.Execute(args);
+            }
+            catch (CommandParsingException cpe)
+            {
+                defaultSettings.Logger.Log(string.Format(Resources.InvalidArgumentsMessage, cpe.Command.Name), LogLevel.Error);
+                cpe.Command.ShowHelp();
+            }
+            catch (AggregateException ae)
+            {
+                //defaultSettings.Logger.Log(ae.Message, LogLevel.Error);
```
jimmylewis: nit: remove commented out code

src/dotnet-libman/dotnet-libman.csproj
```diff
@@ -0,0 +1,65 @@
...
+        description=$(Description);
+        repositoryUrl=$(RepositoryUrl);
+        targetframework=$(TargetFramework);
+        repositoryCommit=$(RepositoryCommit); 
+      </NuspecProperties>
+        <!-- 
+        packageIconUrl=$(PackageIconUrl);
+        -->
+    </PropertyGroup>
+</Target>
```
jimmylewis: nit: indent this line and remove extra lines after

src/dotnet-libman/dotnet-libman.csproj
```diff
@@ -0,0 +1,65 @@
...
+    <EmbeddedResource Update="Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+  </ItemGroup>
+
+
+  <!--
+  **********************************************************
+    For packing global tools we need to use the nuspec file
```
jimmylewis: Is this why we have to do the hack in the build script to cache the Nerdbank-generated version info?

test/LibraryManager.Mocks/Microsoft.Web.LibraryManager.Mocks.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
+    <Import Project="..\..\build\testdependencies.props" />
```
jimmylewis: could be consolidated into a `Directory.Build.props` in the test directory.

tools/ShimGenerator/AppHostRestorerProject/Class1.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
```
jimmylewis: Can this file be removed?

build/testdependencies.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+    <Import Project="dependencies.props" />
```
jimmylewis: Any reason why we can't just define these in the same dependencies.props file?  We could label the PropertyGroup if that would make it more clear?

build/testdependencies.props
```diff
@@ -0,0 +1,7 @@
...
+    <PropertyGroup>
+        <MicrosoftNETTestSDKPackageVersion>15.0.0</MicrosoftNETTestSDKPackageVersion>
+        <MSTestPackageVersion>1.1.17</MSTestPackageVersion>
```
jimmylewis: We should avoid sharing package version properties if the packages could rev independently.

dir.props
```diff
@@ -0,0 +1,31 @@
+<Project>
```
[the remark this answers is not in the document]
prafullbhosale: Did not know about this msbuild feature. 
Will do!

src/LibraryManager/LibraryInstallationState.cs
```diff
@@ -8,11 +8,9 @@
 namespace Microsoft.Web.LibraryManager
 {
 
-    /// <summary>
-    /// For internal use only.
-    /// </summary>
+    /// <inheritdoc />
     /// <seealso cref="Microsoft.Web.LibraryManager.Contracts.ILibraryInstallationState" />
-    internal class LibraryInstallationState : ILibraryInstallationState
+    public class LibraryInstallationState : ILibraryInstallationState
```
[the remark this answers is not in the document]
prafullbhosale: Yeah, I am thinking of adding a different API here named `Install()` corresponding to the `Uninstall()`

src/LibraryManager/Manifest.cs
```diff
@@ -131,7 +131,7 @@ private static bool IsValidManifestVersion(string version)
         /// Adds a library to the <see cref="Libraries"/> collection.
         /// </summary>
         /// <param name="state">An instance of <see cref="ILibraryInstallationState"/> representing the library to add.</param>
-        internal void AddLibrary(ILibraryInstallationState state)
+        public void AddLibrary(ILibraryInstallationState state)
```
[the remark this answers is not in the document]
prafullbhosale: Yeah, I am thinking of adding a different API here named `Install()` corresponding to the `Uninstall()`

src/LibraryManager/Manifest.cs
```diff
@@ -220,6 +220,23 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
             }
         }
 
+        /// <summary>
+        /// Uninstalls the specified library and removes it from the <see cref="Libraries"/> collection.
+        /// </summary>
+        /// <param name="libraryId">The library identifier.</param>
+        /// <param name="provider">Provider id</param>
+        /// <param name="deleteFileAction"></param>
+        public void Uninstall(string libraryId, string provider, Action<string> deleteFileAction)
```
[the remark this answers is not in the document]
prafullbhosale: The existing `Uninstall()` API doesn't have an option to specify a `provider` which is needed in cases where the manifest has two libraries with same name but different providers and the user wants to uninstall the library with a specific provider.

src/dotnet-libman/Commands/CleanCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        protected override async Task<int> ExecuteInternalAsync()
+        {
+            Manifest manifest = await GetManifestAsync();
+            IEnumerable<ILibraryInstallationResult> result = manifest.Clean((s) => HostEnvironment.HostInteraction.DeleteFiles(s));
+            IEnumerable<ILibraryInstallationResult> failures = result.Where(r => !r.Success);
+
+            if (failures.Any())
+            {
+                Logger.Log(Resources.CleanFailed, LogLevel.Error);
```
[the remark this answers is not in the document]
prafullbhosale: The individual files being removed are reported on the output.  Maybe we could specify the summary of the action in the form 
```
Clean <succeeded/failed>. XX libraries cleaned.
```

@justcla  thoughts?

src/dotnet-libman/Commands/InitCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+            await manifest.SaveAsync(Settings.ManifestFileName, CancellationToken.None);
+
+            return 0;
+        }
+
+        private void FailIfLibmanJsonExists()
+        {
+            if (File.Exists(Settings.ManifestFileName))
+            {
+                throw new Exception(Resources.InitFailedLibmanJsonFileExists);
```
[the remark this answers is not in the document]
prafullbhosale: I don't have an opinion either ways. 
@justcla ?

src/dotnet-libman/Commands/InstallCommand.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            Manifest manifest = await GetManifestAsync(createIfNotExists: true);
+
+            ValidateParameters(manifest);
+
+            List<string> files = Files.HasValue() ? Files.Values : null;
+
+            ILibraryInstallationState library = new LibraryInstallationState()
+            {
+                LibraryId = LibraryId.Value,
```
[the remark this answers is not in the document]
prafullbhosale: Not yet. 
Right now it will just add 
```json
{
  "libraryId": "jquery"
}
```
and then restore would fail for cdnjs. I am going to look into what we can do here to either 
1. Fail early and not let the library be added to the json file.
2. Figure out the latest version and use that to add the library.

src/dotnet-libman/Commands/LibmanApp.cs
prafullbhosale: I was debating whether we should show the git commit id on the help. 
But now I think it might be better to show that (atleast early on). It will make it easier to repro customer issues.

src/dotnet-libman/Program.cs
prafullbhosale: This needs to be cleaned up.

src/dotnet-libman/dotnet-libman.csproj
prafullbhosale: Yup. There was a bug in the `pack` which prevented packaging a global tool with ProjectReferences, by just running pack on the `.csproj`

src/dotnet-libman/Contracts/HostInteraction.cs
prafullbhosale: It looks like the return value is being used to determine if the file should be expected in that location or not.

prafullbhosale:
cc @joeloff

src/LibraryManager/LibraryInstallationState.cs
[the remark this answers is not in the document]
raquelsa: We can keep this public for now and when we discuss the extensibility model we can decide where these common types will go.

src/LibraryManager/Manifest.cs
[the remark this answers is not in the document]
raquelsa: This is actually updating the Manifest and not actually installing it so the name is good.

src/dotnet-libman/Commands/CleanCommand.cs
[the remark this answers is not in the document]
raquelsa: I am working on providing an IEnumerable<ILibraryInstallationResult> for all operations. I am implementing the Logging for the .vsix with a summary of these resutls (sucess, failure, cancelled, UpToDate) and we can work on getting these messages consistent  for all extensions.

src/LibraryManager.Contracts/PredefinedErrors.cs
```diff
@@ -92,5 +94,32 @@ public static IError PathOutsideWorkingDirectory()
...
+        /// Library is already installed by the provider.
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="providerId"></param>
+        /// <returns></returns>
+        public static IError LibraryAlreadyInstalled(string libraryId, string providerId)
+            => new Error("LIB010", string.Format(Text.ErrorLibraryAlreadyInstalled, libraryId, providerId));
+
+        /// <summary>
+        /// Library cannot be updated as updated version is already installed.
```
raquelsa: If updated version is already installed, do we need an error for that? Probably just a message "Library is already up to date" or something

src/LibraryManager/Manifest.cs
```diff
@@ -127,6 +127,210 @@ private static bool IsValidManifestVersion(string version)
...
+
+            var desiredState = new LibraryInstallationState()
+            {
+                LibraryId = libraryId,
+                Files = files,
+                ProviderId = providerId,
+                DestinationPath = destination
+            };
+
+            if (CheckAlreadyInstalled(desiredState))
```
raquelsa: This only checks if the library specification is already in the manifest but does not actually install. It would fail to refresh libraries that are newer in the cache than the one in the project. I have implemented a IsLibraryUpToDate check and it would return a LibraryInstallationResult withe new property UpToDate set to true.

raquelsa: Take a look at https://github.com/aspnet/LibraryManager/blob/dev/raquela/StressAndPerfFixes/src/LibraryManager/Manifest.cs RestoreLibraryAsync is probably what you want to call here. I can make it public.

raquelsa: The provider checks weather or not the library is up to date or not

src/LibraryManager/Manifest.cs
```diff
@@ -127,6 +127,210 @@ private static bool IsValidManifestVersion(string version)
...
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="libraryToUpdate"></param>
+        /// <param name="newId"></param>
+        /// <param name="deleteFileAction"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<ILibraryInstallationResult> UpdateLibraryAsync(
+            ILibraryInstallationState libraryToUpdate,
```
raquelsa: Update should just figure out the new version is and call uninstall(ol version)  InstallLibrary(new version), no?

src/LibraryManager/Manifest.cs
```diff
@@ -220,6 +424,20 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
             }
         }
 
+        /// <summary>
+        /// Uninstalls the specified library and removes it from the <see cref="Libraries"/> collection.
+        /// </summary>
+        /// <param name="libraryToUninstall">Provider id</param>
+        /// <param name="deleteFileAction"></param>
+        public void Uninstall(ILibraryInstallationState libraryToUninstall, Action<string> deleteFileAction)
```
raquelsa: I have also implemented this one...

src/LibraryManager.Contracts/PredefinedErrors.cs
prafullbhosale: This is for the following case:

User tries to do `libman update jquery@2.2.0 --to-version jquery@3.3.1` 
when the libman.json is as follows:
```json
{                                
  "version": "1.0",              
  "defaultProvider": "cdnjs",    
  "defaultDestination": "lib",   
  "libraries": [                 
    {                            
      "provider": "cdnjs",       
      "library": "jquery@2.2.0", 
      "destination": "lib"   
    },                           
    {                            
      "library": "jquery@3.3.1"  
    }                            
  ]                              
}                                
```

src/LibraryManager/Manifest.cs
prafullbhosale: This check is to prevent user from installing same library more than once to the same location.

src/libman/Commands/CacheCleanCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+            base.Configure(parent);
+
+            Provider = Argument("provider", Resources.CacheCleanProviderArgumentDesc);
+
+            return this;
+        }
+
+        protected override Task<int> ExecuteInternalAsync()
+        {
+            if (string.IsNullOrWhiteSpace(Provider.Value))
```
raquelsa: I have implemented a CacheService where each provider can use and call CacheService.CleanCache(); So all the code below should move to the individual implementation of the providers CacheService

src/LibraryManager/Manifest.cs
prafullbhosale: No, we need to first make sure that the new Version can be installed. For cases when new version won't install (eg. the files specified in the old version are no longer valid for the new version), we shouldn't delete the files on disk.

src/libman/Commands/CacheListCommand.cs
```diff
@@ -0,0 +1,96 @@
...
+using Microsoft.Extensions.CommandLineUtils;
+using Microsoft.Web.LibraryManager.Contracts;
+
+namespace Microsoft.Web.LibraryManager.Tools.Commands
+{
+    /// <summary>
+    /// Defines the cache list command.
+    /// Allows the user to display the contents of the libman cache.
+    /// </summary>
+    internal class CacheListCommand : BaseCommand
```
raquelsa: Do we need this?

src/libman/Commands/CleanCommand.cs
```diff
@@ -0,0 +1,38 @@
...
+            : base(throwOnUnexpectedArg, "clean", Resources.CleanCommandDesc, hostEnvironment)
+        {
+        }
+
+        public override string Remarks => Resources.CleanCommandRemarks;
+
+        protected override async Task<int> ExecuteInternalAsync()
+        {
+            Manifest manifest = await GetManifestAsync();
+            IEnumerable<ILibraryInstallationResult> result = manifest.Clean((s) => HostEnvironment.HostInteraction.DeleteFiles(s));
```
raquelsa: Task<IEnumerable<ILibraryInstallationResult>> CleanAsync

src/libman/Commands/CacheListCommand.cs
prafullbhosale: The command itself? 
How else would the user know what is installed in the cache? 
Or is the question that do we need to allow the user to know what is in the cache?

src/libman/Commands/CleanCommand.cs
prafullbhosale: Is this a new method being added to the Manifest?

src/libman/Contracts/HostInteraction.cs
```diff
@@ -0,0 +1,120 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Web.LibraryManager.Contracts;
+
```
raquelsa: Why IHostInteration interface was not enough for the CLI. If you need more data you can add to that interface. Also, we should implement IHostInteration at your project level and not in the contracts. For now, just reuse the one we have for .Build project so we are consistent. I am planning on creating a base HostInteration Class that implements common behavior among all our "in box" extensions (.Build, .Vsix, .ClI) implementations (e.g.: Write/Read to files).

src/libman/Contracts/IHostInteractionInternal.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Web.LibraryManager.Contracts;
+
+namespace Microsoft.Web.LibraryManager.Tools.Contracts
+{
+    /// <summary>
+    /// Provides a way to interact with the host environment.
+    /// </summary>
```
raquelsa: Add these two to IHostInteration. Actually I have also added DeleteFiles to it

src/libman/IHostEnvironment.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Provides a logger to display messages to user.
+        /// </summary>
+        ILogger Logger { get; set; }
+
+        /// <summary>
+        /// Provides the host interactions to expose host specific operations.
+        /// </summary>
+        IHostInteractionInternal HostInteraction { get; set; }
```
raquelsa: Isn't ILogger a property on IHostInteraction already?

src/libman/IHostEnvironment.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        /// <summary>
+        /// Provides a input reader to get user input.
+        /// </summary>
+        IInputReader InputReader { get; set; }
+
+        /// <summary>
+        /// Provides a logger to display messages to user.
+        /// </summary>
+        ILogger Logger { get; set; }
```
raquelsa: Isn't ILogger a property on IHostInteraction already?

src/libman/Contracts/HostInteraction.cs
prafullbhosale: The reason for adding the `IHostInteractionsInternal` was to not have force others to implement the `UpdateWorkingDirectory()` 
I did intend to move the `DeleteFiles` to the `IHostInteractions` interface, bu the UpdateWorkingDirectory doesn't seem a useful thing in there.

src/LibraryManager/LibraryInstallationState.cs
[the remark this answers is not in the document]
jimmylewis: If you make it public now and change it later, it becomes a breaking change.

src/LibraryManager/Manifest.cs
prafullbhosale: When are you planning to merge your changes to master? 
If your's are merged first, I will rebase and use your implementation.

src/LibraryManager/LibraryInstallationState.cs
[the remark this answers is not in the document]
prafullbhosale: This is already reverted back to being `internal`.

src/libman/Contracts/HostInteraction.cs
```diff
@@ -0,0 +1,120 @@
...
+        public string WorkingDirectory { get; private set; }
+
+        /// <inheritdoc />
+        public string CacheDirectory { get; }
+
+        /// <inheritdoc />
+        public ILogger Logger { get; }
+
+        /// <inheritdoc />
+        public async Task<bool> WriteFileAsync(string path, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken)
```
raquelsa: Move HostInteration should implement our existing Interface IHostInteraction
Also operations to read and write should be the same as the other two extensions. Do you want move read/write operations to a common HostIntegationBase implementation of IHostIntegration. This way we ensure all extensions are wrtting/reading consistently

src/libman/Contracts/Constants.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Web.LibraryManager.Tools.Contracts
+{
+    /// <summary>
+    /// Defines constants for libman.
```
raquelsa: why do you need a copy of these constants. Aren't they the same for all extensions?

prafullbhosale: They weren't public. They should be. I am planning to file follow up issues for moving duplicated/ shared code to right place. 
So as to not make this PR bigger than it already is.

prafullbhosale:
I have a few follow up items from the PR feedback which I plan to address separately:

- Remove IHostInteractionsInternal from libman CLI project 
  - Add those APIs to IHostInteraction
- Merge Constants class that defines the cache directory into a single one. 
- Use CacheService for Cache Commands (depends on #85)

src/LibraryManager.Contracts/PredefinedErrors.cs
raquelsa: I see now!

src/LibraryManager/Manifest.cs
raquelsa: do we want to error for that? This can break a build is user sets restore on build. Should we mark it as a warning, We can try to add support for that later.

prafullbhosale: This is a new API I added explicitly for the libman install command. This isn't the same one called by during the Restore operation.

src/LibraryManager/Manifest.cs
raquelsa: There is a bigger work item for us that (applies to all operations): we need to come up to a plan to "recover from failure". This would restore the files to prvious status in case any operation fails. We can talk about this early next week. 
To determine the delta between "before" and "after" files, take look at the method 

https://github.com/aspnet/LibraryManager/blob/master/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
 
We can move it outside of TextviewCreationListener so we can all have the same implementation

src/libman/Commands/CacheListCommand.cs
raquelsa: The second.  I sent an email about this... I probably missed the discussion on this. Is this to show the catalog (list of libraries available in a given provider) or to show the full libraries you have on cache?

build/tasks/EmbedAppNameInHost.cs
```diff
@@ -0,0 +1,175 @@
...
+using System;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.DotNet.Cli.Utils
+{
+    /// <summary>
+    /// Embeds the App Name into the AppHost.exe
+    /// </summary>
+    public class EmbedAppNameInHost : Build.Utilities.Task
```
raquelsa: In general, we are marking everything internal unless it needs to be exposed for extensibility. Is this the case?

prafullbhosale: This is not part of what we ship. 
This is a custom build task that is needed for building the shims for global tool. 
MsBuild tasks need to be public.

src/LibraryManager/Manifest.cs
```diff
@@ -121,12 +121,295 @@ private static bool IsValidManifestVersion(string version)
...
+        /// <summary>
+        /// Installs a library with the given libraryId
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="providerId"></param>
+        /// <param name="files"></param>
+        /// <param name="destination"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<ILibraryInstallationResult> InstallLibraryAsync(string libraryId, string providerId, IReadOnlyList<string> files, string destination, CancellationToken cancellationToken)
```
raquelsa: Can't you just make RestoreLibraryAsync public? This way we have consistency

prafullbhosale: This is to add a new library to the manifest. 
RestoreAsync restores the ones that it finds in the file (all of them).

src/LibraryManager/Manifest.cs
```diff
@@ -121,12 +121,295 @@ private static bool IsValidManifestVersion(string version)
...
+
+            if (result.Success)
+            {
+                _libraries.Add(desiredState);
+            }
+
+            return result;
+        }
+
+        private async Task<List<ILibraryInstallationState>> CheckLibraryForConflictsAsync(ILibraryInstallationState desiredState, CancellationToken cancellationToken)
```
raquelsa: This check could be called only at RestoreLibraryAsync , if you replace your Install method by RestoreLibraryAsync .

raquelsa: Can you return a IEnumerable<ITaskInstallationResult>. We just need to know which conflicts we have for Logging, correct?

src/LibraryManager/Manifest.cs
```diff
@@ -121,12 +121,295 @@ private static bool IsValidManifestVersion(string version)
...
+            return null;
+        }
+
+        /// <summary>
+        /// Checks the manifest for conflicting libraries.
+        /// </summary>
+        /// <param name="libraries"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        private async Task<IEnumerable<List<ILibraryInstallationState>>> FindAllConflictingLibrariesAsync(IEnumerable<ILibraryInstallationState> libraries, CancellationToken cancellationToken)
```
raquelsa: Shouldn't this method just check for duplicates on the Libraries property of the manifest? I am assuming we need to check if any entry has the same Library name (extracted from the ID) and destination? We have to add a method to the provider to tell us what the name is given an ID. It would be much simpler.

src/LibraryManager/Manifest.cs
```diff
@@ -121,12 +121,295 @@ private static bool IsValidManifestVersion(string version)
...
+
+        /// <summary>
+        /// Updates a given library to the latest version
+        /// </summary>
+        /// <param name="libraryToUpdate">The library to update</param>
+        /// <param name="usePreRelease">Should prelease versions be considered</param>
+        /// <param name="deleteFileAction"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<ILibraryInstallationResult> UpdateLibraryToLatestAsync(
```
raquelsa: You shouldn't have to implement all that. This should follow the same code path for the Update command from the lightbulb in the editor. Find out the replacement, and call install on it. Don't worry about validation at this point. If anything goes wrong, we revert to original state once we have it implemented..

src/LibraryManager/Manifest.cs
prafullbhosale: > Shouldn't this method just check for duplicates on the Libraries property of the manifest?

Without all the error handling and lookups here, user would potentially get irrelevant errors. 
For instance, below should return that jquery@0.0.0 is an invalid library and not that it conflicts with 
```json
{
  "Libraries": [
    {
      "library": "jquery@3.2.1",
      "destination": "lib",
      "provider": "cdnjs"
    },
    {
      "library": "jquery@0.0.0",
      "destination": "lib",
      "provider": "cdnjs"
    }
  ]
}
```

> We have to add a method to the provider to tell us what the name is given an ID.

The current `Catalog.GetLibraryAsync()` is the same method that tells us what the name of the library. Why add another one?

src/LibraryManager/Manifest.cs
prafullbhosale: I looked at that code. That flow cannot be used to for command line scenarios. 
The VS command basically allows moving to a invalid manifest state, which deletes the files first and then fails at restoring the new one. 
(Change libraryId from jquery@3.3.1 to jquery@0.0.0)
Since in VS you can undo, its fine for the user to manually do this, but on commandline, we cannot delete files first and then tell the user 'oops we cannot install the new one'.

PR #85, opened 2018-05-18
Stress and Perf fixes

raquelsa:
Adding a CacheService
Ading Unit Tests for Cache Service
Deleting Empty folders (Not working for all peoject types yet!)
Improving Logging
Adding detection for up to date files

### DIFF
--- a/src/LibraryManager.Build/Contracts/HostInteraction.cs
+++ b/src/LibraryManager.Build/Contracts/HostInteraction.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Web.LibraryManager.Contracts;
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -25,10 +26,14 @@ namespace Microsoft.Web.LibraryManager.Build
             var absolutePath = new FileInfo(Path.Combine(WorkingDirectory, path));
 
             if (absolutePath.Exists)
+            {
                 return true;
+            }
 
             if (!absolutePath.FullName.StartsWith(WorkingDirectory))
+            {
                 throw new UnauthorizedAccessException();
+            }
 
             if (absolutePath.Exists && (absolutePath.Attributes & FileAttributes.ReadOnly) != 0)
             {
@@ -39,17 +44,9 @@ namespace Microsoft.Web.LibraryManager.Build
 
             using (Stream stream = content.Invoke())
             {
-                if (stream == null)
-                    return false;
-
-                using (FileStream writer = File.Create(absolutePath.FullName, 4096, FileOptions.Asynchronous))
+                if (stream == null || !await WriteToFileAsync(absolutePath.FullName, stream).ConfigureAwait(false))
                 {
-                    if (stream.CanSeek)
-                    {
-                        stream.Seek(0, SeekOrigin.Begin);
-                    }
-
-                    await stream.CopyToAsync(writer, 8192, cancellationToken).ConfigureAwait(false);
+                    return false;
                 }
             }
 
@@ -58,25 +55,28 @@ namespace Microsoft.Web.LibraryManager.Build
             return true;
         }
 
-        public void DeleteFiles(params string[] relativeFilePaths)
+        internal static async Task<bool> WriteToFileAsync(string fileName, Stream libraryStream)
         {
-            foreach (string relativeFilePath in relativeFilePaths)
+            try
             {
-                string absoluteFile = Path.Combine(WorkingDirectory, relativeFilePath);
+                Directory.CreateDirectory(Path.GetDirectoryName(fileName));
 
-                try
+                using (FileStream destination = File.Open(fileName, FileMode.Create, FileAccess.Write, FileShare.ReadWrite))
                 {
-                    if (File.Exists(absoluteFile))
-                    {
-                        File.Delete(absoluteFile);
-                        Logger.Log(string.Format(Resources.Text.FileDeleted, relativeFilePath), LogLevel.Operation);
-                    }
+                    await libraryStream.CopyToAsync(destination);
                 }
-                catch (Exception)
-                {
-                    Logger.Log(string.Format(Resources.Text.FileDeleteFail, relativeFilePath), LogLevel.Operation);
-                }
+
+                return true;
             }
+            catch(Exception)
+            {
+                return false;
+            }
+        }
+
+        public Task<bool> DeleteFilesAsync(IEnumerable<string> relativeFilePaths, CancellationToken cancellationToken, bool deleteCleanFolders = true)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/src/LibraryManager.Contracts/IHostInteraction.cs
+++ b/src/LibraryManager.Contracts/IHostInteraction.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -34,11 +35,22 @@ namespace Microsoft.Web.LibraryManager.Contracts
         /// <summary>
         /// Writes a file to disk based on the specified <see cref="ILibraryInstallationState"/>.
         /// </summary>
-        /// <param name="path">The full file path</param>
+        /// <param name="relativeFilePath">The relative file path to the WorkingDirectory </param>
         /// <param name="content">The content of the file to write.</param>
         /// <param name="state">The desired state of the finished installed library.</param>
         /// <param name="cancellationToken">A token that allows cancellation of the file writing.</param>
         /// <returns><code>True</code> if no issues occured while executing this method; otherwise <code>False</code>.</returns>
-        Task<bool> WriteFileAsync(string path, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken);
+        Task<bool> WriteFileAsync(string relativeFilePath, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Deletes files based on the given collection of relative paths to the WorkingDirectory
+        /// </summary>
+        /// <param name="relativeFilePaths"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="deleteCleanFolders"></param>
+        /// <returns></returns>
+        Task<bool> DeleteFilesAsync(IEnumerable<string> relativeFilePaths, CancellationToken cancellationToken, bool deleteCleanFolders = true);
+
+
     }
 }
--- a/src/LibraryManager.Contracts/ILibraryCatalog.cs
+++ b/src/LibraryManager.Contracts/ILibraryCatalog.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Web.LibraryManager.Contracts
         /// <param name="libraryId">The unique library identifier.</param>
         /// <param name="cancellationToken">A token that allows the search to be cancelled.</param>
         /// <returns>An instance of <see cref="ILibraryGroup"/> or <code>null</code>.</returns>
-        Task<ILibrary> GetLibraryAsync(string libraryId, CancellationToken cancellationToken);
+        Task<ILibrary> GetLibraryMetadataAsync(string libraryId, CancellationToken cancellationToken);
 
         /// <summary>
         /// Searches the catalog for the specified search term.
--- a/src/LibraryManager.Contracts/ILibraryInstallationResult.cs
+++ b/src/LibraryManager.Contracts/ILibraryInstallationResult.cs
@@ -24,6 +24,13 @@ namespace Microsoft.Web.LibraryManager.Contracts
         bool Success { get; }
 
         /// <summary>
+        /// <code>True</code> if the library is up to date; otherwise <code>False</code>.
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        bool UpToDate { get; }
+
+        /// <summary>
         /// A list of errors that occured during library installation.
         /// </summary>
         IList<IError> Errors { get; }
--- a/src/LibraryManager.Contracts/IWebRequestHandler.cs
+++ b/src/LibraryManager.Contracts/IWebRequestHandler.cs
@@ -1,0 +1,22 @@
+﻿using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Net.Http;
+
+namespace Microsoft.Web.LibraryManager.Contracts
+{
+    /// <summary>
+    /// Implements functionality to retrieve a resource as a stream 
+    /// </summary>
+    public interface IWebRequestHandler
+    {
+        /// <summary>
+        /// Returns the requested resource as a Stream 
+        /// </summary>
+        /// <param name="url"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task<Stream> GetStreamAsync(string url, CancellationToken cancellationToken);
+        
+    }
+}
--- a/src/LibraryManager.Contracts/Microsoft.Web.LibraryManager.Contracts.csproj
+++ b/src/LibraryManager.Contracts/Microsoft.Web.LibraryManager.Contracts.csproj
@@ -27,6 +27,7 @@
 
     <ItemGroup>
         <PackageReference Include="NerdBank.GitVersioning" Version="2.1.23" PrivateAssets="all" />
+        <PackageReference Include="System.Net.Http" Version="4.3.1" Condition="'$(TargetFramework)' != 'netstandard1.3'" />
     </ItemGroup>
 
   <!-- Warning if Multilingual App Toolkit isn't installed -->
@@ -44,6 +45,9 @@
     <!-- Needed for Multilingual App Toolkit-->
     <ItemGroup>
       <XliffResource Include="MultilingualResources\*.xlf" />
+    </ItemGroup>
+    <ItemGroup>
+      <Compile Remove="ICacheService.cs" />
     </ItemGroup>
 
     <ItemGroup>
--- a/src/LibraryManager.Contracts/PredefinedErrors.cs
+++ b/src/LibraryManager.Contracts/PredefinedErrors.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using Microsoft.Web.LibraryManager.Contracts.Resources;
 
 namespace Microsoft.Web.LibraryManager.Contracts
@@ -14,6 +15,7 @@ namespace Microsoft.Web.LibraryManager.Contracts
     /// </remarks>
     public static class PredefinedErrors
     {
+
         /// <summary>
         /// Represents an unhandled exception that occured in the provider.
         /// </summary>
@@ -92,5 +94,25 @@ namespace Microsoft.Web.LibraryManager.Contracts
         public static IError VersionIsNotSupported(string version)
            => new Error("LIB009", string.Format(Text.ErrorNotSupportedVersion, version));
 
+        /// <summary>
+        /// Failed to download resource
+        /// </summary>
+        /// <returns>The error code LIB010</returns>
+        public static IError FailedToDownloadResource(string url)
+           => new Error("LIB010", string.Format(Text.ErrorUnableToDownloadResource, url));
+
+        /// <summary>
+        /// Failed to delete library
+        /// </summary>
+        /// <returns>The error code LIB011</returns>
+        public static IError CouldNotDeleteLibrary(string libraryId)
+           => new Error("LIB011", string.Format(Text.ErrorCouldNotDeleteLibrary, libraryId));
+
+        /// <summary>
+        /// Destination path has invalid characters
+        /// </summary>
+        /// <returns>The error code LIB012</returns>
+        public static IError DestinationPathHasInvalidCharacters(string destinationPath)
+           => new Error("LIB012", string.Format(Text.ErrorDestinationPathHasInvalidCharacter, destinationPath));
     }
 }
--- a/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
+++ b/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using EnvDTE;
@@ -21,17 +23,21 @@ namespace Microsoft.Web.LibraryManager.Vsix
 
         public string WorkingDirectory { get; }
         public string CacheDirectory => Constants.CacheFolder;
-        public ILogger Logger { get; internal set; } 
+        public ILogger Logger { get; internal set; }
 
-        public async Task<bool> WriteFileAsync(string path, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken)
+        public async Task<bool> WriteFileAsync(string relativePath, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken)
         {
-            var absolutePath = new FileInfo(Path.Combine(WorkingDirectory, path));
+            FileInfo absolutePath = new FileInfo(Path.Combine(WorkingDirectory, relativePath));
 
             if (absolutePath.Exists)
+            {
                 return true;
+            }
 
             if (!absolutePath.FullName.StartsWith(WorkingDirectory))
+            {
                 throw new UnauthorizedAccessException();
+            }
 
             absolutePath.Directory.Create();
 
@@ -53,43 +59,171 @@ namespace Microsoft.Web.LibraryManager.Vsix
                 }
             }
 
-            Logger.Log(string.Format(LibraryManager.Resources.Text.FileWrittenToDisk, path.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
+            Logger.Log(string.Format(LibraryManager.Resources.Text.FileWrittenToDisk, relativePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
 
             return true;
         }
 
-        public void DeleteFiles(params string[] relativeFilePaths)
+        public async Task<bool> DeleteFilesAsync(IEnumerable<string> relativeFilePaths, CancellationToken cancellationToken, bool deleteCleanFolders = true)
         {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return false;
+            }
+
+            List<Task<bool>> deleteFilesTasks = new List<Task<bool>>();
+            List<Task<bool>> deleteFoldersTasks = new List<Task<bool>>();
+            HashSet<string> directories = new HashSet<string>();
+
             foreach (string relativeFilePath in relativeFilePaths)
             {
                 string absoluteFile = new FileInfo(Path.Combine(WorkingDirectory, relativeFilePath)).FullName;
 
-                try
+                if (cancellationToken.IsCancellationRequested)
                 {
-                    ProjectItem item = VsHelpers.DTE.Solution.FindProjectItem(absoluteFile);
-                    Project project = item?.ContainingProject;
+                    cancellationToken.ThrowIfCancellationRequested();
+                }
 
-                    if (project != null)
+                if (File.Exists(absoluteFile))
+                {
+                    deleteFilesTasks.Add(DeleteFileAsync(absoluteFile));
+                }
+
+                if (deleteCleanFolders)
+                {
+                    string directoryPath = Path.GetDirectoryName(absoluteFile);
+                    if (Directory.Exists(directoryPath))
                     {
-                        item.Delete();
-                    }
-                    else
-                    {
-                        if (File.Exists(absoluteFile))
+                        if (!directories.Contains(directoryPath))
                         {
-                            VsHelpers.CheckFileOutOfSourceControl(absoluteFile);
-                            File.Delete(absoluteFile);
+                            directories.Add(directoryPath);
+                            // TO DO : replace for DeleteFolder that also calls 
+                            // DeleteFolderFromProject as needed
+                            deleteFoldersTasks.Add(DeleteFolderFromDisk(directoryPath));
                         }
                     }
-
-                    Logger.Log(string.Format(LibraryManager.Resources.Text.FileDeleted, relativeFilePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
-                }
-                catch (Exception ex)
-                {
-                    Logger.Log(string.Format(LibraryManager.Resources.Text.FileDeleteFail, relativeFilePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
-                    Telemetry.TrackException("deletefilefailed", ex);
                 }
             }
+
+            await Task.WhenAll(deleteFilesTasks);
+
+            if (deleteCleanFolders)
+            {
+                await Task.WhenAll(deleteFoldersTasks);
+            }
+
+            return deleteFilesTasks.All(t => t.Result);
+        }
+
+        // TO DO: Move to the FileSystemHelpers
+        private async Task<bool> DeleteFileAsync(string filePath)
+        {
+            ProjectItem item = VsHelpers.DTE.Solution.FindProjectItem(filePath);
+            Project project = item?.ContainingProject;
+            bool deleteSucceeded = false;
+
+            if (project != null)
+            {
+                deleteSucceeded = await DeleteFileFromProject(item);
+            }
+            else
+            {
+                deleteSucceeded = await DeleteFileFromDisk(filePath);
+            }
+
+            if (deleteSucceeded)
+            {
+                Logger.Log(string.Format(LibraryManager.Resources.Text.FileDeleted, filePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
+            }
+            else
+            {
+                Logger.Log(string.Format(LibraryManager.Resources.Text.FileDeleteFail, filePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
+            }
+
+            return deleteSucceeded;
+        }
+        private Task<bool> DeleteFolderFromDisk (string folderPath)
+        {
+            return Task.Run(() =>
+            {
+                try
+                {
+                    DirectoryInfo directory = new DirectoryInfo(folderPath);
+                    if (directory.Exists && IsDirectoryEmpty(folderPath))
+                    {
+                        directory.Delete(true);
+                    }
+
+                    return true;
+                }
+                catch (Exception)
+                {
+                    return false;
+                }
+            });
+
+        }
+
+        private bool IsDirectoryEmpty(string path)
+        {
+            return !Directory.EnumerateFileSystemEntries(path).Any();
+        }
+
+        private Task<bool> DeleteFileFromDisk(string filePath)
+        {
+            return Task.Run(() =>
+            {
+                try
+                {
+                    FileInfo fileInfo = new FileInfo(filePath);
+                    if (fileInfo.Exists)
+                    {
+                        VsHelpers.CheckFileOutOfSourceControl(filePath);
+                        File.Delete(filePath);
+                    }
+
+                    return true;
+                }
+                catch (Exception)
+                {
+                    // Add telemetry here
+                    return false;
+                }
+            });
+        }
+
+        // TO DO: Move to VS helpers
+        private Task<bool> DeleteFileFromProject(ProjectItem projectItem)
+        {
+            return Task.Run(() =>
+            {
+                if (projectItem != null)
+                {
+                    try
+                    {
+                        projectItem.Delete();
+                        return true;
+                    }
+                    catch (Exception)
+                    {
+                        // TO DO: log Error 
+                        return false;
+                    }
+                }
+
+                return false;
+            });
+        }
+
+        // TO DO: Move to the FileSystemHelpers
+        private bool IsFileUpToDate(FileInfo cacheFile, FileInfo destinationFile)
+        {
+            if (cacheFile.Length != destinationFile.Length || cacheFile.LastWriteTime.CompareTo(destinationFile.LastWriteTime) > 0)
+            {
+                return false;
+            }
+
+            return true;
         }
     }
 }
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using EnvDTE;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
@@ -9,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using EnvDTE;
 using Microsoft.VisualStudio.Editor;
 using Microsoft.VisualStudio.Language.Intellisense;
 using Microsoft.VisualStudio.Text;
@@ -72,7 +72,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
             textView.Closed += OnViewClosed;
         }
 
-        private async Task RemoveFilesAsync(Manifest newManifest)
+        private async Task RemoveFilesAsync(Manifest newManifest, CancellationToken cancellationToken)
         {
             IEnumerable<FileIdentifier> prevFiles = await GetAllManifestFilesWithVersionsAsync(_manifest).ConfigureAwait(false);
             IEnumerable<FileIdentifier> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest).ConfigureAwait(false);
@@ -81,7 +81,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
             if (filesToRemove.Any())
             {
                 var hostInteraction = _dependencies.GetHostInteractions() as HostInteraction;
-                hostInteraction.DeleteFiles(filesToRemove.ToArray());
+                hostInteraction?.DeleteFilesAsync(filesToRemove, cancellationToken);
             }
         }
 
@@ -121,7 +121,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
         private async Task<IEnumerable<FileIdentifier>> GetFilesWithVersionsAsync(ILibraryInstallationState state)
         {
             ILibraryCatalog catalog = _dependencies.GetProvider(state.ProviderId)?.GetCatalog();
-            ILibrary library = await catalog?.GetLibraryAsync(state.LibraryId, CancellationToken.None);
+            ILibrary library = await catalog?.GetLibraryMetadataAsync(state.LibraryId, CancellationToken.None);
             IEnumerable<FileIdentifier> filesWithVersions = new List<FileIdentifier>();
 
             if (library != null && library.Files != null)
@@ -150,7 +150,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
                         if (newManifest != null)
                         {
-                            await RemoveFilesAsync(newManifest).ConfigureAwait(false);
+                            await RemoveFilesAsync(newManifest, new CancellationToken()).ConfigureAwait(false);
 
                             _manifest = newManifest;
 
@@ -159,11 +159,17 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
                         }
                         else
                         {
+                            // TO DO: Restore to previous state
+                            // and add a warning to the Error List
                             AddErrorToList(PredefinedErrors.ManifestMalformed());
                         }
                     }
                     catch (Exception ex)
                     {
+
+                        // TO DO: Restore to previous state
+                        // and add a warning to the Error List
+
                         string textMessage = string.Concat(Environment.NewLine, LibraryManager.Resources.Text.RestoreHasErrors, Environment.NewLine);
 
                         Logger.LogEvent(textMessage, LogLevel.Task);
--- a/src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
+++ b/src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using EnvDTE;
-using Microsoft.Web.LibraryManager.Contracts;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -10,7 +8,9 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using EnvDTE;
 using Microsoft.VisualStudio.Telemetry;
+using Microsoft.Web.LibraryManager.Contracts;
 
 namespace Microsoft.Web.LibraryManager.Vsix
 {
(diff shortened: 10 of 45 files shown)


src/LibraryManager.Build/Contracts/HostInteraction.cs
```diff
@@ -25,10 +26,14 @@ public HostInteraction(string workingDirectory)
...
+            {
                 return true;
+            }
 
             if (!absolutePath.FullName.StartsWith(WorkingDirectory))
+            {
                 throw new UnauthorizedAccessException();
+            }
 
             if (absolutePath.Exists && (absolutePath.Attributes & FileAttributes.ReadOnly) != 0)
```
raquelsa: Remove it

src/LibraryManager.Build/Contracts/HostInteraction.cs
```diff
@@ -58,25 +55,28 @@ public HostInteraction(string workingDirectory)
             return true;
         }
 
-        public void DeleteFiles(params string[] relativeFilePaths)
+        internal static async Task<bool> WriteToFileAsync(string fileName, Stream libraryStream)
```
raquelsa: Remove and use FileHelpers.WriteToFileAsync

src/LibraryManager.Build/Contracts/HostInteraction.cs
```diff
@@ -58,25 +55,28 @@ public HostInteraction(string workingDirectory)
...
+            }
+            catch(Exception)
+            {
+                return false;
             }
         }
+
+        public Task<bool> DeleteFilesAsync(IEnumerable<string> relativeFilePaths, CancellationToken cancellationToken, bool deleteCleanFolders = true)
+        {
+            throw new NotImplementedException();
```
raquelsa: call to DeleteFilesAsync from FileHelpers.

src/LibraryManager.Contracts/ILibraryInstallationResult.cs
```diff
@@ -23,6 +23,13 @@ public interface ILibraryInstallationResult
         /// </remarks>
         bool Success { get; }
 
+        /// <summary>
+        /// <code>True</code> if the library is up to date; otherwise <code>False</code>.
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        bool UpToDate { get; }
```
raquelsa: New status for libraries that are up to date with the one on cache

src/LibraryManager.Contracts/Microsoft.Web.LibraryManager.Contracts.csproj
```diff
@@ -45,6 +46,9 @@
     <ItemGroup>
       <XliffResource Include="MultilingualResources\*.xlf" />
     </ItemGroup>
+    <ItemGroup>
```
raquelsa: Delete this!

src/LibraryManager.Contracts/PredefinedErrors.cs
```diff
@@ -14,6 +15,7 @@ namespace Microsoft.Web.LibraryManager.Contracts
     /// </remarks>
     public static class PredefinedErrors
     {
+
```
raquelsa: Extra line

src/LibraryManager.Contracts/PredefinedErrors.cs
```diff
@@ -92,5 +94,25 @@ public static IError PathOutsideWorkingDirectory()
         public static IError VersionIsNotSupported(string version)
            => new Error("LIB009", string.Format(Text.ErrorNotSupportedVersion, version));
 
+        /// <summary>
```
raquelsa: @justcla : New strings that neeed review and translation. Can you please review and coordinate the translation? 
@prafullbhosale : looks like we are using the same codes for some of these! errors. Lets coordinate.

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -53,43 +59,171 @@ public HostInteraction(string configFilePath, ILogger logger)
...
+                        // TO DO: log Error 
+                        return false;
+                    }
+                }
+
+                return false;
+            });
+        }
+
+        // TO DO: Move to the FileSystemHelpers
```
raquelsa: Move it to FileSystem Helpers

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -53,43 +59,171 @@ public HostInteraction(string configFilePath, ILogger logger)
...
+        private bool IsFileUpToDate(FileInfo cacheFile, FileInfo destinationFile)
+        {
+            if (cacheFile.Length != destinationFile.Length || cacheFile.LastWriteTime.CompareTo(destinationFile.LastWriteTime) > 0)
```
raquelsa: Check for nulls

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -53,43 +59,171 @@ public HostInteraction(string configFilePath, ILogger logger)
...
+                catch (Exception)
+                {
+                    // Add telemetry here
+                    return false;
+                }
+            });
+        }
+
+        // TO DO: Move to VS helpers
+        private Task<bool> DeleteFileFromProject(ProjectItem projectItem)
```
raquelsa: Move to VSHelpers: need to follow up with Bill to get it to properly work for all projects.

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -53,43 +59,171 @@ public HostInteraction(string configFilePath, ILogger logger)
...
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+                }
+
+                if (File.Exists(absoluteFile))
                 {
-                    ProjectItem item = VsHelpers.DTE.Solution.FindProjectItem(absoluteFile);
-                    Project project = item?.ContainingProject;
+                    deleteFilesTasks.Add(DeleteFileAsync(absoluteFile));
```
raquelsa: This should call the FileHelpers.DeleteFilesAsync( , true)l

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -53,43 +59,171 @@ public HostInteraction(string configFilePath, ILogger logger)
...
+            }
+            else
+            {
+                Logger.Log(string.Format(LibraryManager.Resources.Text.FileDeleteFail, filePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
+            }
 
-                    Logger.Log(string.Format(LibraryManager.Resources.Text.FileDeleted, relativeFilePath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
+            return deleteSucceeded;
+        }
+        private Task<bool> DeleteFolderFromDisk (string folderPath)
```
raquelsa: Move to the FileHelpers

src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -150,7 +150,7 @@ private void OnFileSaved(object sender, TextDocumentFileActionEventArgs e)
 
                         if (newManifest != null)
                         {
-                            await RemoveFilesAsync(newManifest).ConfigureAwait(false);
+                            await RemoveFilesAsync(newManifest, new CancellationToken()).ConfigureAwait(false);
```
raquelsa: cancellation token needs to be replaced by the one we get from Background Tasks. New PR for this is comming so will leave like this for now.

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -47,75 +47,151 @@ internal static class LibraryHelpers
...
-                var errorList = new ErrorList(project?.Name, manifest.Key);
-                hasErrors |= errorList.HandleErrors(results);
+            PostRestoreTelemetryData(totalResults, sw.Elapsed);
+            LogEventsSummary(totalResults, sw.Elapsed);
 
-                resultCount += results.Count();
+        }
 
-                foreach (ILibraryInstallationResult result in results.Where(r => r.Success))
+        private static void LogEventsSummary(List<ILibraryInstallationResult> totalResults, TimeSpan elapsedTime)
```
raquelsa: Make it generic enough to take any operation type (Restore, Clean, Uninstall, Update), IEnumerable<ILibraryInstalltionResult> Move to the Logger implementation

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -47,75 +47,151 @@ internal static class LibraryHelpers
...
+        }
+
+        private static void AddErrorsToErrorList(string projectName, string configFile, IEnumerable<ILibraryInstallationResult> results)
+        {
+            var errorList = new ErrorList(projectName, configFile);
+            errorList.HandleErrors(results);
         }
 
         public static async Task UninstallAsync(string configFilePath, string libraryId, CancellationToken cancellationToken)
         {
```
raquelsa: Add logging for start/summary

test/LibraryManager.Test/CacheServiceTest.cs
```diff
@@ -0,0 +1,237 @@
...
+    public class CacheServiceTest
+    {
+
+        private string _filePath;
+        private string _cacheFolder;
+        private string _projectFolder;
+        private string _catalogCacheFile;
+        private IDependencies _dependencies;
+        private HostInteraction _hostInteraction;
+        private CacheService _cacheService;
```
raquelsa: line

test/LibraryManager.Test/ManifestTest.cs
```diff
@@ -89,17 +88,21 @@ public void Cleanup()
...
             Assert.IsTrue(File.Exists(file2));
+            Assert.IsTrue(results.Count() == 1);
+            Assert.IsTrue(results.First().Success);
 
-            manifest.Uninstall(desiredState.LibraryId, (file) => { _hostInteraction.DeleteFile(file); });
+            ILibraryInstallationResult uninstallResult = await manifest.UninstallAsync(desiredState.LibraryId, (file) => _hostInteraction.DeleteFilesAsync(file, token), token);
 
             Assert.IsFalse(File.Exists(file1));
             Assert.IsFalse(File.Exists(file2));
+            Assert.IsTrue(results.Count() == 1);
```
raquelsa: UninstallResult

src/LibraryManager.Contracts/ILibraryInstallationResult.cs
raquelsa: If the library files are in sync with the one in the cache we want to log message saying that library is up to date. Just added this new property to flag in the ILibraryInstallationResult that nothing was changed for that specific library.

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+namespace Microsoft.Web.LibraryManager
+{
+    /// <summary>
+    /// Service to manage basic operations on libraries cache
+    /// </summary>
+    internal class CacheService
+    {
+        // TO DO: Move these expirations to the provider 
+        private readonly int _catalogExpiresAfterDays = 3;
+        private readonly int _libraryExpiresAfterDays = 1;
```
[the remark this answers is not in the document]
raquelsa: I am not sure about that, Actually, for extensibility the provider should tell us the expiration for their resources. 
@justcla , can you please follow up and give us numbers for these?

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+            return await FileHelpers.ReadFileTextAsync(localFile, cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Refreshes the cache for the given set of files if expired 
+        /// </summary>
+        /// <param name="librariesCacheMetadata"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task HydrateCacheAsync(IEnumerable<CacheServiceMetadata> librariesCacheMetadata, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
raquelsa: Can't find RefreshCacheAsync anymore...

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// Service to manage basic operations on libraries cache
+    /// </summary>
+    internal class CacheService
+    {
+        // TO DO: Move these expirations to the provider 
+        private readonly int _catalogExpiresAfterDays = 3;
+        private readonly int _libraryExpiresAfterDays = 1;
+        private IWebRequestHandler _requestHandler;
+
+        public CacheService(IWebRequestHandler requestHandler)
```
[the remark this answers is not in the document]
raquelsa: Actually CacheService should be part of the HostIntegration which should have the cache location.

src/LibraryManager.Build/Contracts/HostInteraction.cs
```diff
@@ -58,25 +55,28 @@ public HostInteraction(string workingDirectory)
...
-                }
-                catch (Exception)
-                {
-                    Logger.Log(string.Format(Resources.Text.FileDeleteFail, relativeFilePath), LogLevel.Operation);
+                    await libraryStream.CopyToAsync(destination);
                 }
+
+                return true;
+            }
+            catch(Exception)
```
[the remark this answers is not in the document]
raquelsa: I am leaving all the fail recovery to the manifest level.

src/LibraryManager.Contracts/IHostInteraction.cs
```diff
@@ -34,11 +35,22 @@ public interface IHostInteraction
...
+        Task<bool> WriteFileAsync(string relativeFilePath, Func<Stream> content, ILibraryInstallationState state, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Deletes files based on the given collection of relative paths to the WorkingDirectory
+        /// </summary>
+        /// <param name="relativeFilePaths"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="deleteCleanFolders"></param>
+        /// <returns></returns>
+        Task<bool> DeleteFilesAsync(IEnumerable<string> relativeFilePaths, CancellationToken cancellationToken, bool deleteCleanFolders = true);
```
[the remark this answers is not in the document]
raquelsa: Removed the optional parameter. We always want to delete empty folders

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -53,43 +59,171 @@ public HostInteraction(string configFilePath, ILogger logger)
...
         {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return false;
+            }
+
+            List<Task<bool>> deleteFilesTasks = new List<Task<bool>>();
+            List<Task<bool>> deleteFoldersTasks = new List<Task<bool>>();
+            HashSet<string> directories = new HashSet<string>();
+
```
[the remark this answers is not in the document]
raquelsa: This is now Synchronous

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -47,75 +47,151 @@ internal static class LibraryHelpers
...
-            var sw = new Stopwatch();
             sw.Start();
-            int resultCount = 0;
-            bool hasErrors = false;
-            var telResult = new Dictionary<string, double>();
 
             foreach (KeyValuePair<string, Manifest> manifest in manifests)
             {
+                Project project = VsHelpers.DTE.Solution?.FindProjectItem(manifest.Key)?.ContainingProject;
+                Logger.LogEvent(string.Format("Restoring packages for {0}...", project.FullName), LogLevel.Operation);
```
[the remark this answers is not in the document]
raquelsa: I had created a helper method for this in previous check in. I fogot to replace it here on this legacy code

src/LibraryManager/Manifest.cs
```diff
@@ -157,67 +157,75 @@ public void AddVersion(string version)
...
 
             foreach (ILibraryInstallationState state in Libraries)
             {
-                if (cancellationToken.IsCancellationRequested)
-                {
-                    results.Add(LibraryInstallationResult.FromCancelled(state));
-                    _hostInteraction.Logger.Log(Resources.Text.RestoreCancelled, LogLevel.Task);
-                    return results;
-                }
+                tasks.Add(RestoreLibraryAsync(state, cancellationToken));
```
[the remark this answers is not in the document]
raquelsa: Yes, I measured and there is a significant perf. benefit. However, we have racing condition when reading/writing to a metadata file. I will await on each individual Task for now.

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /// <summary>
+        /// Returns the provider's catalog from the provided Url to cacheFile
+        /// </summary>
+        /// <param name="url"></param>
+        /// <param name="cacheFile"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<string> GetCatalogAsync(string url, string cacheFile, CancellationToken cancellationToken)
```
prafullbhosale: All these methods at the very least need to verify that the cacheFile is inside the cache root directory and not some arbitrary location.

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Web.LibraryManager.Contracts;
+
+namespace Microsoft.Web.LibraryManager
+{
+    /// <summary>
+    /// Service to manage basic operations on libraries cache
+    /// </summary>
+    internal class CacheService
```
prafullbhosale: This class needs to ensure that different providers don't stomp on each other's files.

src/LibraryManager/CacheService/CacheService.cs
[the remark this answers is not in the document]
prafullbhosale: Sorry, I meant rename `HydrateCacheAsync` to `RefreshCacheAsync`

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+
+            List<Task> hydrateTasks = new List<Task>();
+
+            foreach (CacheServiceMetadata metadata in librariesCacheMetadata)
+            {
+                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-_libraryExpiresAfterDays))
+                {
+                    Task readFileTask = DownloadToFileAsync(metadata.Source, metadata.DestinationPath, cancellationToken);
```
prafullbhosale: Will this potentially lead to firing off many many network requests in parallel?

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /// <summary>
+        /// Downloads a resource from specified url to a destination file
+        /// </summary>
+        /// <param name="url"></param>
+        /// <param name="fileName"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        private async Task DownloadToFileAsync(string url, string fileName, CancellationToken cancellationToken)
```
prafullbhosale: I assume this is not intended to work with filesystem provider?

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        /// <summary>
+        /// Returns library metadata from provided Url to cacheFile
+        /// </summary>
+        /// <param name="url"></param>
+        /// <param name="cacheFile"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+
+        public async Task<string> GetMetadataAsync(string url, string cacheFile, CancellationToken cancellationToken)
```
prafullbhosale: `GetCatalogAsync`, `GetMetadataAsync`, `GetLibraryAsync` 
All do the exact same thing, have the exact same signature and return the same type. 
Why do we need 3 different method names for them? 

Can these methods return the object in the corresponding types instead?

src/LibraryManager/CacheService/WebRequestHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public async Task<Stream> GetStreamAsync(string url, CancellationToken cancellationToken)
+        {
+            try
+            {
+                return await _httpClient.GetStreamAsync(url).WithCancellation(cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception)
+            {
+                throw new ResourceDownloadException(url);
```
prafullbhosale: Set the caught exception as InnerException?

src/LibraryManager/CacheService/WebRequestHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+        private HttpClient _httpClient;
+        private WebRequestHandler()
+        {
+            HttpClientHandler httpMessageHandler = new HttpClientHandler();
+            httpMessageHandler.MaxConnectionsPerServer = 1000;
+
+            _httpClient = new HttpClient(httpMessageHandler);
+        }
+
+        public static IWebRequestHandler Instance { get; } = new WebRequestHandler();
```
prafullbhosale: If this implements `IDisposable`, it doesn't seem right to make it a singleton.

src/LibraryManager/Helpers/FileHelpers.cs
```diff
@@ -12,59 +10,28 @@ namespace Microsoft.Web.LibraryManager
...
             }
 
-            return null;
-        }
+            string directoryPath = Path.GetDirectoryName(fileName);
 
-        public static async Task<Stream> GetFileAsync(string url, string localFile, int expiresAfterDays, CancellationToken cancellationToken)
-        {
-            if (!File.Exists(localFile) || File.GetLastWriteTime(localFile) < DateTime.Now.AddDays(-expiresAfterDays))
+            if (directoryPath != null && !string.IsNullOrEmpty(directoryPath))
```
prafullbhosale: `string.IsNullOrEmpty` check is enough here.

src/LibraryManager/Helpers/FileHelpers.cs
```diff
@@ -12,59 +10,28 @@ namespace Microsoft.Web.LibraryManager
...
+                Directory.CreateDirectory(directoryPath);
 
-        public static async Task<Stream> DownloadFileAsync(string url, string fileName, CancellationToken cancellationToken)
-        {
-            Stream content = null;
-
-            try
-            {
-                using (var client = new HttpClient())
+                using (FileStream destination = File.Open(fileName, FileMode.Create, FileAccess.Write, FileShare.ReadWrite))
```
prafullbhosale: Same as @jodavis' comment above.

src/LibraryManager.Contracts/ILibraryCatalog.cs
```diff
@@ -25,7 +25,7 @@ public interface ILibraryCatalog
         /// <param name="libraryId">The unique library identifier.</param>
         /// <param name="cancellationToken">A token that allows the search to be cancelled.</param>
         /// <returns>An instance of <see cref="ILibraryGroup"/> or <code>null</code>.</returns>
-        Task<ILibrary> GetLibraryAsync(string libraryId, CancellationToken cancellationToken);
+        Task<ILibrary> GetLibraryMetadataAsync(string libraryId, CancellationToken cancellationToken);
```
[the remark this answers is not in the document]
raquelsa: Done

src/LibraryManager.Contracts/IWebRequestHandler.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.IO;
```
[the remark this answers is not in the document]
raquelsa: Done

src/LibraryManager.Contracts/ILibraryCatalog.cs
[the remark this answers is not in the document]
raquelsa: Renamed to GetLibraryAsync

src/LibraryManager/CacheService/WebRequestHandler.cs
raquelsa: Do we need this info?

raquelsa: We need to log telemetry on that

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -24,6 +27,7 @@ internal class CdnjsProvider : IProvider
         public CdnjsProvider(IHostInteraction hostInteraction)
         {
             HostInteraction = hostInteraction;
+            _cacheService = new CacheService(WebRequestHandler.Instance);
```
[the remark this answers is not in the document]
raquelsa: Will do this on the extensibility PR

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -222,6 +226,73 @@ public static bool IsKind(this Project project, params string[] kindGuids)
...
+            Project project = GetDTEProjectFromConfig(configFilePath);
+            IVsHierarchy hierarchy = GetHierarchy(project);
+            IVsProjectBuildSystem bldSystem = hierarchy as IVsProjectBuildSystem;
+            List<string> filesToRemove = filePaths.ToList();
+
+            while (filesToRemove.Any())
+            {
+                List<string> nextBatch = filesToRemove.Take(batchSize).ToList();
+
+                await DeleteProjectItemsInBatchAsync(hierarchy, nextBatch, logAction, cancellationToken);
```
raquelsa: Return here if it fail or continue deleting files ?

src/LibraryManager.Contracts/IHostInteraction.cs
```diff
@@ -34,11 +35,36 @@ public interface IHostInteraction
         /// <summary>
         /// Writes a file to disk based on the specified <see cref="ILibraryInstallationState"/>.
         /// </summary>
-        /// <param name="path">The full file path</param>
+        /// <param name="filePath">The relative file path to the WorkingDirectory </param>
         /// <param name="content">The content of the file to write.</param>
         /// <param name="state">The desired state of the finished installed library.</param>
         /// <param name="cancellationToken">A token that allows cancellation of the file writing.</param>
         /// <returns><code>True</code> if no issues occured while executing this method; otherwise <code>False</code>.</returns>
```
raquelsa: All of these File IO methods will move to another type once I send the PR for extensibility

src/LibraryManager.Contracts/ILibraryInstallationResult.cs
raquelsa: We should probably rename this type to ILibraryOperationResult since we use that for all types of operations and not only Install.

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -2,94 +2,121 @@
...
+            //Delete from project 
+            var logAction = new Action<string, LogLevel>((message, level) => { Logger.Log(message, level); });
+            Project project = VsHelpers.GetDTEProjectFromConfig(_configFilePath);
+            bool deleteFromProject = await VsHelpers.DeleteFilesFromProjectAsync(project, absolutePaths, logAction, cancellationToken);
+            if (deleteFromProject)
+            {
+                return true;
+            }
+
+            // Delete from file system
```
raquelsa: Is this the logic we want? If we fail to delete from project we delete from disk?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -336,5 +372,138 @@ public static Project GetDTEProjectFromConfig(string file)
...
+                foreach (string filePath in filePaths)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    ProjectItem item = DTE.Solution.FindProjectItem(filePath);
+
+                    if (item != null)
+                    {
+                        ProjectItem parentFolder = item.Collection.Parent as ProjectItem;
+                        folders.Add(parentFolder);
```
raquelsa: It only deletes one level up. We need to add logic to determine how far up in the tree we can delete these empty folders

src/LibraryManager/CacheService/CacheService.cs
raquelsa: CacheService will move out of the provider so we own the location the files will be saved. These changes should be in the extensibility PR

src/LibraryManager/CacheService/CacheService.cs
raquelsa: No, FileSystemProvider does not have a cache

src/LibraryManager/CacheService/CacheService.cs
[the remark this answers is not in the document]
raquelsa: Sure!

src/LibraryManager/CacheService/CacheService.cs
raquelsa: Yes, we decided it was ok for files within libraries. We are still restoring libraries sequentially

src/LibraryManager.Contracts/CancellationHelpers.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Web.LibraryManager
```
prafullbhosale: Placing Helper classes in a `Contracts` assembly feels really weird. 
We should ideally be creating a `Common` assembly for this stuff (which would be a good place to put default implementations as well).
Other alternative is to share this code between projects for now.

src/LibraryManager/Manifest.cs
```diff
@@ -251,48 +256,143 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
...
+                if (filesToRemove.Any())
+                {
+                    IHostInteraction hostInteraction = _dependencies.GetHostInteractions();
+                    return await hostInteraction.DeleteFilesAsync(filesToRemove, cancellationToken);
+                }
+            }
 
-            IProvider provider = _dependencies.GetProvider(state.ProviderId);
-            ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, CancellationToken.None).Result;
+            return false;
```
raquelsa: Should return true!!

src/LibraryManager.Contracts/CancellationHelpers.cs
raquelsa: Yes, I agree. I did not want to create a new assembly for that since we are not exposing extensibility for now. This will be part of the extensibility work and we can decide it we want to expose these helper classed or not in the feature. For this specific one, I am more inclined of not exposing at all

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// </summary>
+        /// <param name="fileName"></param>
+        /// <param name="sourceStream"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public static async Task<bool> WriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            string directoryPath = Path.GetDirectoryName(fileName);
```
raquelsa: Null check for filePath

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
jimmylewis: In Core or WSP, this basically amounts to a retry since those are based on file-system (though WSP doesn't refresh).  For WAP the user would see the file go missing?  I think that might be ok, if it's just going to be restored later.  

On the other hand, I suppose on upgrades/uninstalls, they could get left with turds in their project file, but something triggered the delete...

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -2,94 +2,121 @@
...
+        {
+            return System.Threading.Tasks.Task.Run(() =>
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                return FileHelpers.CopyFile(sourcePath, destinationPath);
+            }, cancellationToken);
+        }
+
+        private Task<bool> DeleteFilesFromDisk(IEnumerable<string> filePaths, CancellationToken cancellationToken)
```
jimmylewis: Is this needed if it's just a wrapper?  Should `FileHelpers.DeleteFiles` be async instead?

src/LibraryManager.Vsix/Contracts/Logger.cs
```diff
@@ -88,5 +90,142 @@ private static bool EnsurePane()
...
+                string successText = operation == OperationType.Clean ?
+                                     string.Format(LibraryManager.Resources.Text.CleanLibrariesSucceeded, totalResultsCounts, Math.Round(elapsedTime.TotalSeconds, 2)) :
+                                     string.Format(LibraryManager.Resources.Text.LibrariesRestored, totalResultsCounts, Math.Round(elapsedTime.TotalSeconds, 2));
+
+                LogEvent(successText + Environment.NewLine, LogLevel.Operation);
+            }
+            else if (allCancelled)
+            {
+                string canceledText = operation == OperationType.Clean ?
+                                    string.Format(LibraryManager.Resources.Text.RestoreLibrariesCancelled, totalResultsCounts):
```
jimmylewis: This order feels backwards.  If it's Clean, you're using the Restore string?

src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
```diff
@@ -30,5 +33,78 @@ public void Log(string message, LogLevel level)
...
+            IEnumerable<ILibraryInstallationResult> cancelledRestores = results.Where(r => r.Cancelled);
+            IEnumerable<ILibraryInstallationResult> upToDateRestores = results.Where(r => r.UpToDate);
+
+            bool allSuccess = successfulRestores.Count() == totalResultsCounts;
+            bool allFailed = failedRestores.Count() == totalResultsCounts;
+            bool allCancelled = cancelledRestores.Count() == totalResultsCounts;
+            bool allUpToDate = upToDateRestores.Count() == totalResultsCounts;
+            bool partialSuccess = successfulRestores.Count() < totalResultsCounts;
+
+            Logger.LogEvent("Restoring libraries completed", LogLevel.Status);
```
jimmylewis: hard coded strings here and below?

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -47,85 +46,87 @@ internal static class LibraryHelpers
...
-            var telResult = new Dictionary<string, double>();
 
             foreach (KeyValuePair<string, Manifest> manifest in manifests)
             {
-                IEnumerable<ILibraryInstallationResult> results = await RestoreLibrariesAsync(manifest.Value, cancellationToken).ConfigureAwait(false);
-
-                Project project = VsHelpers.DTE.Solution?.FindProjectItem(manifest.Key)?.ContainingProject;
-                AddFilesToProject(manifest.Key, project, results);
+                Project project = VsHelpers.GetDTEProjectFromConfig(manifest.Key);
+                Logger.LogEvent(string.Format("Restoring packages for {0}...", project.FullName), LogLevel.Operation);
```
jimmylewis: hardcoded string?

src/LibraryManager.Vsix/Shared/LibraryHelpers.cs
```diff
@@ -47,85 +46,87 @@ internal static class LibraryHelpers
...
+            foreach (ILibraryInstallationResult result in results.Where(r => r.Success))
             {
-                Logger.LogEvent(Environment.NewLine + Resources.Text.LibraryRestoredNoChange + Environment.NewLine, LogLevel.Task);
+                if (result.InstallationState.ProviderId != null)
+                {
+                    telResult.TryGetValue(result.InstallationState.ProviderId, out double count);
+                    telResult[result.InstallationState.ProviderId] = count + 1;
+                }
             }
+            telResult.Add("time", elapsedTime.TotalMilliseconds);
```
jimmylewis: is "elapsedtime" a more accurate name for this field?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -222,6 +226,38 @@ public static bool IsKind(this Project project, params string[] kindGuids)
             return false;
         }
 
+        public static async Task<bool> DeleteFilesFromProjectAsync (Project project, IEnumerable<string> filePaths, Action<string, LogLevel> logAction, CancellationToken cancellationToken)
+        {
+            if (IsCapabilityMatch(project, Constants.DotNetCoreWebCapability))
+            {
+                return false;
```
jimmylewis: Should we return false here?  Does that mean that we failed to delete the files from the project?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -222,6 +226,38 @@ public static bool IsKind(this Project project, params string[] kindGuids)
...
+            List<string> filesToRemove = filePaths.ToList();
+
+            while (filesToRemove.Any())
+            {
+                List<string> nextBatch = filesToRemove.Take(batchSize).ToList();
+                bool success = await DeleteProjectItemsInBatchAsync(hierarchy, nextBatch, logAction, cancellationToken);
+
+                if(!success)
+                {
+                    return false;
```
jimmylewis: should we have a retry?  I don't know if it's worth it, depends on if DeleteProjectItemsInBatchAsync is flaky I guess.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -222,6 +226,38 @@ public static bool IsKind(this Project project, params string[] kindGuids)
...
+                }
+
+                await System.Threading.Tasks.Task.Yield();
+
+                int countToDelete = filesToRemove.Count() >= batchSize ? batchSize : filesToRemove.Count();
```
jimmylewis: `Math.Min(batchSize, filesToRemove.Count())`?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -222,6 +226,38 @@ public static bool IsKind(this Project project, params string[] kindGuids)
...
         }
 
+        public static async Task<bool> DeleteFilesFromProjectAsync (Project project, IEnumerable<string> filePaths, Action<string, LogLevel> logAction, CancellationToken cancellationToken)
+        {
+            if (IsCapabilityMatch(project, Constants.DotNetCoreWebCapability))
+            {
+                return false;
+            }
+
+            int batchSize = 10;
```
jimmylewis: Did we do any tuning on batch sizes?  How was 10 picked?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -336,5 +372,138 @@ public static Project GetDTEProjectFromConfig(string file)
 
             return null;
         }
+
+        private static async Task<bool> AddFilesToHierarchyAsync(IVsHierarchy hierarchy, IEnumerable<string> filePaths, Action<string, LogLevel> logAction, CancellationToken cancellationToken)
+        {
+            int batchSize = 10;
```
jimmylewis: Should the batchSize be shared with Delete?

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -336,5 +372,138 @@ public static Project GetDTEProjectFromConfig(string file)
...
+            }
+
+            return true;
+        }
+
+        private static async Task<bool> AddProjectItemsInBatchAsync(IVsHierarchy vsHierarchy, List<string> filePaths, Action<string, LogLevel> logAction, CancellationToken cancellationToken)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            IVsProjectBuildSystem bldSystem = vsHierarchy as IVsProjectBuildSystem;
```
jimmylewis: nit: don't abbreviate names.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -336,5 +372,138 @@ public static Project GetDTEProjectFromConfig(string file)
...
+                            result);
+
+                foreach (string filePath in filePaths)
+                {
+                    logAction.Invoke(string.Format(Resources.Text.LibraryAddedToProject, filePath.Replace('\\', '/')), LogLevel.Operation);
+                }
+            }
+            catch (Exception)
+            {
+                logAction.Invoke(string.Format(Resources.Text.FailToAddLibrariesToProject), LogLevel.Operation);
```
jimmylewis: Is it possible to report which succeeded and failed?  Or does the vsProject commit the batch atomically?

src/LibraryManager/CacheService/CacheService.cs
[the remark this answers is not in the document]
jimmylewis: I would start with like 30 days for library content.  Or let providers optionally provide, but default to something large?

Catalog might expire after 1 day.

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            _requestHandler = requestHandler;
+        }
+
+        /// <summary>
+        /// Deletes files from cache 
+        /// </summary>
+        public void ClearCache()
+        {
+            throw new NotImplementedException();
```
jimmylewis: This is either dead code or a bug waiting to happen.

src/LibraryManager/CacheService/CacheService.cs
jimmylewis: What is the threshold that is acceptable?  What if a library has 50 files (no joke)?

src/LibraryManager/CacheService/CacheServiceMetadata.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// Path for cache file to be written
+        /// </summary>
+        public string DestinationPath { get; private set; }
+
+        /// <summary>
+        /// Source for the cache's contents  
+        /// </summary>
+        public string Source { get; private set; }
+    }
+}
```
jimmylewis: nit: newline at EOF.

src/LibraryManager/CacheService/CacheServiceMetadataComparer.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            return false;
+        }
+
+        public int GetHashCode(CacheServiceMetadata obj)
+        {
+            return obj.GetHashCode();
+        }
+    }
+}
```
jimmylewis: nit: newline at EOF

src/LibraryManager/CacheService/WebRequestHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+        private HttpClient _httpClient;
+        private WebRequestHandler()
+        {
+            HttpClientHandler httpMessageHandler = new HttpClientHandler();
+            _httpClient = new HttpClient(httpMessageHandler);
+        }
+
+        public static IWebRequestHandler Instance { get; } = new WebRequestHandler();
+
+        public void Dispose()
```
jimmylewis: Should this implement the disposable pattern with the boolean flag?

src/LibraryManager/CacheService/WebRequestHandler.cs
jimmylewis: Is this the right place to log that telemetry?  This seems like just a utility and can't provide good perspective on the operation that failed?

src/LibraryManager/LibraryInstallationResult.cs
```diff
@@ -26,16 +27,23 @@ public LibraryInstallationResult(ILibraryInstallationState installationState, pa
         }
 
         /// <summary>Internal use only</summary>
-        public LibraryInstallationResult(IEnumerable<IError> errors)
+        public LibraryInstallationResult(params IError[] error)
         {
-            Errors = new List<IError>(errors);
+            var list = new List<IError>();
```
jimmylewis: Why do we need to allocate a temp (`list`)?

src/LibraryManager/LibraryInstallationResult.cs
```diff
@@ -74,9 +82,18 @@ public static LibraryInstallationResult FromCancelled(ILibraryInstallationState
...
         /// <summary>Internal use only</summary>
-        public static LibraryInstallationResult FromErrors(IEnumerable<IError> errors)
+        public static LibraryInstallationResult FromError(IError error)
         {
-            return new LibraryInstallationResult(errors);
+            return new LibraryInstallationResult(error);
+        }
+
+        /// <summary>Internal use only</summary>
+        public static ILibraryInstallationResult FromUpToDate(ILibraryInstallationState installationState)
```
jimmylewis: can this be internal?

jimmylewis: Or rather, does the comment make sense given that this is on an internal class?

src/LibraryManager/Manifest.cs
```diff
@@ -251,48 +256,143 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
...
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<bool> RemoveUnwantedFilesAsync(Manifest newManifest, CancellationToken cancellationToken)
         {
-            int filesDeleted = 0;
+            if (newManifest != null)
+            {
+                IEnumerable<FileIdentifier> existingFiles = await GetAllManifestFilesWithVersionsAsync(Libraries).ConfigureAwait(false);
+                IEnumerable<FileIdentifier> newFiles = await GetAllManifestFilesWithVersionsAsync(newManifest.Libraries).ConfigureAwait(false);
+                IEnumerable<string> filesToRemove = existingFiles.Where(f => !newFiles.Contains(f)).Select(f => f.Path);
```
jimmylewis: can `newFiles` be a HashSet to avoid making this O(n<sup>2</sup>)?

src/LibraryManager/Manifest.cs
jimmylewis: Unit tests should catch issues like this.

src/LibraryManager/Manifest.cs
```diff
@@ -251,48 +256,143 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
...
+                }
+            }
 
-            IProvider provider = _dependencies.GetProvider(state.ProviderId);
-            ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, CancellationToken.None).Result;
+            return false;
+        }
 
-            if (updatedStateResult.Success)
+        private async Task<IEnumerable<FileIdentifier>> GetAllManifestFilesWithVersionsAsync(IEnumerable<ILibraryInstallationState> libraries)
```
jimmylewis: This method goes really deep.  Can it be refactored to improve readability?

src/LibraryManager/Manifest.cs
```diff
@@ -251,48 +256,143 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
...
+        {
+            var files = new List<FileIdentifier>();
+
+            if (libraries != null)
             {
-                state = updatedStateResult.InstallationState;
-                foreach (string file in state.Files)
+                foreach (ILibraryInstallationState state in libraries.Where(l => l.IsValid(out IEnumerable<IError> errors)))
```
jimmylewis: It doesn't look like you're using errors at all.  Use a [discard](https://docs.microsoft.com/en-us/dotnet/csharp/discards) instead:

    foreach (ILibraryInstallationState state in libraries.Where(l => l.IsValid(out _)))

src/LibraryManager/Manifest.cs
```diff
@@ -251,48 +256,143 @@ public void Uninstall(string libraryId, Action<string> deleteFileAction)
...
 
-            IProvider provider = _dependencies.GetProvider(state.ProviderId);
-            ILibraryInstallationResult updatedStateResult = provider.UpdateStateAsync(state, CancellationToken.None).Result;
+            return false;
+        }
 
-            if (updatedStateResult.Success)
+        private async Task<IEnumerable<FileIdentifier>> GetAllManifestFilesWithVersionsAsync(IEnumerable<ILibraryInstallationState> libraries)
+        {
+            var files = new List<FileIdentifier>();
```
jimmylewis: Use a HashSet to make the `Contains` call faster?  

Does order matter in the return value?

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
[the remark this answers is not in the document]
jimmylewis: You should do it now.  It has nothing to do with extensibility.

test/LibraryManager.Test/CacheServiceTest.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        [TestInitialize]
+        public void Setup()
+        {
+            _cacheFolder = Environment.ExpandEnvironmentVariables(@"%localappdata%\Microsoft\Library\");
```
jimmylewis: This is Windows specific?  Do we need to do something different for other OSs?

jimmylewis: Also, are the unit tests working against the user's live cache?  That seeems... dangerous... if there's a bug?

test/LibraryManager.Test/CacheServiceTest.cs
```diff
@@ -0,0 +1,238 @@
...
+        [TestMethod]
+        [Ignore]
+        public async Task CacheAndRestoreMultipleLibraries_WritesAndReadsSuccessfuly()
+        {
+            IEnumerable<CacheServiceMetadata> metadata = GetMetadata(100);
+            await _cacheService.RefreshCacheAsync(metadata, CancellationToken.None);
+            var manifest = Manifest.FromJson(GetLibManConfig(100), _dependencies);
+            
+            // act 
+            IEnumerable<ILibraryInstallationResult> results = await manifest.RestoreAsync(CancellationToken.None).ConfigureAwait(false);
```
jimmylewis: How is this testing the cache?

test/LibraryManager.Test/CacheServiceTest.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            // verify
+            Assert.IsTrue(true);
```
jimmylewis: ??

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -0,0 +1,219 @@
...
+        /// <param name="secondFile"></param>
+        /// <returns></returns>
+        public static bool AreFilesUpToDate(FileInfo firstFile, FileInfo secondFile)
+        {
+            if (firstFile == null || secondFile == null || !firstFile.Exists || !secondFile.Exists)
+            {
+                return false;
+            }
+
+            if (string.Compare(firstFile.Name, secondFile.Name, true) != 0 ||
```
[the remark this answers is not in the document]
prafullbhosale: > Also, do we need to be case-sensitive on non-Windows platforms?

Not always. 
Mac OS by default is case insensitive.  
dotnet core fx has an open feature request to provide an API to determine this 
https://github.com/dotnet/corefx/issues/1086

This has been open since 2015 though.

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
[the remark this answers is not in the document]
raquelsa: Why do I have to pass it in the constructor now? I will have to remove it later. We don't want providers changing cache services

src/LibraryManager/CacheService/WebRequestHandler.cs
raquelsa: Not at this file, but wherever we are treating this exception.

src/LibraryManager.Contracts/IHostInteraction.cs
raquelsa: That's why I need a new PR for extensibility even before we start exposing extensibility. Unless you want this PR to grow even bigger ... :)

src/LibraryManager.Contracts/CancellationHelpers.cs
raquelsa: Yes, I will move these

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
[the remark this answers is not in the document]
jimmylewis: a) The provider should not be in charge of picking the caching service, that should be provided by the host.
b) It is impossible to provide a mock CacheService in order to unit test the provider.

src/LibraryManager.Vsix/Shared/VsHelpers.cs
prafullbhosale: I think the behavior here is questionable. 
`AddFilesToProjectAsync` method above does not add this file to the project if its a DotNet Core project. 
However, if the user manually includes the file in the project, (for whatever reason), should this method remove it?

prafullbhosale: But you are right. This should not return false.

src/LibraryManager/CacheService/CacheService.cs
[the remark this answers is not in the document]
prafullbhosale: I think till we define what extensibility means, the cache service needs to have its own expiration values
I will change these to 30 days for library and 1 day for catalog and metadata.

test/LibraryManager.Test/CacheServiceTest.cs
raquelsa: 1. Probably, but we haven't ran these tests on other OS's as far as I know.
2. No, this is not the users live cache.

src/LibraryManager/CacheService/CacheServiceMetadataComparer.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public int GetHashCode(CacheServiceMetadata obj)
+        {
+            return obj.GetHashCode();
```
prafullbhosale: This will not work for a dictionary at all.

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -0,0 +1,219 @@
...
+            if (directoryPath != null && !string.IsNullOrEmpty(directoryPath))
+            {
+                DirectoryInfo dir = Directory.CreateDirectory(directoryPath);
+                using (FileStream destination = File.Open(fileName, FileMode.Create, FileAccess.Write, FileShare.None))
+                {
+                    await sourceStream.CopyToAsync(destination);
+                }
+
+                FileInfo fileInfo = new FileInfo(fileName);
+                Debug.Assert(fileInfo.Length != 0, string.Format("File {0} is empty", fileName));
```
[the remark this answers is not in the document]
raquelsa: We had a bug where all files ended up being zero bytes. I put this check in there in case we regress.

test/LibraryManager.Test/CacheServiceTest.cs
raquelsa: It is not, that's why I marked as [Ignore]. I will remove it for now until I get it right.

test/LibraryManager.Test/CacheServiceTest.cs
raquelsa: Test is Ignored. To be removed

src/LibraryManager.Contracts/FileHelpers.cs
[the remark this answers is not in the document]
raquelsa: This is only used for cdnjs provider to check if the destination file is the same as the one in the Cache folder.

src/LibraryManager.Contracts/FileHelpers.cs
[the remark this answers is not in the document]
jimmylewis: Then shouldn't you compare against the source file size instead of 0?

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -0,0 +1,219 @@
...
+                    DirectoryInfo directory = new DirectoryInfo(path);
+                    if (directory.Exists && IsDirectoryEmpty(path))
+                    {
+                        directory.Delete(true);
+                    }
+                }
+
+                return true;
+            }
+            catch (Exception)
```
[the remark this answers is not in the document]
raquelsa: Yes, if one fail, we want to fail all the Clean operation

src/LibraryManager/CacheService/WebRequestHandler.cs
prafullbhosale: Spoke to @raquelsa, we want to keep this class as a singleton. So going to remove the IDisposable for now.

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
raquelsa: No, we can remove this method and call FileHe

src/LibraryManager.Vsix/Contracts/Logger.cs
raquelsa: Please ignore these methods. I have re-written and wil be avialble in my next PR

src/LibraryManager.Vsix/Contracts/PerProjectLogger.cs
raquelsa: Ignore these as well... My new PR has all fixed

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -0,0 +1,219 @@
...
+        /// <param name="sourceStream"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public static async Task<bool> WriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            string directoryPath = Path.GetDirectoryName(fileName);
+
+            if (directoryPath != null && !string.IsNullOrEmpty(directoryPath))
```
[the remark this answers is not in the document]
raquelsa: Fixed

src/LibraryManager.Contracts/FileHelpers.cs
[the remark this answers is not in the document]
raquelsa: removed

test/LibraryManager.Test/CacheServiceTest.cs
raquelsa: removed

PR #97, opened 2018-06-05
Added logic to get project from selected item if null

bhsubra:
Added null check for project and getting it from VsHelpers.GetProjectOfSelectedItem if null


bhsubra:
Raquel is going to look into this further to see why project is null.

bhsubra:
Closing this as @raquelsa is fixing this issue in https://github.com/aspnet/LibraryManager/pull/105

Issue #67, opened 2018-04-19
GitHub as a Provider


bchavez:
As the title suggests, would be nice to set **GitHub** as a provider at a **_package_** level. Sometimes package maintainers do not immediately publish to NPM or bower, and they have fixed bugs sitting in their **GitHub** repo after a **PR** or **commit**. It would be nice to point `LibMan` to a repository on **GitHub** to retrive some code.

Allow me to specify:
* The **GitHub** release **Tag**.
* Or the **commit hash** of a particular branch.
* Or just the **branch** to pull the latest `HEAD`.

Thanks,
Brian

:hourglass_flowing_sand: :mag: ***["But I still haven't found what I'm for..."](https://www.youtube.com/watch?v=wdCJRybAtso)***

davidfowl:
This is a good idea.

tb-mtg:
Can other providers be added too? (like jsdelivr or even npm)

nickalbrecht:
Just branstorming some ideas I wanted to put here. To drive the intellisense, each package source implementation likely needs to be custom. Especially for the Library list. Intellisense for the files within a package is a little easier, provided the source provides a way to download the package in it's entirety, then you can build the intellisense from the contents after expanding it to the temp folder. I'm assuming you wouldn't want a GitHub plugin to depend on a full github client? If a developer is willing to sacrifice the intellisense aspect (especially for unique/obscure sources), it gets even simpler as theoretically you could simply provide an HTTP provider and pull files directly from a URL, resulting in a config similar to this example maybe? Could also allow for the library URL to be a zip file and simply have it download that and extract it to keep intellisense for choosing files.

**Note:** My example below does NOT work. It's just something it'd be nice to be able to do and would be very easy to implement if there was a low level HTTP provider that sacrificed intellisense in favor of wide spread support of disparate package sources.

``` jsonc
{
    "libraries": [
        {
            //GitHub
            "provider": "http",
            //Maybe even a path to a release/zip?
            //"library": "https://github.com/twbs/bootstrap/releases/download/v4.1.0/bootstrap-4.1.0-dist.zip",
            "library": "https://raw.githubusercontent.com/twbs/bootstrap/v4-dev/dist/",
            "files": [
                "js/bootstrap.bundle.js",
                "css/bootstrap.css"
            ],
            "destination": "wwwroot/lib/bootstrap_from_GitHub"
        },
        {
            //Gist
            "provider": "http",
[...]

PR #103, opened 2018-06-06
Adding unpkg support

alexgav:
Adding a new provider for unpkg.com. Please be aware that:

1. Some clean-up is needed, e.g. UnpkgProvider.cs is 90% the same as CdnjsProvider.cs - we need a base class for code re-use. I'll address it for Preview 4.

2. No version completion support for this PR right now - I'll try to push another update in an hour, but it would be great if whoever is reviewing would start now.

3. NpmPackageSearch.cs, NpmPackageVersion.cs and SemanticVersion.cs files were copied from WTE. Minimal changes were made to those files - only JSON parsing and HTTP request handling was modified. So it's probably easiest to compare them with WTE code instead of reviewing them "from scratch."


MartinKolarik:
Why not implement this as a [jsDelivr](https://www.jsdelivr.com/) provider as discussed in #67, which would give you support for npm and GitHub at the same time. There's also an [API](https://github.com/jsdelivr/data.jsdelivr.com) with everything necessary in one place.

alexgav:
I merged this optimistically. It would be good if someone would take a look tomorrow. We can always stop the insertion into VS if needed.

PR #105, opened 2018-06-07
Move DTE calls to main thread

raquelsa:


src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -52,21 +52,21 @@ private void BeforeQueryStatus(object sender, EventArgs e)
             }
         }
 
-        private void ExecuteAsync(object sender, EventArgs e)
+        private async void ExecuteAsync(object sender, EventArgs e)
```
jimmylewis: We should avoid `async void` method which do any substantial work.  Refactor the bulk into an `async Task` method to avoid the pitfalls of `async void`, then await that new method from this one.

This applies to other `async void` methods which also seem to be doing a fair bit of work.

src/LibraryManager.Vsix/Commands/ManageLibrariesCommand.cs
```diff
@@ -41,45 +41,37 @@ private void BeforeQueryStatus(object sender, EventArgs e)
...
+                    manifest.DefaultProvider = "cdnjs";
+                    manifest.Version = Manifest.SupportedVersions.Max().ToString();
 
-                    if (!File.Exists(configFilePath))
-                    {
-                        var dependencies = Dependencies.FromConfigFile(configFilePath);
-                        Manifest manifest = await Manifest.FromFileAsync(configFilePath, dependencies, token);
-                        manifest.DefaultProvider = "cdnjs";
-                        manifest.Version = Manifest.SupportedVersions.Max().ToString();
+                        await manifest.SaveAsync(configFilePath, default(CancellationToken));
```
jimmylewis: Use `CancellationToken.None` instead of `default(CancellationToken)` to indicate that you don't intent to cancel it.

jimmylewis: (This also applies multiple places)

src/LibraryManager.Vsix/Commands/RestoreOnBuildCommand.cs
```diff
@@ -53,7 +72,7 @@ private void BeforeQueryStatus(object sender, EventArgs e)
                 return;
             }
 
-            ProjectItem item = VsHelpers.GetSelectedItem();
+            ProjectItem item = await VsHelpers.GetSelectedItemAsync();
```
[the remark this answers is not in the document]
raquelsa: Yes!

PR #109, opened 2018-06-07
Removing unpkg from UI

alexgav:


src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -376,9 +382,9 @@ private bool CanInstallPackage()
                 await manifest.SaveAsync(_configFileName, CancellationToken.None).ConfigureAwait(false);
 
                 EnvDTE.Project project = VsHelpers.DTE.SelectedItems.Item(1)?.ProjectItem?.ContainingProject;
-                project?.AddFileToProjectAsync(_configFileName);
+                project?.AddFileToProject(_configFileName);
```
bhsubra: Why did we have to do this?

bhsubra: @raquelsa changed this recently. Could you please pull her changes?

alexgav: Ah, thanks, good catch. That was not intended.

PR #112, opened 2018-06-10
Fix hang in UI while installing libraries for unPkg provider

bhsubra:
Made the following changes to fix hang while installing libraries for unPkg provider:
- Use DialogWindow base class
- Set visual studio window's handle as the owner of the dialog
- Enable code path to show UnPkg provider in the UI

Removed some unwanted code as the functionality is provided by default by DialogWindow class


ToddGrun:
>                     string configFilePath = Path.Combine(rootFolder, Constants.ConfigFileName);

I assume we're guaranteed to be on the UI thread at this point?

---
Refers to: src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs:92 in 27a5d69. [](commit_id = 27a5d69e5783f6089215e61221e42942eaaed550, deletion_comment = False)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -46,11 +45,6 @@ public InstallDialogViewModel(Dispatcher dispatcher, ILibraryCommandService libr
             List<IProvider> providers = new List<IProvider>();
             foreach (IProvider provider in deps.Providers.OrderBy(x => x.Id))
             {
-                if (provider.Id == UnpkgProvider.IdText)
```
ToddGrun: Can you explain this a bit to me. It looks like this change adds UnPkg back to the list of providers. Wasn't there some work where the default provider was supposed to be read from the libman.json file and used as the default?

ToddGrun:
>                     string configFilePath = Path.Combine(rootFolder, Constants.ConfigFileName);

Perhaps the lack of using ConfigureAwaits is what guarantees you this (which is questionable in it's own regard)

---
In reply to: [396325754](https://github.com/aspnet/LibraryManager/pull/112#issuecomment-396325754) [](ancestors = 396325754)

---
Refers to: src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs:92 in 27a5d69. [](commit_id = 27a5d69e5783f6089215e61221e42942eaaed550, deletion_comment = False)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Making Unpkg the default provider is something that we want to do for preview 4. This change is for preview 3.

bhsubra:
>                     string configFilePath = Path.Combine(rootFolder, Constants.ConfigFileName);

Made the changes that we discussed offline

Thanks!

---
In reply to: [396327921](https://github.com/aspnet/LibraryManager/pull/112#issuecomment-396327921) [](ancestors = 396327921,396325754)

---
Refers to: src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs:92 in 27a5d69. [](commit_id = 27a5d69e5783f6089215e61221e42942eaaed550, deletion_comment = False)

PR #113, opened 2018-06-11
Dev/yalyu/file save restore3

YangLyu89:
Adding file save restore tests for:

cdnjs and unpackage provider, including adding/deleting file/library.


test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+
+        [TestInitialize()]
+        public void initialize()
+        {
+            _webProject = Solution.ProjectsRecursive[_projectName];
+            _libManConfig = _webProject.Find(SolutionItemFind.FileName, "libman.json");
+        }
+
+        [TestCleanup]
+        public void Cleanup()
```
YangLyu89: This is for revert the libman.json file to its original status and also delete the libraries downloads during each test so as to not affect later tests.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            Editor.KeyboardCommands.Type("\"LICENSE.txt\"");
+
+            _libManConfig.Save();
+            string cwd = Path.Combine(Path.GetDirectoryName(_webProject.FullPath), "wwwroot", "lib", "jquery");
+            LibmanTestsUtility.WaitForRestoredFile(cwd, "LICENSE.txt", caseInsensitive: true);
+            LibmanTestsUtility.WaitForRestoredFileNotPresent(cwd, "dist", caseInsensitive: true);
+        }
+
+        [TestMethod]
+        public void FileSaveRestore_Cdnjs_DeleteLibrary()
```
YangLyu89: >FileSaveRestore_Cdnjs_DeleteLibrary() [](start = 20, length = 37)

Using another existing project with a ready libman.json can reduced the amount of Editor typing related code here.

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,164 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
...
     public class LibmanTestsUtility
     {
+        private static HashSet<string> GetTopLevelDirectoriesAndFiles(string cwd, bool caseInsensitive)
+        {
+            HashSet<string> topLevelItems = caseInsensitive ? new HashSet<string>(StringComparer.OrdinalIgnoreCase) : new HashSet<string>();
+            var dir = new DirectoryInfo(cwd);
+
+            if(dir.Exists)
+            {
+                foreach(FileSystemInfo item in dir.EnumerateDirectories())
```
ToddGrun: Is this the same as Directory.EnumerateFileSystemEntries ?

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,164 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
...
+        public static void WaitForRestoredFile(string cwd, string expectedFileOrFolder, bool caseInsensitive, int timeout = 10000)
+        {
+            WaitForRestoredFiles(cwd, new[] { expectedFileOrFolder }, caseInsensitive, timeout);
+        }
+
+        public static void WaitForRestoredFileNotPresent(string cwd, string expectedFileOrFolder, bool caseInsensitive, int timeout = 10000)
+        {
+            string errorMessage = WaitForRestoredFilesHelper(cwd, new[] { expectedFileOrFolder }, caseInsensitive, timeout);
+
+            if (errorMessage == null)
```
ToddGrun: This is similar to the completion issue in that you appear to be waiting for a timeout for this method to pass

test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
```diff
@@ -67,6 +67,14 @@ public SolutionService Solution
             }
         }
 
+        public string SolutionRootPath
+        {
+            get
+            {
+                return _solutionRootPath;
```
jimmylewis: >_solutionRootPath [](start = 23, length = 17)

Can it just be auto-property instead of having a backing field?

    public string SolutionRootPath { get; private set; }

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,164 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
...
+
+                if (newErrorMessage != null)
+                {
+                    errorMessage = String.Concat(errorMessage, "\r\nFailed even when forcing completion with double timeout");
+                }
+                else
+                {
+                    errorMessage = String.Concat(errorMessage, "\r\n*Didn't* fail when forcing completion with double timeout");
+                }
+
```
jimmylewis: If we're going to do the "doubled timeout" thing in multiple places, it should be its own helper.

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,164 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
 {
     public class LibmanTestsUtility
     {
+        private static HashSet<string> GetTopLevelDirectoriesAndFiles(string cwd, bool caseInsensitive)
```
jimmylewis: >cwd [](start = 77, length = 3)

don't use abbreviated variable names

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,164 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
...
+                foreach(FileSystemInfo item in dir.EnumerateFiles())
+                {
+                    topLevelItems.Add(item.Name);
+                }
+            }
+
+            return topLevelItems;
+        }
+
+        public static void WaitForRestoredFiles(string cwd, IEnumerable<string> expectedFilesAndFolders, bool caseInsensitive, int timeout = 10000)
```
jimmylewis: >cwd [](start = 55, length = 3)

don't use abbreviated variable names

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+    {
+        ProjectTestExtension _webProject;
+        ProjectItemTestExtension _libManConfig;
+        const string _projectName = @"TestProjectCore20";
+
+        [TestInitialize()]
+        public void initialize()
+        {
+            _webProject = Solution.ProjectsRecursive[_projectName];
+            _libManConfig = _webProject.Find(SolutionItemFind.FileName, "libman.json");
```
jimmylewis: >Find(SolutionItemFind.FileName, "libman.json") [](start = 40, length = 46)

This should be in the project root, so use `_webProject["libman.json"]`

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+    public class FileSaveRestoreTests : VisualStudioLibmanHostTest
+    {
+        ProjectTestExtension _webProject;
+        ProjectItemTestExtension _libManConfig;
+        const string _projectName = @"TestProjectCore20";
+
+        [TestInitialize()]
+        public void initialize()
+        {
+            _webProject = Solution.ProjectsRecursive[_projectName];
```
jimmylewis: >ProjectsRecursive [](start = 35, length = 17)

Is the project not in the root?  Usually we do `Solution[_projectName]`

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
jimmylewis: If you have to explain your code with a comment, then the code should be made more clear.  I think you could just add a comment `// Reset libman.json file contents`

You also assume here that the undo stack will always work/be accurate.  A better way to reset would be for TestInitialize to get the file content, and for Cleanup to set it back to the exact same content.

---
In reply to: [194349724](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194349724) [](ancestors = 194349724)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            Editor.Caret.MoveToEndOfLine();
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"defaultProvider\":");
+            LibmanTestsUtility.WaitForCompletionEntry(Editor, "cdnjs", caseInsensitive: true);
+            Editor.KeyboardCommands.Type("cdnjs");
+
+            Editor.Caret.MoveToExpression("\"libraries\"");
+            Editor.Caret.MoveDown(2);
+            Editor.KeyboardCommands.Type("\"library\":");
+            LibmanTestsUtility.WaitForCompletionEntry(Editor, "jquery", caseInsensitive: true, timeout: 5000);
```
jimmylewis: >5000 [](start = 104, length = 4)

This is a really long timeout.  Can we do something to make this shorter?

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            _libManConfig.Save();
+            string cwd = Path.Combine(Path.GetDirectoryName(_webProject.FullPath), "wwwroot", "lib", "jquery");
+            LibmanTestsUtility.WaitForRestoredFiles(cwd, expectedFilesAndFolders, caseInsensitive: true);
+        }
+
+        [TestMethod]
+        public void FileSaveRestore_Unpkg_AddNewLibraryWithoutSpecifingFiles()
+        {
+            _libManConfig.Open();
+            string[] expectedFilesAndFolders = new[] {
```
jimmylewis: >expectedFilesAndFolders [](start = 21, length = 23)

Are you going to verify that there's anything inside of dist?

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+
+            Editor.KeyboardCommands.Right();
+            Editor.KeyboardCommands.Type(",");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"files\": [");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"jquery.js\"");
+
+            _libManConfig.Save();
+            string cwd = Path.Combine(Path.GetDirectoryName(_webProject.FullPath), "wwwroot", "lib", "jquery");
```
jimmylewis: >cwd [](start = 19, length = 3)

don't abbreviate this name.  I can't understand what it represents.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            Editor.KeyboardCommands.Type("\"destination\":");
+            Editor.KeyboardCommands.Type("wwwroot/lib/jquery");
+
+            Editor.KeyboardCommands.Right();
+            Editor.KeyboardCommands.Type(",");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"files\": [");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"LICENSE.txt\"");
+
```
jimmylewis: A much more realistic user scenarios (i.e. much more worth protecting with a test) would be to get the jquery.min.js file.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
jimmylewis: yes it could, that's a good idea.

---
In reply to: [194350175](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194350175) [](ancestors = 194350175)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            _libManConfig.Save();
+            string cwd = Path.Combine(Path.GetDirectoryName(_webProject.FullPath), "wwwroot", "lib", "jquery");
+            LibmanTestsUtility.WaitForRestoredFiles(cwd, expectedFilesAndFolders, caseInsensitive: true);
+
+            Editor.Caret.MoveToExpression("{", 0, 2);
+
+            for (int i = 0; i < 4; ++i)
+            {
+                Editor.Edit.DeleteLine();
+            }
```
jimmylewis: DeleteLine takes a parameter for how many times to do it.  You don't need a loop here.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            Editor.KeyboardCommands.Type("\"defaultProvider\":");
+            LibmanTestsUtility.WaitForCompletionEntry(Editor, "cdnjs", caseInsensitive: true);
+            Editor.KeyboardCommands.Type("cdnjs");
+
+            Editor.Caret.MoveToExpression("\"libraries\"");
+            Editor.Caret.MoveDown(2);
+            Editor.KeyboardCommands.Type("\"library\":");
+            LibmanTestsUtility.WaitForCompletionEntry(Editor, "jquery", caseInsensitive: true, timeout: 5000);
+            Editor.KeyboardCommands.Type("jquery@3.3.1");
+            Editor.KeyboardCommands.Right();
```
jimmylewis: Why do you need this line?  Why can't we type something from jquery through the comma?

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            Editor.Caret.MoveToExpression("{", 0, 2);
+
+            for (int i = 0; i < 4; ++i)
+            {
+                Editor.Edit.DeleteLine();
+            }
+
+            _libManConfig.Save();
+            LibmanTestsUtility.WaitForDeletedFiles(cwd, expectedFilesAndFolders, caseInsensitive: true);
+        }
```
jimmylewis: Shouldn't this just check that there are no files left under the path?  Instead of checking a hardcoded list?

jimmylewis: For that matter, we should be deleting the folder too, but I can't remember if that fix is submitted yet.

---
In reply to: [194555302](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194555302) [](ancestors = 194555302)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
jimmylewis: +1.  Also, shouldn't this be recursive?  You could just use `Directory.EnumerateFileSystemEntries(path, "*", SearchOption.AllDirectories)`

---
In reply to: [194396415](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194396415) [](ancestors = 194396415)

jimmylewis: (That could replace this whole method)

---
In reply to: [194556460](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194556460) [](ancestors = 194556460,194396415)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,344 @@
...
+            Editor.KeyboardCommands.Type("\"jquery.js\",");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"jquery.min.js\"");
+
+            _libManConfig.Save();
+            string cwd = Path.Combine(Path.GetDirectoryName(_webProject.FullPath), "wwwroot", "lib", "jquery");
+            LibmanTestsUtility.WaitForRestoredFiles(cwd, expectedFilesAndFolders, caseInsensitive: true);
+
+            Editor.Edit.DeleteToBeginningOfLine();
+            _libManConfig.Save();
```
jimmylewis: Won't this leave a comma after "jquery.js", which invalidates the json file?  I'm surprised this even works if the file doesn't parse.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: That make sense, I will try using this way to revert the change in the file.

---
In reply to: [194549775](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194549775) [](ancestors = 194549775,194349724)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Looks like there's a bug make this comma work. However, I should not leave the extra comma here.

---
In reply to: [194557761](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194557761) [](ancestors = 194557761)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: During the test design discussion, we are going to verify only top level items. If it's not too much work, I think verifying the sub-folder items also make sense to me.

---
In reply to: [194556684](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194556684) [](ancestors = 194556684,194556460,194396415)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: Yes. But it will have different passing parameters and different message.

---
In reply to: [194397045](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194397045) [](ancestors = 194397045)

test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
YangLyu89: Yes. It can be. I can change it.

---
In reply to: [194547773](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194547773) [](ancestors = 194547773)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: will change

---
In reply to: [194548392](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194548392) [](ancestors = 194548392)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: will change

---
In reply to: [194548281](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194548281) [](ancestors = 194548281)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: For here. I use the doubled timeout for WaitForRestoredFile for file restore. But not using timeout for WaitForRestoredFileNotPresent because it is called after the WaitForRestore just verifying the not specified files are not included. So I think this doubled timeout can be written here.

---
In reply to: [194547985](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194547985) [](ancestors = 194547985)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Will try

---
In reply to: [194548636](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194548636) [](ancestors = 194548636)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: The project folder will be in the root of the solution folder.

---
In reply to: [194548753](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194548753) [](ancestors = 194548753)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Will change. This is current working directory I think. I follow the name pattern from otherwhere in libman.

---
In reply to: [194551848](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194551848) [](ancestors = 194551848)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: I see. I will do this check.

---
In reply to: [194552135](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194552135) [](ancestors = 194552135)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Will try.

---
In reply to: [194552547](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194552547) [](ancestors = 194552547)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Looks like there's a bug. If have comma in the line before, there will be two commas!

---
In reply to: [194552816](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194552816) [](ancestors = 194552816)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Yes. It make sense. I can try verifying there's no files in it

---
In reply to: [194555373](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194555373) [](ancestors = 194555373,194555302)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: I see. This can be optimized.

---
In reply to: [194397045](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194397045) [](ancestors = 194397045)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: I could probably try getting content from the other json file as you point out for reverting file change.

---
In reply to: [194552391](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194552391) [](ancestors = 194552391,194350175)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: This would be an different issue. Because the restoring is in progress after save the libman.json file. It needs sufficient time for restoring before we believe the unexpected file is not presented. 

---
In reply to: [194397045](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194397045) [](ancestors = 194397045)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: get rid of this line as using new json file

---
In reply to: [194894203](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194894203) [](ancestors = 194894203,194550054)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Looks like Editor.Edit.deleteLine doesn't accept parameter

---
In reply to: [194894785](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194894785) [](ancestors = 194894785,194552547)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: get rid of it in new iteration.

---
In reply to: [194895044](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194895044) [](ancestors = 194895044,194552816)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: This function is deleted.

---
In reply to: [194937801](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194937801) [](ancestors = 194937801,194397045)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: It's now only used here for the waitforrestoredfiles.

---
In reply to: [194883862](https://github.com/aspnet/LibraryManager/pull/113#discussion_r194883862) [](ancestors = 194883862,194547985)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,142 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
 {
     public class LibmanTestsUtility
     {
+        private static HashSet<string> GetSubDirectoriesAndFiles(string currentWorkingDirectory, bool caseInsensitive)
```
ToddGrun: >private static HashSet<string> GetSubDirectoriesAndFiles(string currentWorkingDirectory, bool caseInsensitive) [](start = 8, length = 110)

nit: I would probably implement this like the following

        private static HashSet<string> GetSubDirectoriesAndFiles(string currentWorkingDirectory, bool caseInsensitive)
        {
            StringComparer comparer = caseInsensitive ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal;
            IEnumerable<string> subItems = Directory.EnumerateFileSystemEntries(currentWorkingDirectory, "*", SearchOption.AllDirectories);

            return new HashSet(subItems, comparer);
        }

But it doesn't really bother me the other way.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            Editor.Caret.MoveLeft();
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("{");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"destination\":");
+            LibmanTestsUtility.WaitForCompletionEntries(Editor, new string[] { }, caseInsensitive: true);
+            Editor.KeyboardCommands.Type("wwwroot/lib/jquery-validate");
+            Editor.Caret.MoveRight();
+            Editor.KeyboardCommands.Type(",");
+            Editor.KeyboardCommands.Enter();
```
YangLyu89: PathCompletionProvider will throw an exception if I combine it with "destination":
so currently I have to separate the comma to make the test work

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            libManConfig.Open();
+
+            string pathToLibrary = Path.Combine(SolutionRootPath, _projectName, "wwwroot", "lib", "jquery-validate");
+            string[] expectedFiles = new[]
+            {
+                Path.Combine(pathToLibrary, "jquery.validate.js"),
+                Path.Combine(pathToLibrary, "localization", "messages_ar.js"),
+            };
+
+            Editor.Caret.MoveToExpression("\"libraries\"");
```
jimmylewis: This whole block of typing - now you depend on the content generated above from being constant.  Instead just replace the text buffer contents with the document you want:

```
string libmanContents = @"{
everything that goes in the file
}";

Editor.Selection.SelectAll();
Editor.KeyboardCommands.Backspace();
Editor.Edit.InsertTextInBuffer(libmanContents);
libmanConfig.Save();
```

jimmylewis: You should probably do this 3 times in this test:

0. Save the original contents of the file for use in (3)
1. Paste the file contents you want for the first library and verify.
2. Paste the file contents you want for the changed file (added/removed libraries) and verify.
3. Restore the original file contents from 0.

jimmylewis: And those 4 lines could be refactored into a method like `ReplaceFileContents(string)`

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
jimmylewis: The point is that we're copying the logic of retrying with an extra timeout duration from the completion helper method.  Is this a pattern we want to do commonly?  Then there should be a base waiter method to contain that retry logic.

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
```diff
@@ -8,6 +9,142 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
 {
     public class LibmanTestsUtility
```
jimmylewis: This utility class is just becoming a dumping ground for various helpers.  Can you organize these to make them discoverable to new test authors in the future?  This has been a significant compaint about test code on our team for a while, let's not repeat past mistakes.

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: I see, thanks for pointing out

---
In reply to: [195502551](https://github.com/aspnet/LibraryManager/pull/113#discussion_r195502551) [](ancestors = 195502551)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
YangLyu89: Will do

---
In reply to: [195512080](https://github.com/aspnet/LibraryManager/pull/113#discussion_r195512080) [](ancestors = 195512080)

YangLyu89: Will add the function

---
In reply to: [195511264](https://github.com/aspnet/LibraryManager/pull/113#discussion_r195511264) [](ancestors = 195511264)

YangLyu89: will add

---
In reply to: [195512249](https://github.com/aspnet/LibraryManager/pull/113#discussion_r195512249) [](ancestors = 195512249)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: Seperate it into two helper class

---
In reply to: [195513583](https://github.com/aspnet/LibraryManager/pull/113#discussion_r195513583) [](ancestors = 195513583)

test/LibraryManager.IntegrationTest/LibmanTestsUtility.cs
YangLyu89: I see. using a delagate waiter for both waitforrestoredfileshelper and waitfordeletedfileshelper function 

---
In reply to: [195513255](https://github.com/aspnet/LibraryManager/pull/113#discussion_r195513255) [](ancestors = 195513255)

YangLyu89:
close this one as merged to preview 4

PR #115, opened 2018-06-11
Treating exceptions at provider/catalog methods 

raquelsa:


src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -144,12 +144,21 @@ public ILibraryCatalog GetCatalog()
...
                 {
-                    throw new InvalidLibraryException(desiredState.LibraryId, Id);
+                    return new LibraryOperationResult(desiredState, PredefinedErrors.UnableToResolveSource(desiredState.LibraryId, desiredState.ProviderId));
                 }
 
                 if (desiredState.Files != null && desiredState.Files.Count > 0)
                 {
-                    return LibraryOperationResult.FromSuccess(desiredState);
+                    IReadOnlyList<string> invalidFiles = library.GetInvalidFiles(desiredState.Files);
+                    if (invalidFiles.Any())
```
ToddGrun: >Any() [](start = 37, length = 5)

No need for Linq here, right?

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -235,17 +235,17 @@ public ILibraryCatalog GetCatalog()
...
 
             return name;
         }
 
-        private string GetLibraryVersion(ILibraryInstallationState state)
+        public string GetLibraryVersion(string libraryId)
         {
-            string[] args = state.LibraryId.Split('@');
+            string[] args = libraryId.Split('@');
             string version = args[1];
```
ToddGrun: >string version = args[1]; [](start = 12, length = 25)

what happens if no '@' ?

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
```diff
@@ -235,17 +235,17 @@ public ILibraryCatalog GetCatalog()
             return null;
         }
 
-        private string GetLibraryName(ILibraryInstallationState state)
+        public string GetLibraryName(string libraryId)
```
ToddGrun: >GetLibraryName [](start = 22, length = 14)

Would it make sense to instead have one method like 

public void GetLibraryNameAndVersion(string libraryName, out string name, out string version)

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -158,53 +172,69 @@ private void GetFiles(JObject fileObject, List<string> files)
...
         public async Task<IReadOnlyList<ILibraryGroup>> SearchAsync(string term, int maxHits, CancellationToken cancellationToken)
         {
-            IEnumerable<string> packageNames = await NpmPackageSearch.GetPackageNamesAsync(term, CancellationToken.None);
-            IEnumerable<ILibraryGroup> libraryGroups = packageNames.Select(packageName => new UnpkgLibraryGroup(packageName));
+            List<ILibraryGroup> libraryGroups = new List<ILibraryGroup>();
+
+            try
+            {
+                IEnumerable<string> packageNames = await NpmPackageSearch.GetPackageNamesAsync(term, CancellationToken.None);
+                libraryGroups = packageNames.Select(packageName => new UnpkgLibraryGroup(packageName)).ToList<ILibraryGroup>();
```
ToddGrun: >ToList<ILibraryGroup>() [](start = 103, length = 23)

You've already newed up a list, might as well just AddRange to it.

src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
```diff
@@ -129,12 +129,10 @@ public CdnjsCatalog(CdnjsProvider provider)
...
             {
-                string[] args = libraryId.Split('@');
-                string name = args[0];
-                string version = (args.Count() > 1) ? args[1] :null;
+                string name = _provider.GetLibraryName(libraryId);
+                string version = _provider.GetLibraryVersion(libraryId);
 
                 IEnumerable<Asset> assets = await GetAssetsAsync(name, cancellationToken).ConfigureAwait(false);
-
                 Asset asset = assets.FirstOrDefault(a => a.Version == version);
```
ToddGrun: >FirstOrDefault [](start = 37, length = 14)

might as well just do a .First(...) and have the exception thrown on your behalf and delete the asset == null check below.

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
raquelsa: Yes, this is another bug. I started implementing this as part of #123 But, this won't make to Escrow bar so leaving the proper fix to preview4

src/LibraryManager/Providers/Cdnjs/CdnjsProvider.cs
raquelsa: Planning on fixing this as part of #123. Unless, we get approval for the proper fix to make to preview3 escrow. 
@jodavis , can you please comment?

jodavis: The goal for Preview 3 is to just get us to not crash. As long as the IndexOutOfRange exception is handled, I think we're okay for now. For Preview 4, we should handle this error case better.

---
In reply to: [194856346](https://github.com/aspnet/LibraryManager/pull/115#discussion_r194856346) [](ancestors = 194856346)

bhsubra:
Had an offline discussion with @jodavis , @raquelsa and @ToddGrun . We decided on taking the changes in this pr as is and include a proper fix for preview 4.

PR #126, opened 2018-06-11
Recursively check if files are selected

bhsubra:
Recursively check if files are selected to makes sure if we are good to install.

### DIFF
--- a/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
+++ b/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
@@ -320,12 +320,12 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Models
         private bool CanInstallPackage()
         {
             if (_isInstalling)
-{
-    return false;
-}
+            {
+                return false;
+            }
 
-AnyFileSelected = IsAnyFileSelected(DisplayRoots);
-return AnyFileSelected;
+            AnyFileSelected = IsAnyFileSelected(DisplayRoots);
+            return AnyFileSelected;
         }
 
         private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
@@ -345,7 +345,7 @@ return AnyFileSelected;
 
                     toProcess.AddRange(child.Children);
                 }
-            }        
+            }
 
             return false;
         }
--- a/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
+++ b/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
@@ -319,38 +319,38 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Models
 
         private bool CanInstallPackage()
         {
-            if (DisplayRoots != null)
-            {
-                CheckIfFilesAreSelected(DisplayRoots);
-            }
-
-            return !_isInstalling && !NoFilesSelected;
+            return !_isInstalling && IsAtleastOneFileSelected(DisplayRoots);
         }
 
-        private void CheckIfFilesAreSelected(IReadOnlyList<PackageItem> children)
+        private bool IsAtleastOneFileSelected(IReadOnlyList<PackageItem> children)
         {
-            foreach (PackageItem child in children)
+            if (children != null)
             {
-                if (child.Children.Any())
+                foreach (PackageItem child in children)
                 {
-                    CheckIfFilesAreSelected(child.Children);
-
-                    if (!NoFilesSelected)
+                    if (child.Children.Count != 0)
                     {
-                        return;
+                        IsAtleastOneFileSelected(child.Children);
+
+                        if (!NoFilesSelected)
+                        {
+                            return true;
+                        }
+                    }
+
+                    if (child.IsChecked.HasValue && child.IsChecked.Value)
+                    {
+                        NoFilesSelected = false;
+                        return true;
+                    }
+                    else
+                    {
+                        NoFilesSelected = true;
                     }
                 }
-
-                if (child.IsChecked.HasValue && child.IsChecked.Value)
-                {
-                    NoFilesSelected = false;
-                    return;
-                }
-                else
-                {
-                    NoFilesSelected = true;
-                }
             }
+
+            return false;
         }
 
         private async void InstallPackageAsync()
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -321,26 +321,36 @@ private bool CanInstallPackage()
...
+            }
+
+            return !_isInstalling && !NoFilesSelected;
+        }
+
+        private void CheckIfFilesAreSelected(IReadOnlyList<PackageItem> children)
+        {
+            foreach (PackageItem child in children)
+            {
+                if (child.Children.Any())
```
ToddGrun: >Any() [](start = 35, length = 5)

No need for linq, right?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -321,26 +321,36 @@ private bool CanInstallPackage()
         {
             if (DisplayRoots != null)
             {
-                foreach (PackageItem packageItem in DisplayRoots)
+                CheckIfFilesAreSelected(DisplayRoots);
+            }
+
+            return !_isInstalling && !NoFilesSelected;
```
ToddGrun: >_isInstalling [](start = 20, length = 13)

This doesn't change during the CheckIfFilesAreSelected call, right? If so, can you do this first?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -321,26 +321,36 @@ private bool CanInstallPackage()
...
+        }
+
+        private void CheckIfFilesAreSelected(IReadOnlyList<PackageItem> children)
```
ToddGrun: >CheckIfFilesAreSelected [](start = 21, length = 23)

By name, I wouldn't have thought this method would have changed state, but instead would return a bool.

bhsubra: Renamed this method and refactored the code. I need to maintain state for NoFilesSelected as bind this in InstallDialogXaml to refresh state for the tree view for files

---
In reply to: [194555082](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194555082) [](ancestors = 194555082)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Checking with count instead.

---
In reply to: [194554437](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194554437) [](ancestors = 194554437)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Refactored this code a bit. Could you please take a look at the new iteration and let me know what you think?

---
In reply to: [194554829](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194554829) [](ancestors = 194554829)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
ToddGrun: Why not have the caller change the state if that's what's desired.

---
In reply to: [194565600](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194565600) [](ancestors = 194565600,194555082)

bhsubra: I renamed the field, the negations were confusing. Sorry

---
In reply to: [194567317](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194567317) [](ancestors = 194567317,194565600,194555082)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -311,36 +311,44 @@ private static void SetNodeOpenStates(PackageItem item)
...
+        }
 
-                    foreach (PackageItem child in children)
+        private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
+        {
+            if (children != null)
+            {
+                foreach (PackageItem child in children)
+                {
+                    if (child.Children.Count != 0)
```
ToddGrun: >if (child.Children.Count != 0) [](start = 20, length = 30)

gitting nitty now, but two things:
1) the child.Children.Count condition isn't necessary
2) I'd do the child.IsChecked.HasValue check first before recursing

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -311,36 +311,44 @@ private static void SetNodeOpenStates(PackageItem item)
...
-            get { return _noFilesSelected; }
-            set { Set(ref _noFilesSelected, value); }
+            get { return _anyFileSelected; }
+            set { Set(ref _anyFileSelected, value); }
         }
 
         private bool CanInstallPackage()
         {
-            if (DisplayRoots != null)
+            if (!_isInstalling)
```
ToddGrun: >if (!_isInstalling) [](start = 12, length = 19)

Do you think it would be cleaner to do this (the answer might well be no):

if (_isInstalling)
{
    return false;
}

AnyFileSelected = IsAnyFileSelected(DisplayRoots);
return AnyFileSelected;

bhsubra: Yes, that looks cleaner. Updated

---
In reply to: [194574761](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194574761) [](ancestors = 194574761)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
ToddGrun: I've made you suffer enough, so I'll go ahead and sign off. Just for funsies though, here's the way I would have implemented this. The advantages of this solution would be no recursion and you check the higher nodes in the tree before needing to recurse all the way down to the leafs.

private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
{
    List<PackageItem> toProcess = children.ToList();
    for(int i = 0; i < toProcess.Count; i++)
    {
        if (child.IsChecked.HasValue && child.IsChecked.Value)
        {
            return true;
        }

        toProcess.AddRange(child.Children);        
    }

    return false;
}

---
In reply to: [194574114](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194574114) [](ancestors = 194574114)

bhsubra: :)

Updated to remove recursive call. As always looks better than before. I don't know why I didnt think of that earlier.

---
In reply to: [194578712](https://github.com/aspnet/LibraryManager/pull/126#discussion_r194578712) [](ancestors = 194578712,194574114)

PR #153, opened 2018-06-12
Fix unpkg file selection issue

bhsubra:
Recursively check if files are selected


bhsubra:
Merging this as the code was reviewed in master.

PR #154, opened 2018-06-12
Setting NuGetPackageId for UnpkgProvider and filtering null NuGetPack…

raquelsa:
…ageId's from install steps.

### DIFF
--- a/src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
+++ b/src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Web.LibraryManager.Providers.Unpkg
 
         public string Id => IdText;
 
-        public string NuGetPackageId => null;
+        public string NuGetPackageId { get; } = "Microsoft.Web.LibraryManager.Build";
 
         public IHostInteraction HostInteraction { get; }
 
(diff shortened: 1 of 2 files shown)


src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -22,7 +22,7 @@ public UnpkgProvider(IHostInteraction hostInteraction)
 
         public string Id => IdText;
 
-        public string NuGetPackageId => null;
+        public string NuGetPackageId { get; } = "Microsoft.Web.LibraryManager.Build";
```
alexgav: You could've just left the =>, but it doesn't really matter.

jimmylewis: Why do providers know anything about the NuGet package for restore?  This seems like a broken design.  Should this be fixed in the command instead?

alexgav: I think this is good  Preview3 as the simplest fix, but in general I completely agree.

jimmylewis: Can't we just change the `packageIds` line in RestoreOnBuildCommand.cs to be "Microsoft.Web.LibraryManager.Build"?

jimmylewis: oh, but currently this allows a provider to specify another package... hm, maybe not then.

raquelsa: I would just leave as is for now. Not sure what was the intent of the original design. We can propose design changes later since this fix if for preview3

raquelsa: Logged issue #155 for the design review

PR #159, opened 2018-06-13
Make version list work after typing @ for library id

YangLyu89:


src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
```diff
@@ -151,10 +151,10 @@ private void OnCommitted(object sender, EventArgs e)
             }
         }
 
-        private JSONParseItem GetItemBeforePosition(int pos, JSONEditorDocument document)
+        private JSONParseItem GetItemBeforePosition(int pos, JSONComplexItem document)
```
jimmylewis: >document [](start = 77, length = 8)

This is no longer taking a document, so it should be renamed.

src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
```diff
@@ -173,7 +173,7 @@ private JSONParseItem GetItemBeforePosition(int pos, JSONEditorDocument document
                     if (item is JSONComplexItem)
                     {
```
jimmylewis: use `if (item is JSONComplexItem complexItem)` here so you don't have to cast below.

src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
YangLyu89: Will rename

src/LibraryManager.Vsix/Json/Completion/CompletionController.cs
YangLyu89: Will do

YangLyu89:
https://github.com/aspnet/LibraryManager/issues/135 is fixed in this one

Issue #150, opened 2018-06-12
Unpkg provider: no intellisense after typing colon after "library"


YangLyu89:
### Minimal repro steps
1. open libman.json file
2. in the libraries: [
{
}
]
block, type "library":

### Expected result
A list of libraries displayed

### Actual result
No list displayed ( it will have intellisense afer typing characters after the colon)

PR #165, opened 2018-06-14
make colon works for library id intellisense

YangLyu89:
This is to resolve bug https://github.com/aspnet/LibraryManager/issues/150

### DIFF
--- a/src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
+++ b/src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
@@ -18,16 +18,19 @@ namespace Microsoft.Web.LibraryManager.Providers.Unpkg
 
         public static async Task<IEnumerable<string>> GetPackageNamesAsync(string searchTerm, CancellationToken cancellationToken)
         {
+            if (searchTerm == null)
+            {
+                return new string[0];
+            }
+
             if (searchTerm.StartsWith("@"))
             {
                 return await GetPackageNamesWithScopeAsync(searchTerm, cancellationToken).ConfigureAwait(false);
             }
-            else if (!string.IsNullOrEmpty(searchTerm))
+            else
             {
                 return await GetPackageNamesFromSimpleQueryAsync(searchTerm, cancellationToken).ConfigureAwait(false);
             }
-
-            return new string[0];
         }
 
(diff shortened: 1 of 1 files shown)


src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
```diff
@@ -18,16 +18,19 @@ internal class NpmPackageSearch
 
         public static async Task<IEnumerable<string>> GetPackageNamesAsync(string searchTerm, CancellationToken cancellationToken)
         {
+            if (searchTerm == null)
```
YangLyu89: >searchTerm [](start = 16, length = 10)

This is an empty string when we just type colon

jimmylewis: String checks should usually use `string.IsNullOrEmpty` unless empty is an allowed value (but in this case, I'm guessing it's not?)

jimmylewis: hm. I guess it is actually, because below you would use an empty string for the search term.  So this is probably fine

src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
```diff
@@ -18,16 +18,19 @@ internal class NpmPackageSearch
...
+            {
+                return new string[0];
+            }
+
             if (searchTerm.StartsWith("@"))
```
jimmylewis: nit: This could be an `else if`.

src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
YangLyu89: Yes. unlike using library name as the key in Editor's npm package completion. In libman it works as a value so after typing colon, we should allow the empty string to invoke completion.

src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
YangLyu89: Changed. Is the purpose of this change to make code looks cleaner?

---
In reply to: [195601754](https://github.com/aspnet/LibraryManager/pull/165#discussion_r195601754) [](ancestors = 195601754)

YangLyu89:
This is to resolve https://github.com/aspnet/LibraryManager/issues/150

Issue #120, opened 2018-06-11
Remove Bootstrap special case alias for twitter-bootstrap


raquelsa:
### Functional impact
We currently have a special case for twitter-bootstrap and we add an alias for it called "bootstrap" for cdnjs provider,

We should honor the provider's library names for consistency.

PR #167, opened 2018-06-14
Remove bootstrap special case alias

YangLyu89:
This is to resolve https://github.com/aspnet/LibraryManager/issues/120

### DIFF
--- a/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
+++ b/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
@@ -214,16 +214,14 @@ namespace Microsoft.Web.LibraryManager.Providers.Cdnjs
 
         private static string NormalizedGroupName(string groupName)
         {
-            string cleanName = groupName;
-
-            if (cleanName.EndsWith("js"))
+            if (groupName.EndsWith("js"))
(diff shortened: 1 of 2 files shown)


src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
```diff
@@ -214,7 +214,7 @@ private IEnumerable<CdnjsLibraryGroup> GetSortedSearchResult(string term)
 
         private static string NormalizedGroupName(string groupName)
         {
-            string cleanName = AliasedName(groupName);
+            string cleanName = groupName;
```
prafullbhosale: No need for this assignment. You can directly use groupName below.

prafullbhosale:
Oh, just realised this PR is against `master` branch, which is being used for 15.8 Preview3. **Do not** merge this to master yet.

src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
YangLyu89: Will do

YangLyu89:
Close this one and merge it to the rel/preview4 branch.

PR #169, opened 2018-06-15
Fix prompts for Init and Install CLI commands

prafullbhosale:
Changes:
- During init prompt for `DefaultProvider` with a suggested default of `cdnjs` (System default)
- For the prompts, keep the console on the same line instead of forcing the user to the next line.
- During `install` if libman.json doesn't exist, changed the behavior to write the json file to disk and then to read it again, so we get hold of proper defaults to use for validation.

### DIFF
--- a/src/libman/Contracts/ConsoleLogger.cs
+++ b/src/libman/Contracts/ConsoleLogger.cs
@@ -30,13 +30,13 @@ namespace Microsoft.Web.LibraryManager.Tools.Contracts
         {
             lock(_syncObject)
             {
-                Console.Out.WriteLine(fieldName);
+                Console.Out.Write($"{fieldName}: ");
                 return Console.ReadLine();
             }
         }
 
         /// <summary>
-        /// 
+        ///
         /// </summary>
         /// <param name="fieldName"></param>
         /// <param name="defaultValue"></param>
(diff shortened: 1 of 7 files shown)


src/libman/Contracts/ConsoleLogger.cs
```diff
@@ -46,7 +46,7 @@ public string GetUserInputWithDefault(string fieldName, string defaultValue)
             lock (_syncObject)
             {
                 string message = $"{fieldName} [{defaultValue}]:";
```
ToddGrun: Should there be a space here?

ToddGrun:
>         public string GetUserInput(string fieldName)

Seems inconsistent with GetUserInputWithDefault

---
Refers to: src/libman/Contracts/ConsoleLogger.cs:29 in 17b50c6. [](commit_id = 17b50c624fa25137a808600ba7c8cd6c281f1ab5, deletion_comment = False)

src/libman/Contracts/ConsoleLogger.cs
```diff
@@ -46,7 +46,7 @@ public string GetUserInputWithDefault(string fieldName, string defaultValue)
...
-                Console.Out.WriteLine(message);
+                Console.Out.Write(message);
                 string value = Console.ReadLine();
 
                 if (string.IsNullOrEmpty(value?.Trim()))
```
ToddGrun: >string.IsNullOrEmpty(value?.Trim() [](start = 20, length = 34)

string.IsNullOrWhitespace(value)

prafullbhosale:
Updated.

PR #178, opened 2018-06-18
Removing duplicate output window meesage

raquelsa:


src/LibraryManager.Vsix/ErrorList/ErrorList.cs
```diff
@@ -52,7 +52,6 @@ public void HandleError(IError error)
         {
             Errors.Add(new DisplayError(error));
 
-            Logger.LogEvent(error.Message, LogLevel.Operation);
```
alexgav: Which other call ends writing to the output window?

raquelsa: The HandleError method is only called for the Libman.json parser errors (manifest == null). And for parser errors in the manifest we already call in Manifest.FromJson():
dependencies.GetHostInteractions().Logger.Log(PredefinedErrors.ManifestMalformed().Message, LogLevel.Task);

PR #179, opened 2018-06-18
Fix incomplete file download

alexgav:
Fix for internal bug 612818 [LibMan] Download huge file and quit VS: partial download orphaned.

The fix is to 

1. Save file stream to a temporary file first
2. Once all stream content is saved locally, move the temp file to the destination file
3. Clean up temporary file/folder

This should greatly reduce the likelihood of ending up with a  corrupt/partially downloaded file if the user exits VS before file  download is complete.

### DIFF
--- a/src/LibraryManager.Contracts/FileHelpers.cs
+++ b/src/LibraryManager.Contracts/FileHelpers.cs
@@ -17,6 +17,55 @@ namespace Microsoft.Web.LibraryManager.Contracts
     /// </summary>
     public static class FileHelpers
     {
+        /// <summary>
+        /// Writes the stream to a temporary file first, then moves the temporary file to the destination file
+        /// </summary>
+        /// <param name="fileName"></param>
+        /// <param name="sourceStream"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public static async Task<bool> SafeWriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            string tempFileName = Path.GetTempFileName();
+
+            bool result = true;
+
+            if (string.IsNullOrWhiteSpace(tempFileName))
+            {
+                Debug.Fail($"Unexpected: null or empty {nameof(tempFileName)}");
+
+                // If we are on a platform where Path.GetTempFileName() isn't implemented and returns null or empty string,
+                // fallback to just writing directly to the destination file
+                result = await WriteToFileAsync(fileName, sourceStream, cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                result = await WriteToFileAsync(tempFileName, sourceStream, cancellationToken).ConfigureAwait(false);
+
+                if (result)
+                {
+                    result = await MoveFileAsync(tempFileName, fileName);
+                }
+
+                // Clean up temp file if we didn't move it to the desination file successfully
+                if (!result)
+                {
+                    try
+                    {
+                        DeleteFiles(new string[] { tempFileName });
+                    }
+                    catch
+                    {
+                        Debug.Fail($"Could not clean up temporary file {tempFileName}");
+                        // Don't fail the operation if we couldn't clean up temporary file
+                    }
+                }
+            }
+
+            return result;
+        }
 
         /// <summary>
         /// Writes a Stream to a destination file
@@ -25,7 +74,7 @@ namespace Microsoft.Web.LibraryManager.Contracts
         /// <param name="sourceStream"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        public static async Task<bool> WriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
+        private static async Task<bool> WriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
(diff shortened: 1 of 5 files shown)


src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -17,6 +17,42 @@ namespace Microsoft.Web.LibraryManager.Contracts
...
+            bool result = await WriteToFileAsync(tempFileName, sourceStream, cancellationToken).ConfigureAwait(false);
+
+            if (result)
+            {
+                result = await MoveFileAsync(tempFileName, fileName);
+            }
+
+            try
+            {
+                DeleteFiles(new string[] { tempFileName });
```
jimmylewis: >DeleteFiles [](start = 16, length = 11)

This should only be necessary when `!result`, since the tempFile would have been moved.  Is it worth putting in a conditional block?

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -17,6 +17,42 @@ namespace Microsoft.Web.LibraryManager.Contracts
...
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public static async Task<bool> SafeWriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            string tempFileName = Path.Combine(
+                Environment.GetFolderPath(Environment.SpecialFolder.InternetCache),
+                "libman",
+                fileName.GetHashCode().ToString(),
```
jimmylewis: >GetHashCode() [](start = 25, length = 13)

Does string.GetHashCode give enough entropy to avoid conflicts?  Is it case-aware?

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -17,6 +17,42 @@ namespace Microsoft.Web.LibraryManager.Contracts
...
+                Path.GetFileName(fileName));
```
jimmylewis: I wonder if we should just use Path.GetTempFileName for now, until we nave a need for something more complicated?

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -17,6 +17,42 @@ namespace Microsoft.Web.LibraryManager.Contracts
...
+        /// <param name="fileName"></param>
+        /// <param name="sourceStream"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public static async Task<bool> SafeWriteToFileAsync(string fileName, Stream sourceStream, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            string tempFileName = Path.Combine(
+                Environment.GetFolderPath(Environment.SpecialFolder.InternetCache),
```
jimmylewis: >Environment.SpecialFolder.InternetCache [](start = 42, length = 39)

What would this be on a non-Windows box?

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: It actually also deletes empty folders. I wanted to clean up the "hash" folders we create for the file. So if the file got moved successfully, it'll delete empty folder. If not, it'll delete that temp file and folder.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: I believe it's case aware, so we should probably do ToLower on the filename. I'm using library file name in addition to the full path hash, so I believe it should be enough. Theoretically two different full paths could get hashed to the same value, in which case the file names will be different. If we want to be super-safe, I can check for an existing folder I guess. Also since we are not doing resume for now, we could just get unique temp path to eliminate the possibility of hashing to the same value. What do you think?

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: Haha, OK, I just commented on that above. OK, I'll change that.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: Not sure, but I would hope it would be a meaningful value? Do we have a linux box handy? :)

prafullbhosale: Tried this on Ubuntu, `Environment.SpecialFolder.InternetCache` just returns the string `InternetCache`.

`GetFolderPath(Environment.SpecialFolder.InternetCache)` returns an empty string.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: Made it conditional since I don't create a subfolder anymore

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: Changed to Path.GetTemporaryFiles and added a fallback to the old behavior in case if that's not implemented on some platform.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: Yep, ended up just doing that.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: I guess there is no guarnatee that Path.GetTempFileName will return correct results on all platforms either, so I added a fallback to the old behavior.

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -91,6 +140,49 @@ public static bool CopyFile(string sourceFile, string destinationFile)
...
+        /// </summary>
+        /// <param name="sourceFile"></param>
+        /// <param name="destinationFile"></param>
+        /// <returns></returns>
+        public static async Task<bool> MoveFileAsync(string sourceFile, string destinationFile)
+        {
+            bool result = true;
+
+            if (string.IsNullOrEmpty(sourceFile) || string.IsNullOrEmpty(destinationFile) || !File.Exists(sourceFile))
+            {
```
jimmylewis: these seem like appropriate cases to throw ArgumentNullException, ArgumentException, or FileNotFoundException

jimmylewis: I guess we're not doing that elsewhere, so we should be consistent for now.

---
In reply to: [197862456](https://github.com/aspnet/LibraryManager/pull/179#discussion_r197862456) [](ancestors = 197862456)

PR #180, opened 2018-06-18
Add name and version to ILibraryInsallationState

prafullbhosale:
- Adds `Name` and `Version` properties to `ILibraryInstallationState`
- Adds a naming scheme which splits `LibraryId` into `Name` and `Version` (and combines them back) using `@` as the split character

### DIFF
--- a/src/LibraryManager/Helpers/LibraryNamingScheme.cs
+++ b/src/LibraryManager/Helpers/LibraryNamingScheme.cs
@@ -1,0 +1,71 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Web.LibraryManager.Helpers
+{
+    /// <summary>
+    /// Defines the way LibraryId is split into name and Version
+    /// </summary>
+    public class LibraryNamingScheme
+    {
+        private const char _separator = '@';
+        private LibraryNamingScheme() { }
+
+        /// <summary>
+        /// Singleton instance of the LibraryNamingScheme
+        /// </summary>
+        public static LibraryNamingScheme Instance { get; } = new LibraryNamingScheme();
+
+        /// <summary>
+        /// Splits libraryId into name and version using '@' as the split char.
+        /// Only the last appearance of '@' is used to split the libraryId.
+        /// Note: If libraryId starts with '@' and has no other occurences of '@',
+        /// the entire libraryId is considered to be the name and version is considered to be
+        /// empty.
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="name"></param>
+        /// <param name="version"></param>
+        /// <returns></returns>
+        public void GetLibraryNameAndVersion(string libraryId, out string name, out string version)
+        {
+            if (string.IsNullOrEmpty(libraryId))
+            {
+                name = string.Empty; 
+                version = string.Empty;
+
+                return;
+            }
+
+            int indexOfAt = libraryId.LastIndexOf(_separator);
+
+            name = libraryId;
+            version = string.Empty;
+
+            if (indexOfAt > 0 && indexOfAt < libraryId.Trim().Length -1)
+            {
+                name = libraryId.Substring(0, indexOfAt);
+                version = libraryId.Substring(indexOfAt+1);
+            }
+        }
+
+        /// <summary>
+        /// Generates a libraryId by concatenating name and version.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="version"></param>
+        /// <returns></returns>
+        public string GetLibraryId(string name, string version)
+        {
+            if (string.IsNullOrWhiteSpace(name) && string.IsNullOrWhiteSpace(version))
+            {
+                return string.Empty;
+            }
+
+            return string.IsNullOrWhiteSpace(version)
+                ? name
+                : $"{name}{_separator}{version}";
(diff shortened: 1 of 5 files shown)


src/LibraryManager/Helpers/LibraryNamingScheme.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// Only the last appearance of '@' is used to split the libraryId.
+        /// Note: If libraryId starts with '@' and has no other occurences of '@',
+        /// the entire libraryId is considered to be the name and version is considered to be
+        /// empty.
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="name"></param>
+        /// <param name="version"></param>
+        /// <returns></returns>
+        public void GetLibraryNameAndVersion(string libraryId, out string name, out string version)
```
jimmylewis: why not return a tuple?

test/LibraryManager.Test/LibraryNamingSchemeTest.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        [DataTestMethod]
+        [DataRow("jquery@3.3.1", "jquery", "3.3.1")]
+        [DataRow("@angular/cli@1.0.0", "@angular/cli","1.0.0")]
+        [DataRow("My@Random@Library@1.0.0-preview3-final", "My@Random@Library", "1.0.0-preview3-final")]
+        [DataRow("@MyLibraryWithoutVersion", "@MyLibraryWithoutVersion", "")]
+        [DataRow("Library@Version", "Library", "Version")]
+        [DataRow(null, "", "")]
+        [DataRow("", "", "")]
+        public void GetLibraryNameAndVersion(string libraryId, string expectedName, string expectedVersion)
```
jimmylewis: What about `[DataRow("PartialId@", "PartialId", "")]` or similar to cover the case when it's only partially specified?  How should this handle the `@`?

test/LibraryManager.Test/LibraryNamingSchemeTest.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        [DataTestMethod]
+        [DataRow("jquery", "3.3.1", "jquery@3.3.1")]
+        [DataRow("@angular/cli","1.0.0", "@angular/cli@1.0.0")]
+        [DataRow("My@Random@Library", "1.0.0-preview3-final", "My@Random@Library@1.0.0-preview3-final")]
+        [DataRow("@MyLibraryWithoutVersion", "", "@MyLibraryWithoutVersion")]
+        [DataRow("Library", "Version", "Library@Version")]
+        [DataRow("", "", "")]
+        public void GetLibraryId(string name, string version, string expectedLibraryId)
```
jimmylewis: `[DataRow(null, null, "")]`?

test/LibraryManager.Mocks/LibraryInstallationState.cs
```diff
@@ -31,5 +31,15 @@ public class LibraryInstallationState : ILibraryInstallationState
         /// The path relative to the working directory to copy the files to.
         /// </summary>
         public virtual string DestinationPath { get; set; }
+
```
jimmylewis: should LibraryId be reimplemented as Name@Version?  Right now the mock can set all 3 of these independently, which could be a little confusing if it ever comes up.

test/LibraryManager.Test/LibraryNamingSchemeTest.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            string libraryId = LibraryNamingScheme.Instance.GetLibraryId(name, version);
+
+            Assert.AreEqual(expectedLibraryId, libraryId);
+        }
+    }
+}
```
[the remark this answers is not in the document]
prafullbhosale: > (and why is VS not doing this for you from the .editorconfig?)

probably because I am using Vs Code

test/LibraryManager.Test/LibraryNamingSchemeTest.cs
prafullbhosale: Right now if libraryId is `Partial@` it will return name as `Partial@` and Version as an empty string. 
That is in order to make sure we persist the `@` when writing libraryId back to the libman.json file.
Always adding an `@` while writing to file causes issues when the libraryId was just `Partial` to begin with.

test/LibraryManager.Test/LibraryNamingSchemeTest.cs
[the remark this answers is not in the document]
jimmylewis: Fair enough, please install the editorconfig extension for VS code. 😜

test/LibraryManager.Test/LibraryNamingSchemeTest.cs
jimmylewis: I see, that makes sense.  Can we have a test that verifies that behavior?  This might not be the right place to enforce it, but it sounds like it's a subtle scenario we wouldn't want to regress.

test/LibraryManager.Mocks/LibraryInstallationState.cs
prafullbhosale: Looks like a lot of the current tests set the `LibraryId` directly in the mock. 
Going to leave it as is for now. When we get to removing usage from libraryId, we should hit these test failures and fix them then.

prafullbhosale:
Will do a rebase and merge once the tests pass.

PR #183, opened 2018-06-19
Refactor helpers into classes under a consolidated hierarchy

jimmylewis:
The goal is to create a simpler API for discovering test utilities through the Helpers property, instead of having to reference static utilities.


jimmylewis:
@YangLyu89 this is what I was envisioning for refactoring the helpers, so that they are consistently organized and easily discoverable to test authors.  

@abpiskunov, is this the sort of pattern you were thinking of for WTE as well?  I'm trying to be consistent across our repos if possible.

abpiskunov:
Yes. Exactly like that. 👍

PR #184, opened 2018-06-19
Do not make "Loading..." getting inserted when commited

YangLyu89:
Do not make "Loading..." getting inserted when committed (enter or tab) during the time Loading for completion list.

### DIFF
--- a/src/LibraryManager.Vsix/Json/Completion/SimpleCompletionEntry.cs
+++ b/src/LibraryManager.Vsix/Json/Completion/SimpleCompletionEntry.cs
@@ -24,6 +24,11 @@ namespace Microsoft.Web.LibraryManager.Vsix
         {
         }
 
+        public SimpleCompletionEntry(string text, string insertionText, ImageMoniker moniker, IIntellisenseSession session)
+            : base(text, insertionText, null, null, null, false, session as ICompletionSession)
+        {
+        }
+
         public SimpleCompletionEntry(string text, string insertionText, string description, ImageMoniker moniker, IIntellisenseSession session, int specificVersion = 0)
             : base(text, "\"" + insertionText + "\"", description, null, null, false, session as ICompletionSession)
         {
--- a/src/LibraryManager.Vsix/Json/Completion/SimpleCompletionEntry.cs
+++ b/src/LibraryManager.Vsix/Json/Completion/SimpleCompletionEntry.cs
@@ -24,28 +24,28 @@ namespace Microsoft.Web.LibraryManager.Vsix
         {
         }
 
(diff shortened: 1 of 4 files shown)


src/LibraryManager.Vsix/Json/Completion/SimpleCompletionEntry.cs
```diff
@@ -24,6 +24,11 @@ public SimpleCompletionEntry(string text, ImageMoniker moniker, IIntellisenseSes
         {
         }
 
+        public SimpleCompletionEntry(string text, string insertionText, ImageMoniker moniker, IIntellisenseSession session)
```
jimmylewis: name this `displayText`, as that's how we distinguish in the completion APIs

jimmylewis: We should fixup the other methods where we have separate display/insertion text parameters as well.

---
In reply to: [196958553](https://github.com/aspnet/LibraryManager/pull/184#discussion_r196958553) [](ancestors = 196958553)

test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
```diff
@@ -157,5 +157,22 @@ public void LibmanCompletion_LibraryForUnpkg()
...
+            _libManConfig.Open();
+            string expectedText = "      \"library\": ";
+
+            Editor.Caret.MoveToExpression("\"libraries\"");
+            Editor.Caret.MoveDown(2);
+            Editor.KeyboardCommands.Type("\"provider\": \"cdnjs\",");
+            Editor.KeyboardCommands.Enter();
+            Editor.KeyboardCommands.Type("\"library\":");
+            Editor.KeyboardCommands.Type("j\t");
+
```
jimmylewis: If you run this with other tests, do you hit Loading here or does it go straight to cached data.

test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
```diff
@@ -157,5 +157,22 @@ public void LibmanCompletion_LibraryForUnpkg()
             Editor.KeyboardCommands.Type("\"library\":");
             LibmanTestsUtility.WaitForCompletionEntries(Editor, expectedCompletionEntries, caseInsensitive: true, timeout: 5000);
         }
+
+        [TestMethod]
+        public void LibmanCompletion_LoadingNotInserted()
```
jimmylewis: >LibmanCompletion_LoadingNotInserted [](start = 20, length = 35)

Let's just remove this test because it is timing/order dependent.

PR #185, opened 2018-06-19
Decompose tuple instead of using an instance

jimmylewis:


jimmylewis:
This style reads more naturally to me.  What do you think?

PR #219, opened 2018-06-21
[LibMan] Do not support multiple entries of the same library (of any version)

raquelsa:


raquelsa:
Merging to wrong branch

PR #224, opened 2018-06-21
Change LibraryResolver to resolve using LibraryId or LibraryName

prafullbhosale:
- `LibraryResolver` No longer uses Catalogs to resolve names to Ids.

Earlier `LibraryResolver` would try to ask each catalog to check if the given term matched display name or library. 
Now since we have information about name and version of each library in the manifest, we no longer need to ask the Catalog.

### DIFF
--- a/src/libman/Commands/UninstallCommand.cs
+++ b/src/libman/Commands/UninstallCommand.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Web.LibraryManager.Tools.Commands
         {
             Manifest manifest = await GetManifestAsync();
 
-            IEnumerable<ILibraryInstallationState> installedLibraries = await ValidateParametersAndGetLibrariesToUninstallAsync(manifest, CancellationToken.None);
+            IEnumerable<ILibraryInstallationState> installedLibraries = ValidateParametersAndGetLibrariesToUninstall(manifest);
 
             if (installedLibraries == null || !installedLibraries.Any())
             {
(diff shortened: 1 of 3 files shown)


src/libman/LibraryResolver.cs
```diff
@@ -21,16 +21,12 @@ internal static class LibraryResolver
         /// </summary>
         /// <param name="partialName">Can be display name or library id.</param>
         /// <param name="manifest"></param>
-        /// <param name="manifestDependencies"></param>
         /// <param name="provider"></param>
-        /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        public static async Task<IReadOnlyList<ILibraryInstallationState>> ResolveAsync(
+        public static IReadOnlyList<ILibraryInstallationState> Resolve(
```
raquelsa: Can we use the SearchAsync on the Catalogs for that?

src/libman/Commands/UninstallCommand.cs
```diff
@@ -116,11 +115,9 @@ public override BaseCommand Configure(CommandLineApplication parent)
                 throw new InvalidOperationException(string.Join(Environment.NewLine, errors));
             }
```
raquelsa: Isn't all this validation supposed to be at Manifest? I believe we have it all covered already

src/libman/LibraryResolver.cs
prafullbhosale: No, we don't need to, now that library name and version are already split. 
And also `SearchAsync` makes network calls.

src/libman/Commands/UninstallCommand.cs
prafullbhosale: This are command line parameter validations.

PR #246, opened 2018-06-22
Write to buffer directly if libman.json file is open

bhsubra:
Changes to write to buffer directly if libman.json file is open.

Its hard to mock textView. So I didn't attempt to add tests. I added manual walthrough instead: https://github.com/aspnet/Tooling-ManualTests/pull/117

### DIFF
--- a/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
+++ b/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
@@ -464,7 +464,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Models
                             insertionText = newLibrary + Environment.NewLine;
                         }
 
-                        int insertionIndex = FindInsertionIndex(textBuffer);
+                        int insertionIndex = libraries.AfterEnd - 1;
 
                         if (insertionIndex > 0)
                         {
@@ -526,20 +526,6 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Models
             }
 
             return null;
-        }
-
-        private int FindInsertionIndex(ITextBuffer textBuffer)
-        {
-            int insertionIndex = -1;
-
-            JSONMember libraries = GetLibraries(textBuffer);
-
-            if (libraries != null)
-            {
-                insertionIndex = libraries.AfterEnd - 1;
-            }
-
-            return insertionIndex;
         }
 
         private string GetLibraryToBeInserted(LibraryInstallationState libraryInstallationState)
--- a/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
+++ b/src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
@@ -4,10 +4,24 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Windows;
 using System.Windows.Input;
 using System.Windows.Threading;
+using Microsoft.JSON.Core.Parser.TreeItems;
+using Microsoft.JSON.Editor.Document;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Editor;
+using Microsoft.VisualStudio.OLE.Interop;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+using Microsoft.Web.Editor.Undo;
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.Web.LibraryManager.Vsix.Resources;
+using Newtonsoft.Json;
+using IVsTextBuffer = Microsoft.VisualStudio.TextManager.Interop.IVsTextBuffer;
+using RunningDocumentTable = Microsoft.VisualStudio.Shell.RunningDocumentTable;
+using Shell = Microsoft.VisualStudio.Shell;
 
 namespace Microsoft.Web.LibraryManager.Vsix.UI.Models
 {
@@ -30,6 +44,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Models
         private FileSelectionType _fileSelectionType;
         private bool _anyFileSelected;
         private bool _isTreeViewEmpty;
+        private const string _commaDelimiter = ",";
 
         public InstallDialogViewModel(Dispatcher dispatcher, ILibraryCommandService libraryCommandService, string configFileName, IDependencies deps, string targetPath, Action<bool> closeDialog)
         {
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
                     LibraryId = PackageId,
                     ProviderId = selectedPackage.ProviderId,
                     DestinationPath = InstallationFolder.DestinationFolder,
                     Files = SelectedFiles.ToList()
-                });
-
-                await manifest.SaveAsync(_configFileName, CancellationToken.None).ConfigureAwait(false);
+                };
 
                 EnvDTE.Project project = VsHelpers.DTE.SelectedItems.Item(1)?.ProjectItem?.ContainingProject;
```
ToddGrun: >EnvDTE.Project project = VsHelpers.DTE.SelectedItems.Item(1)?.ProjectItem?.ContainingProject; [](start = 16, length = 93)

I didn't know DTE access was thread safe

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
                 project?.AddFileToProjectAsync(_configFileName);
 
-                await _libraryCommandService.RestoreAsync(_configFileName, CancellationToken.None).ConfigureAwait(false);
+                await Shell.ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+                RunningDocumentTable rdt = new RunningDocumentTable(Shell.ServiceProvider.GlobalProvider);
+                IVsTextBuffer document = rdt.FindDocument(Path.GetFullPath(_configFileName)) as IVsTextBuffer;
```
ToddGrun: >Path.GetFullPath(_configFileName) [](start = 58, length = 33)

used twice, put into local?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+            Keyboard.Focus(editorWindow);
+        }
+
+        private JSONMember GetLibraries(ITextBuffer textBuffer)
+        {
+            JSONEditorDocument document = JSONEditorDocument.FromTextBuffer(textBuffer);
+
+            if (document != null)
+            {
+                IEnumerable<JSONMember> jsonMembers = document.JSONDocument.TopLevelValue?.FindType<JSONObject>().Children?.OfType<JSONMember>();
```
ToddGrun: >. [](start = 113, length = 1)

This doesn't need to be checked?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+
+                if (jsonMembers.Any())
```
ToddGrun: >if (jsonMembers.Any()) [](start = 16, length = 22)

not really necessary

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+
+            if (textBuffer != null)
+            {
+                string insertionText;
+
+                JSONMember libraries = GetLibraries(textBuffer);
+
+                if (libraries != null && libraries.Children != null)
+                {
+                    JSONArray jsonArray = libraries.Children.OfType<JSONArray>().First();
```
ToddGrun: >First [](start = 81, length = 5)

Was this supposed to be FirstOrDefault or is the if(jsonArray != null) not necessary?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+
+                    if (jsonArray != null)
+                    {
+                        string newLibrary = GetLibraryToBeInserted(libraryInstallationState);
+                        bool containsLibrary = jsonArray.BlockItemChildren.Count() > 0;
```
ToddGrun: >Count() [](start = 75, length = 7)

This is Linq, right? I'd just use .Any instead of this as there's no need to walk the whole enumeration

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+
+                        if (containsLibrary)
+                        {
+                            insertionText = _commaDelimiter + Environment.NewLine + newLibrary + Environment.NewLine;
```
ToddGrun: >Environment.NewLine [](start = 62, length = 19)

I wonder if you should be getting the newline from the buffer?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+                        }
+                        else
+                        {
+                            insertionText = newLibrary + Environment.NewLine;
+                        }
+
+                        int insertionIndex = FindInsertionIndex(textBuffer);
```
ToddGrun: >FindInsertionIndex [](start = 45, length = 18)

Can you just use 

insertionIndex = libraries.AfterEnd - 1;

since you have it here instead of calculating liibraries again in that method?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +388,186 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+            compoundUndoAction.Open(Text.FormatUndoAction);
+
+            using (ITextEdit textEdit = textBuffer.CreateEdit())
+            {
+                textEdit.Insert(insertionIndex, insertionText);
+                textEdit.Apply();
+
+                IOleCommandTarget commandTarget = Shell.Package.GetGlobalService(typeof(Shell.Interop.SUIHostCommandDispatcher)) as IOleCommandTarget;
+
+                if (commandTarget != null)
```
ToddGrun: >if (commandTarget != null) [](start = 16, length = 26)

I could go either way, but there probably isn't really a need to check this point. If it's null, VS is toast.

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: We won't hit this code path if manifest file had syntax errors. E.g: if someone manually deleted the contents of libman.json and then tried to add libraries from UI, we would fail much early in the stack. Let me know if null check here is necessary.

---
In reply to: [197855007](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197855007) [](ancestors = 197855007)

ToddGrun: OK, as long as you know there is an object in there, then no check is necessary.

---
In reply to: [197922002](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197922002) [](ancestors = 197922002,197855007)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Currently we only have one JSONArray for libraries and its a required property. So First should be fine. Null check for jsonArray might not be required.

---
In reply to: [197855633](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197855633) [](ancestors = 197855633)

ToddGrun: Yeah, the null check doesn't do anything here as the .First() call will throw if there isn't anything

---
In reply to: [197926691](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197926691) [](ancestors = 197926691,197855633)

bhsubra: Updated in version 8.

---
In reply to: [197927099](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197927099) [](ancestors = 197927099,197926691,197855633)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +387,160 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+                    manifest.AddLibrary(libraryInstallationState);
+
+                    await manifest.SaveAsync(_configFileName, CancellationToken.None).ConfigureAwait(false);
+
+                    await _libraryCommandService.RestoreAsync(_configFileName, CancellationToken.None).ConfigureAwait(false);
+
+                    _dispatcher.Invoke(() =>
+                    {
+                        _closeDialog(true);
+                    });
```
ToddGrun: Does this have to be after awaits in this case, or could you just always do this before the if?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +387,160 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
 
                 EnvDTE.Project project = VsHelpers.DTE.SelectedItems.Item(1)?.ProjectItem?.ContainingProject;
                 project?.AddFileToProjectAsync(_configFileName);
 
-                await _libraryCommandService.RestoreAsync(_configFileName, CancellationToken.None).ConfigureAwait(false);
+                RunningDocumentTable rdt = new RunningDocumentTable(Shell.ServiceProvider.GlobalProvider);
+                string configFilePath = Path.GetFullPath(_configFileName);
 
-                _dispatcher.Invoke(() =>
+                IVsTextBuffer document = rdt.FindDocument(configFilePath) as IVsTextBuffer;
```
ToddGrun: >document [](start = 30, length = 8)

can you rename this to better reflect the type?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +387,160 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+
+        private JSONMember GetLibraries(ITextBuffer textBuffer)
+        {
+            JSONEditorDocument document = JSONEditorDocument.FromTextBuffer(textBuffer);
+
+            if (document != null)
+            {
+                IEnumerable<JSONMember> jsonMembers = document.JSONDocument.TopLevelValue?.FindType<JSONObject>().Children?.OfType<JSONMember>();
+
+                return jsonMembers.FirstOrDefault(m => m.UnquotedNameText == ManifestConstants.Libraries);
```
ToddGrun: >jsonMembers [](start = 23, length = 11)

Can this be null?

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +387,160 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+
+                commandTarget.Exec(
+                    ref guidVSStd2K,
+                    (uint)VSConstants.VSStd2KCmdID.FORMATSELECTION,
+                    (uint)OLECMDEXECOPT.OLECMDEXECOPT_DODEFAULT,
+                    IntPtr.Zero,
+                    IntPtr.Zero);
+
+                textView.Selection.Clear();
+            }
```
ToddGrun: >} [](start = 12, length = 1)

Does disposing after all the format selection make this a single edit? (IE, I'm trying to understand why the using block isn't closed earlier)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -372,29 +387,160 @@ private static bool IsAnyFileSelected(IReadOnlyList<PackageItem> children)
...
+        {
+            ITextBuffer textBuffer = GetTextBuffer(document);
+
+            if (textBuffer != null)
+            {
+                string insertionText;
+
+                JSONMember libraries = GetLibraries(textBuffer);
+
+                if (libraries != null && libraries.Children != null)
```
ToddGrun: >libraries != null && libraries.Children != null [](start = 20, length = 47)

nit: 

Could be 

if (libraries?.Children != null)

and insertionText declaration could be moved inside

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
```diff
@@ -30,6 +44,7 @@ internal class InstallDialogViewModel : BindableBase
         private FileSelectionType _fileSelectionType;
         private bool _anyFileSelected;
         private bool _isTreeViewEmpty;
+        private const string _commaDelimiter = ",";
```
ToddGrun: >_commaDelimiter [](start = 29, length = 15)

nit: As this is only used in one location and doesn't change, I wouldn't have bothered with this

bhsubra: Updated in version 10.

---
In reply to: [197934076](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197934076) [](ancestors = 197934076)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Added logic to get line break text from previous line in version 10.

---
In reply to: [197856178](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197856178) [](ancestors = 197856178)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Sure, renamed it to textBuffer in version 10.

---
In reply to: [197931858](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197931858) [](ancestors = 197931858)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Updated in version 10.

---
In reply to: [197933718](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197933718) [](ancestors = 197933718)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Actually it can be outside. Moved it in iteration 10.

---
In reply to: [197931544](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197931544) [](ancestors = 197931544)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: No, we wont hit this code path if it is null. If the manifest file has syntax errors we will error out much earlier in the stack.

---
In reply to: [197932421](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197932421) [](ancestors = 197932421)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: The using can be closed earlier. I see two entries in undo stack:

Format 
TextBufferChange

Updated for version 10

---
In reply to: [197933019](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197933019) [](ancestors = 197933019)

src/LibraryManager.Vsix/UI/Models/InstallDialogViewModel.cs
bhsubra: Removed all ?. None of those should be null at this point actually. Updated in version 11.

---
In reply to: [197947861](https://github.com/aspnet/LibraryManager/pull/246#discussion_r197947861) [](ancestors = 197947861,197942995,197932421)

Issue #213, opened 2018-06-21
[Add Files Dialog] IntelliSense not selecting item typed.


justcla:
As a result, tabbing always inserts the first item in the list.
	- Library: type "jquery@2.2.1" [Tab]
		○ You end up with jquery@3.3.1

YangLyu89:
I cannot repro this one. After typing "jquery@2.2.1", there is only one item in the list and it's jquery@2.2.1

PR #247, opened 2018-06-23
Making UI Work correctly for intellisense of item typed

YangLyu89:


src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
```diff
@@ -204,6 +205,7 @@ internal void RefreshSearch()
 
             string text = Text;
             int caretIndex = text.Length;
+            int at = text.IndexOf('@');
```
raquelsa: More descriptive variable name?

YangLyu89: atIndex probably?

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
```diff
@@ -241,7 +249,8 @@ internal void RefreshSearch()
 
                             if (Items != null && Items.Count > 0 && Options.SelectedIndex == -1)
                             {
-                                SelectedItem = Items.FirstOrDefault(x => x.CompletionItem.InsertionText == lastSelected) ?? Items[0];
+                                SelectedItem = ((at >= 0) ? Items.FirstOrDefault(x => x.CompletionItem.DisplayText.StartsWith(text.Substring(at + 1))) : Items.FirstOrDefault(x => x.CompletionItem.InsertionText == lastSelected)) ?? Items[0];
```
alexgav: This is really hard to read. If-then-else would make this more readable,

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
```diff
@@ -257,5 +266,10 @@ private void ThisControl_GotFocus(object sender, RoutedEventArgs e)
                 LibrarySearchBox.Focus();
             }
         }
+
+        private void FilterOutUnmatchedItems(ref IEnumerable<CompletionItem>items, string versionPrefix)
```
alexgav: Nit: versionSuffix.

Also, why not just return the filtered enum instead of passing in original by ref?

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
```diff
@@ -257,5 +266,10 @@ private void ThisControl_GotFocus(object sender, RoutedEventArgs e)
...
+        {
+            items = items.Where(x => x.DisplayText.Contains(versionPrefix));
```
raquelsa: Is completion Item for version only? If so, should we filter by x.DisplayText.StartsWith?

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
raquelsa: yes

YangLyu89:
This is to resolve https://github.com/aspnet/LibraryManager/issues/213

PR #248, opened 2018-06-25
Do not list all files when include all files option is checked

bhsubra:
Change to not include all files when include all files option is checked in the UI.

### DIFF
--- a/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
@@ -198,12 +198,6 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI
 
         private void InstallButton_Clicked(object sender, RoutedEventArgs e)
         {
-            // When "Include all files" option is checked, we don't want to write out the files to libman.json
-            if (IncludeAllLibraryFilesRb.IsChecked.Value)
-            {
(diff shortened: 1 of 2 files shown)


src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
```diff
@@ -198,6 +198,12 @@ private void ProviderComboBox_SelectionChanged(object sender, SelectionChangedEv
 
         private void InstallButton_Clicked(object sender, RoutedEventArgs e)
         {
+            // When "Include all files" option is checked, we don't want to write out the files to libman.json
+            if (IncludeAllLibraryFilesRb.IsChecked.Value)
```
ToddGrun: >IncludeAllLibraryFilesRb.IsChecked.Value [](start = 16, length = 40)

Issues discussed in person:
1) What to do if all selected and checkbox not selected
2) Why files: [] instead of not having it?
3) Sort order seems wonky in dialog

bhsubra: As discussed offline, leaving [1] as is.
Fixed 2 in iteration 2.

Justin/me will log bug for 3.

---
In reply to: [197885065](https://github.com/aspnet/LibraryManager/pull/248#discussion_r197885065) [](ancestors = 197885065)

bhsubra:
Manually merging this to master.

PR #261, opened 2018-06-27
Added VSSDK analyzer to VSIX project

justcla:
Provides helpful tips about async package load and other VSSDK issues.

### DIFF
--- a/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
+++ b/src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
@@ -203,6 +203,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime" />
     <PackageReference Include="Microsoft.VisualStudio.Language.Intellisense" />
     <PackageReference Include="Microsoft.VisualStudio.SDK.EmbedInteropTypes" />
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -207,6 +207,9 @@
     <PackageReference Include="Microsoft.VisualStudio.Imaging" />
     <PackageReference Include="Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime" />
     <PackageReference Include="Microsoft.VisualStudio.Language.Intellisense" />
+    <PackageReference Include="Microsoft.VisualStudio.SDK.Analyzers">
```
jimmylewis: The version should not be specified here.  There's a .targets file where we specify all package versions.

Also, should this not use `PrivateAssets="all"`?

src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -460,4 +463,4 @@
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
+</Project>
```
jimmylewis: nit: why did you remove the EOL at EOF?

Issue #260, opened 2018-06-27
Upgrading from 1.0.20 to 1.0.75 broke build. Failed to find RestoreTask.


BobbyCannon:
### Functional impact

Build breaks with the following error message. 

> The "Microsoft.Web.LibraryManager.Build.RestoreTask" task could not be loaded from the assembly C:\Users\bobby\.nuget\packages\microsoft.web.librarymanager.build\1.0.75\build\..\tools\net461\Microsoft.Web.LibraryManager.Build.dll. 

Full message details included in "Further technical details"

### Minimal repro steps
1. Pull Speedy repository - https://github.com/BobbyCannon/Speedy.git
1. Upgrade from 1.0.20 to 1.0.75 for Microsoft.Web.LibraryManager.Build
1. Build and it will error.

### Expected result
Build should continue to work.

### Actual result
Build breaks?

### Further technical details

C:\Users\bobby\.nuget\packages\microsoft.web.librarymanager.build\1.0.75\build\Microsoft.Web.LibraryManager.Build.targets(35,9): error MSB4062: The "Microsoft.Web.LibraryManager.Build.RestoreTask" task could not be loaded from the assembly C:\Users\bobby\.nuget\packages\microsoft.web.librarymanager.build\1.0.75\build\..\tools\net461\Microsoft.Web.LibraryManager.Build.dll.  Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [C:\Workspaces\GitHub\Speedy\Speedy.Website\Speedy.Website.csproj]
  Successfully created package 'C:\Workspaces\GitHub\Speedy\Speedy\bin\Release\Speedy.5.0.0.nupkg'.
  Speedy.Samples -> C:\Workspaces\GitHub\Speedy\Speedy.Samples\bin\Release\netstandard2.0\Speedy.Samples.dll
  Speedy.Tests -> C:\Workspaces\GitHub\Speedy\Speedy.Tests\bin\Release\netcoreapp2.1\Speedy.Tests.dll
  Speedy.Samples.Tests -> C:\Workspaces\GitHub\Speedy\Speedy.Samples.Tests\bin\Release\netcoreapp2.1\Speedy.Samples.Tests.dll

nickalbrecht:
[...]

PR #262, opened 2018-06-27
Making RestoreTask public

raquelsa:
Fixing issue @ #260


jimmylewis:
Can you add a test for this scenario?  We've shipped it broken twice already...

Issue #223, opened 2018-06-21
[File System Provider] Typing in non-existing path results in an Exception/ActivityLog dialog


alexgav:
### Functional impact
*Ugly error dialog while typing*  

### Minimal repro steps

1. Open or add libman.json
2. Add new library with provider "filesystem"
3. Add "library" field and start typing ```\\foo``` as the value of the field

### Expected result
*No error dialogs, but and error message in error list and/or output window about invalid folder*  

### Actual result
*
```
---------------------------
Microsoft Visual Studio
---------------------------
An exception has been encountered. This may be caused by an extension.

You can get more information by examining the file 'C:\Users\alexgav\AppData\Roaming\Microsoft\VisualStudio\15.0_ee8648cd\ActivityLog.xml'.
---------------------------
OK   
---------------------------
```
*

### Further technical details
*Callstack below*
```
Microsoft.Web.LibraryManager.Contracts.InvalidLibraryException: The "\\f" library could not be resolved by the "filesystem" provider   at Microsoft.Web.LibraryManager.Providers.FileSystem.FileSystemCatalog.GetLibraryCompletionSetAsync(String value, Int32 caretPosition)   at Microsoft.Web.LibraryManager.Vsix.LibraryIdCompletionProvider.<GetEntries>d__4.MoveNext()   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)   at Microsoft.JSON.Editor.Completion.JSONCompletionSource.<>c__DisplayClass3_2.<AugmentCompletionSession>b__0()   at Microsoft.Web.Editor.Utility.GuardedOperations.InvokeExtensionPoint(Object errorSource, Action action) --- End of stack trace from previous location where exception was thrown ---    at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject
```

YangLyu89:
Looks like needing handle the invalid Directory form

PR #273, opened 2018-06-29
Dev/yalyu/filesystem library exception

YangLyu89:
Allow the user to type invalid path but we will return an empty completion set.
Resolve https://github.com/aspnet/LibraryManager/issues/223

### DIFF
--- a/src/LibraryManager/Providers/FileSystem/FileSystemCatalog.cs
+++ b/src/LibraryManager/Providers/FileSystem/FileSystemCatalog.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Security;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/LibraryManager/Providers/FileSystem/FileSystemCatalog.cs
```diff
@@ -96,8 +96,20 @@ public Task<CompletionSet> GetLibraryCompletionSetAsync(string value, int caretP
 
                 return Task.FromResult(set);
             }
-            catch
+            catch (Exception ex)
```
jimmylewis: >catch [](start = 12, length = 5)

catch(ArgumentException)
{
   .... 
}
catch
{
    ....
}

YangLyu89: Will do.

PR #283, opened 2018-07-05
Fix parsing of libraryId into name and version.

prafullbhosale:
- Introduce two naming schemes: Versioned and Simple
- Providers specify version support
- Based on provider either versioned or simple naming scheme is used

### DIFF
--- a/src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
+++ b/src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
@@ -1,0 +1,88 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Web.LibraryManager.Contracts;
+
+namespace Microsoft.Web.LibraryManager.LibraryNaming
+{
+    /// <summary>
+    /// Defines the way LibraryId is split into name and Version
+    /// </summary>
+    public class LibraryIdToNameAndVersionConverter
+    {
+        /// <summary>
+        /// Singleton instance for the <see cref="LibraryIdToNameAndVersionConverter" />
+        /// </summary>
+        public static LibraryIdToNameAndVersionConverter Instance { get; } = new LibraryIdToNameAndVersionConverter();
+
+        private LibraryIdToNameAndVersionConverter()
+        {
+            _versionedNamingScheme = new VersionedLibraryNamingScheme();
+            _simpleNamingScheme = new SimpleLibraryNamingScheme();
+            _default = _simpleNamingScheme;
+        }
+
+        private readonly object _syncObject = new object();
+        private readonly ILibraryNamingScheme _versionedNamingScheme = new VersionedLibraryNamingScheme();
+        private readonly ILibraryNamingScheme _simpleNamingScheme = new SimpleLibraryNamingScheme();
+        private readonly ILibraryNamingScheme _default;
+
+
+        private IDependencies _dependencies;
+        private Dictionary<string, ILibraryNamingScheme> _perProviderNamingScheme = new Dictionary<string, ILibraryNamingScheme>();
+
+        /// <summary>
+        /// Initializes the LibraryNaming Schemes for the manifest.
+        /// </summary>
+        /// <param name="dependencies"></param>
+        public void Initialize (IDependencies dependencies)
+        {
+            _dependencies = dependencies ?? throw new ArgumentNullException(nameof(dependencies));
+
+            lock(_syncObject)
+            {
+                _perProviderNamingScheme.Clear();
+
+                foreach(IProvider p in _dependencies.Providers)
+                {
+                    _perProviderNamingScheme[p.Id] = p.SupportsLibraryVersions ? _versionedNamingScheme : _simpleNamingScheme;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Splits the libraryId into name and version based on the provider.
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="providerId"></param>
+        /// <returns></returns>
+        public (string Name, string Version) GetLibraryNameAndVersion(string libraryId, string providerId)
+        {
+            if (!string.IsNullOrEmpty(providerId) && _perProviderNamingScheme.TryGetValue(providerId, out ILibraryNamingScheme _scheme))
+            {
+                return _scheme.GetLibraryNameAndVersion(libraryId);
+            }
+
+            return _default.GetLibraryNameAndVersion(libraryId);
+        }
+
+        /// <summary>
+        /// Gets the libraryId given the name and version, based on the provider.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="version"></param>
+        /// <param name="providerId"></param>
+        /// <returns></returns>
+        public string GetLibraryId(string name, string version, string providerId)
+        {
+            if (!string.IsNullOrEmpty(providerId) && _perProviderNamingScheme.TryGetValue(providerId, out ILibraryNamingScheme _scheme))
+            {
+                return _scheme.GetLibraryId(name, version);
+            }
+
+            return _default.GetLibraryId(name, version);
+        }
+    }
+}
--- a/src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
+++ b/src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
@@ -22,27 +22,42 @@ namespace Microsoft.Web.LibraryManager.LibraryNaming
             _versionedNamingScheme = new VersionedLibraryNamingScheme();
             _simpleNamingScheme = new SimpleLibraryNamingScheme();
             _default = _simpleNamingScheme;
+            _isInitialized = false;
+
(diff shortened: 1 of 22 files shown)


src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        private LibraryIdToNameAndVersionConverter()
+        {
+            _versionedNamingScheme = new VersionedLibraryNamingScheme();
+            _simpleNamingScheme = new SimpleLibraryNamingScheme();
+            _default = _simpleNamingScheme;
+        }
+
+        private readonly object _syncObject = new object();
+        private readonly ILibraryNamingScheme _versionedNamingScheme = new VersionedLibraryNamingScheme();
```
alexgav: You are creating values both here and in the constructor.

src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Initializes the LibraryNaming Schemes for the manifest.
+        /// </summary>
+        /// <param name="dependencies"></param>
+        public void Initialize (IDependencies dependencies)
+        {
+            _dependencies = dependencies ?? throw new ArgumentNullException(nameof(dependencies));
+
+            lock(_syncObject)
```
alexgav: Should we also lock on reading? What if one thread is initializing while another is calling one of the GetXXX methods? And/or should we use ConcurrentDictionary?

src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Splits the libraryId into name and version based on the provider.
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="providerId"></param>
+        /// <returns></returns>
+        public (string Name, string Version) GetLibraryNameAndVersion(string libraryId, string providerId)
+        {
+            if (!string.IsNullOrEmpty(providerId) && _perProviderNamingScheme.TryGetValue(providerId, out ILibraryNamingScheme _scheme))
```
alexgav: Another thread could call Initialize in the middle of the TryGetValue. Seems that we should either lock here as well or use ConcurrentDictionary?

src/LibraryManager/Manifest.cs
```diff
@@ -97,6 +98,7 @@ internal static Manifest FromJson(string json, IDependencies dependencies)
         {
             try
             {
+                LibraryIdToNameAndVersionConverter.Instance.Initialize(dependencies);
```
alexgav: This doesn't seem like the right place to init a shared static instance of LibraryIdToNameAndVersionConverter. It would be fine if we had per-manifest instances of LibraryIdToNameAndVersionConverter, but we don't. Since all manifests share it, I don't believe each should re-initialize it. If we want to keep init here, it should be something like "EnsureInitialized" with an assumption that the set of providers is always the same for all manifests, and init doesn't get re-done every time. If the set of providers may be different for different manifests, then we should have per-manifest LibraryIdToNameAndVersionConverter instances.

src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
```diff
@@ -134,7 +134,7 @@ public CdnjsCatalog(CdnjsProvider provider)
         /// <returns></returns>
         public async Task<ILibrary> GetLibraryAsync(string libraryId, CancellationToken cancellationToken)
         {
-            (string name, string version) = LibraryNamingScheme.Instance.GetLibraryNameAndVersion(libraryId);
+            (string name, string version) = LibraryIdToNameAndVersionConverter.Instance.GetLibraryNameAndVersion(libraryId, this._provider.Id);
```
alexgav: Nit: do we use "this." prefix? Normally it's either "_" prefix or "this." prefix for member variables.

src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
```diff
@@ -168,7 +168,7 @@ public CdnjsCatalog(CdnjsProvider provider)
 
         public async Task<string> GetLatestVersion(string libraryId, bool includePreReleases, CancellationToken cancellationToken)
         {
-            (string name, string version) = LibraryNamingScheme.Instance.GetLibraryNameAndVersion(libraryId);
+            (string name, string version) = LibraryIdToNameAndVersionConverter.Instance.GetLibraryNameAndVersion(libraryId,  this._provider.Id);
```
alexgav: Same - "this." prefix?

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
```diff
@@ -39,10 +31,10 @@ public class LibraryNamingScheme
             name = libraryId;
             version = string.Empty;
 
-            if (indexOfAt > 0 && indexOfAt < libraryId.Trim().Length -1)
+            if (indexOfAt > 0 && indexOfAt < libraryId.Trim().Length - 1)
```
jodavis: > libraryId.Trim() [](start = 44, length = 17)

libraryId should be trimmed earlier in the method, so indexOfAt is valid.

jodavis: >Trim [](start = 55, length = 4)

Use TrimEnd() instead

src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private readonly object _syncObject = new object();
+        private readonly ILibraryNamingScheme _versionedNamingScheme = new VersionedLibraryNamingScheme();
+        private readonly ILibraryNamingScheme _simpleNamingScheme = new SimpleLibraryNamingScheme();
+        private readonly ILibraryNamingScheme _default;
+
+
+        private IDependencies _dependencies;
+        private Dictionary<string, ILibraryNamingScheme> _perProviderNamingScheme = new Dictionary<string, ILibraryNamingScheme>();
```
jodavis: >Dictionary [](start = 88, length = 10)

Need a case-insensitive string comparer for this Dictionary.

src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// Splits the libraryId into name and version based on the provider.
+        /// </summary>
+        /// <param name="libraryId"></param>
+        /// <param name="providerId"></param>
+        /// <returns></returns>
+        public (string Name, string Version) GetLibraryNameAndVersion(string libraryId, string providerId)
+        {
+            if (!string.IsNullOrEmpty(providerId) && _perProviderNamingScheme.TryGetValue(providerId, out ILibraryNamingScheme _scheme))
+            {
+                return _scheme.GetLibraryNameAndVersion(libraryId);
```
jodavis: >_scheme [](start = 23, length = 7)

Nit: local variable shouldn't have an underscore.

src/LibraryManager/LibraryNaming/LibraryIdToNameAndVersionConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="name"></param>
+        /// <param name="version"></param>
+        /// <param name="providerId"></param>
+        /// <returns></returns>
+        public string GetLibraryId(string name, string version, string providerId)
+        {
+            if (!string.IsNullOrEmpty(providerId) && _perProviderNamingScheme.TryGetValue(providerId, out ILibraryNamingScheme _scheme))
+            {
+                return _scheme.GetLibraryId(name, version);
+            }
```
jodavis: This could be a shared helper between GetLibraryId and GetNameAndVersion. Needs to have a common lock as well.

jodavis:
>             return assets?.Select(a => $"{groupName}@{a.Version}");

This should also be using the naming scheme converter.

---
Refers to: src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs:274 in 773db1b. [](commit_id = 773db1bd866cb479855cae9aaf20b24c62639749, deletion_comment = False)

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -5,7 +5,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Web.LibraryManager.Contracts;
-using Microsoft.Web.LibraryManager.Helpers;
+using Microsoft.Web.LibraryManager.LibraryNaming;
```
jodavis: Nit: Is this namespace used in this file?

src/LibraryManager/Manifest.cs
prafullbhosale: The providers will remain the same for all providers for now. 
So changing this to be `EnsureInitialized()` for now. 
I am going to open a new issue to change the converter to be per manifest instead (Its going to be a much bigger change).

prafullbhosale:
:up: :date:

Issue #201, opened 2018-06-21
[Add Files Dialog] Target location needs to append library name to current folder


justcla:
When the user tabs out of the Library field, if there is a valid library (ie. files exist in the Files box), reset the Specified location field to a new value that is: [LaunchFolder]/[LibraryName]
(where [LaunchFolder] might be (wwwroot)/lib if launched from the Project root. See #200)

Keep a track of the last suggested location.

Before resetting to a new suggested location, first check if the current value in the location field is the same as the last suggestion. If it is different, do make reset the location field.
If the value is the same as the last suggestion, replace it with the new suggested location (and store that as the new last-suggested-location value).

jodavis:
This is part of a larger design change that isn't finished yet.

justcla:
Not working correctly in all cases.

Need to check logic for when to reset user's manually changed target location.

Logic should be:
            // initialSuggestion(lastSuggestion) = (init - launch folder or wwwroot/lib)

            // when a new library is chosen
            // if (current value in target location == the LastSuggestion) {
                // newSuggestion = (generate new suggestion: [LaunchFolder]/[libraryName]
                // update the LastSuggegstion to = newSuggestion
            // }
        }

YangLyu89:
resolved the issue Justin mentioned.

PR #285, opened 2018-07-06
Make change to the target location for valid libraries

YangLyu89:
This is to resolve https://github.com/aspnet/LibraryManager/issues/201


src/LibraryManager.Vsix/UI/Controls/MutualPropertyChange.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.ComponentModel;
+
+namespace Microsoft.Web.LibraryManager.Vsix.UI.Controls
+{
+    internal class MutualPropertyChange : INotifyPropertyChanged
```
bhsubra: >INotifyPropertyChanged [](start = 42, length = 22)

Use BindableBase instead.

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
```diff
@@ -284,5 +291,39 @@ private void Library_LostFocus(object sender, RoutedEventArgs e)
...
+                CompletionItem item = items.First();
+
+                if (item.InsertionText.Equals(Text))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
bhsubra: Why are we duplicating the code here? You can get the name from SelectedPackage in ViewModel. Let's sync up offline.

src/LibraryManager.Vsix/UI/Controls/MutualPropertyChange.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private string _libraryName;
+        public event PropertyChangedEventHandler PropertyChanged;
```
bhsubra: Can go away.

src/LibraryManager.Vsix/UI/Controls/MutualPropertyChange.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        protected void OnPropertyChanged(string name)
+        {
+            PropertyChangedEventHandler handler = PropertyChanged;
+
+            if (handler != null)
+            {
+                handler(this, new PropertyChangedEventArgs(name));
+            }
+        }
```
bhsubra: Can go away.

src/LibraryManager.Vsix/UI/Controls/MutualPropertyChange.cs
YangLyu89: Will do

---
In reply to: [200684735](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200684735) [](ancestors = 200684735)

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
```diff
@@ -284,5 +291,39 @@ private void Library_LostFocus(object sender, RoutedEventArgs e)
                 Flyout.IsOpen = false;
             }
         }
+
+        private string GetLibraryName(string insertionText)
```
prafullbhosale: If I understand correctly, this value gets appended to the `_baseFolder` in `TargetLocation.xaml.cs` 
If that is the case, you should be calling the provider to get the suggested folder name. 

The current logic doesn't work well for `unpkg` and `filesystem` providers.

For eg: `unpkg` allows library Ids like `@angular/cli@4.0.0` , the current logic would return empty string. 
For `filesystem` it would end up returning the entire path of the library (if it doesn't contain `@`)

YangLyu89: I am start searching the index of @ from the second character so I think it would be Okay for unpkg (index will be -1 if it's not completed). If it was filesystem provider (at index == -1), I will return the insertiontext.

---
In reply to: [200735579](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200735579) [](ancestors = 200735579)

YangLyu89: Now changed to get the value from InstallDialogViewModel.cs, so delete the function.

---
In reply to: [200749858](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200749858) [](ancestors = 200749858,200735579)

src/LibraryManager.Vsix/UI/Controls/Library.xaml.cs
YangLyu89: As discussed, using Installdialogviewmodel instead

---
In reply to: [200697404](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200697404) [](ancestors = 200697404)

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
```diff
@@ -270,6 +275,8 @@ private void TargetLocationSearchTextBox_TextChanged(object sender, TextChangedE
                     Flyout.IsOpen = true;
                 });
             }
+
+            InstallationFolder.DestinationFolder = TargetLocationSearchTextBox.Text;
```
bhsubra: Do you still need this?

src/LibraryManager.Vsix/UI/Controls/MutualPropertyChange.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.ComponentModel;
+using Microsoft.Web.LibraryManager.Vsix.UI.Models;
+
+namespace Microsoft.Web.LibraryManager.Vsix.UI.Controls
+{
+    internal class MutualPropertyChange : BindableBase
```
jodavis: >MutualPropertyChange [](start = 19, length = 20)

This needs a comment explaining what it is for. It's not obvious from the code.

Also, it should have a more descriptive name. Maybe "BindLibraryNameToTargetLocation". It doesn't hav eto be that, but choose a name so when I look at the class I know what it's doing.

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
```diff
@@ -26,14 +26,19 @@ public partial class TargetLocation : INotifyPropertyChanged
...
         public TargetLocation()
         {
             InitializeComponent();
 
             // Pre populate textBox with folder name
             TargetLocationSearchTextBox.Text = InstallationFolder.DestinationFolder;
-
+            _baseFolder = InstallationFolder.DestinationFolder;
+            _lastTargetLocation = InstallationFolder.DestinationFolder;
+            MutualPropertyChange.Instance.TargetLibrary = string.Empty;
```
jodavis: >Instance [](start = 33, length = 8)

Let's not use a singleton here. You can just create an instance of the class for this dialog. It can be part of the InstallDialogViewModel.

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
```diff
@@ -279,5 +286,26 @@ private void TargetLocation_LostFocus(object sender, RoutedEventArgs e)
...
+            string targetLibrary = MutualPropertyChange.Instance.TargetLibrary;
+
+            this.Dispatcher.Invoke(() =>
+            {
+                if (TargetLocationSearchTextBox.Text.Equals(_lastTargetLocation))
+                {
+                    if (targetLibrary.Length > 0 && targetLibrary[targetLibrary.Length - 1] == '/')
+                    {
+                        targetLibrary = targetLibrary.Substring(0, targetLibrary.Length - 1);
+                    }
```
jodavis: Prafull: The provider will have a suggestion for what folder should be appended. We want to match what the CLI would suggest, so look at that logic.

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
```diff
@@ -279,5 +286,26 @@ private void TargetLocation_LostFocus(object sender, RoutedEventArgs e)
...
             }
         }
+
+        private void LibraryNameChanged(object sender, PropertyChangedEventArgs e)
+        {
+            string targetLibrary = MutualPropertyChange.Instance.TargetLibrary;
+
+            this.Dispatcher.Invoke(() =>
+            {
+                if (TargetLocationSearchTextBox.Text.Equals(_lastTargetLocation))
```
jodavis: >Text [](start = 48, length = 4)

Can this ever be null?

jodavis: Bhavya thinks this should always be at least an empty string, not null.

---
In reply to: [200772612](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200772612) [](ancestors = 200772612)

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
jodavis: Yes, this is still needed.

---
In reply to: [200753183](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200753183) [](ancestors = 200753183)

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
YangLyu89: Yes. I am able to use SelectedPackage to get the suggested destination for the selectedProvider

---
In reply to: [200772301](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200772301) [](ancestors = 200772301)

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
YangLyu89: Need to set the event listener after the targetLcation get loaded because the initialization of target location takes place prior to the initialization of InstallDialogViewModel

---
In reply to: [200771611](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200771611) [](ancestors = 200771611)

src/LibraryManager.Vsix/UI/Controls/MutualPropertyChange.cs
YangLyu89: Will do

---
In reply to: [200770461](https://github.com/aspnet/LibraryManager/pull/285#discussion_r200770461) [](ancestors = 200770461)

src/LibraryManager.Vsix/UI/Controls/TargetLocation.xaml.cs
```diff
@@ -279,5 +290,26 @@ private void TargetLocation_LostFocus(object sender, RoutedEventArgs e)
                 Flyout.IsOpen = false;
             }
         }
+
+        private void LibraryNameChanged(object sender, PropertyChangedEventArgs e)
+        {
+            string targetLibrary = _libraryNameChange.LibraryName;
```
prafullbhosale: >targetLibrary [](start = 19, length = 13)

Check this string to be null or empty.

YangLyu89: Will do

---
In reply to: [201086028](https://github.com/aspnet/LibraryManager/pull/285#discussion_r201086028) [](ancestors = 201086028)

Issue #119, opened 2018-06-11
Delete empty folders after a clean operation


raquelsa:
### Functional impact
Currently only the immediate parent of a deleted library is also deleted if empty. We need to delete all folders (if empty) up the chain until working directory.

justcla:
Ensure we update the docs when this fix is rolled out.

"Delete library files" section says:

> To prevent unintentional removal of non-library files, the clean operation doesn't delete whole directories. It only removes files that were included in the previous restore.

justcla:
My bad. That statement is about not deleting directories with other files. Closing issue.

PR #286, opened 2018-07-06
delete empty folders after clean operation.

YangLyu89:
This is to resolve https://github.com/aspnet/LibraryManager/issues/119

### DIFF
--- a/src/LibraryManager.Contracts/FileHelpers.cs
+++ b/src/LibraryManager.Contracts/FileHelpers.cs
@@ -210,8 +210,9 @@ namespace Microsoft.Web.LibraryManager.Contracts
         /// Deletes files 
         /// </summary>
         /// <param name="filePaths"></param>
+        /// <param name="rootDirectory"></param>
         /// <returns></returns>
-        public static bool DeleteFiles(IEnumerable<string> filePaths)
+        public static bool DeleteFiles(IEnumerable<string> filePaths, string rootDirectory = null)
         {
             HashSet<string> directories = new HashSet<string>();
 
@@ -224,14 +225,21 @@ namespace Microsoft.Web.LibraryManager.Contracts
                         DeleteFileFromDisk(filePath);
                     }
 
-                    string directoryPath = Path.GetDirectoryName(filePath);
-                    if (Directory.Exists(directoryPath))
+                    if (rootDirectory != null)
                     {
-                        directories.Add(directoryPath);
+                        string directoryPath = Path.GetDirectoryName(filePath);
+
+                        if (Directory.Exists(directoryPath))
+                        {
+                            directories.Add(directoryPath);
+                        }
                     }
                 }
 
-                DeleteEmptyFoldersFromDisk(directories);
+                if (rootDirectory != null)
+                {
+                    DeleteEmptyFoldersFromDisk(directories, rootDirectory);
+                }
 
                 return true;
             }
@@ -245,21 +253,33 @@ namespace Microsoft.Web.LibraryManager.Contracts
         /// Deletes folders if empty
         /// </summary>
         /// <param name="folderPaths"></param>
+        /// <param name="rootDirectory"></param>
         /// <returns></returns>
-        public static bool DeleteEmptyFoldersFromDisk(IEnumerable<string> folderPaths)
+        public static bool DeleteEmptyFoldersFromDisk(IEnumerable<string> folderPaths, string rootDirectory)
         {
+            if (folderPaths.Count() == 0)
+            {
+                return true;
+            }
+
+            HashSet<string> newFolderPaths = new HashSet<string>();
+
             try
             {
                 foreach (string path in folderPaths)
                 {
-                    DirectoryInfo directory = new DirectoryInfo(path);
-                    if (directory.Exists && IsDirectoryEmpty(path))
+                    if (!path.Equals(rootDirectory, StringComparison.CurrentCultureIgnoreCase))
                     {
-                        directory.Delete(true);
+                        DirectoryInfo directory = new DirectoryInfo(path);
+                        if (directory.Exists && IsDirectoryEmpty(path))
+                        {
+                            directory.Delete(true);
+                            newFolderPaths.Add(Directory.GetParent(path).FullName);
+                        }
                     }
                 }
 
-                return true;
+                return DeleteEmptyFoldersFromDisk(newFolderPaths, rootDirectory);
             }
             catch (Exception)
             {
--- a/src/LibraryManager.Contracts/FileHelpers.cs
+++ b/src/LibraryManager.Contracts/FileHelpers.cs
@@ -263,13 +263,14 @@ namespace Microsoft.Web.LibraryManager.Contracts
                 return true;
             }
 
-            HashSet<string> newFolderPaths = new HashSet<string>(new LocalPathComparer());
+            LocalPathComparer comparer = new LocalPathComparer();
+            HashSet<string> newFolderPaths = new HashSet<string>(comparer);
 
             try
             {
                 foreach (string path in folderPaths)
                 {
-                    if (!string.Equals(Path.GetFullPath(path), Path.GetFullPath(rootDirectory)))
+                    if (!comparer.Equals(path, rootDirectory))
                     {
                         if (Directory.Exists(path) && !Directory.GetFileSystemEntries(path).Any())
                         {
--- a/src/LibraryManager.Contracts/FileHelpers.cs
+++ b/src/LibraryManager.Contracts/FileHelpers.cs
@@ -214,33 +214,42 @@ namespace Microsoft.Web.LibraryManager.Contracts
         /// <returns></returns>
         public static bool DeleteFiles(IEnumerable<string> filePaths, string rootDirectory = null)
         {
-            HashSet<string> directories = new HashSet<string>(new LocalPathComparer());
+            HashSet<string> directories = new HashSet<string>();
 
             try
             {
                 foreach (string filePath in filePaths)
                 {
-                    if (File.Exists(filePath))
-                    {
-                        DeleteFileFromDisk(filePath);
-                    }
+                    Debug.Assert(Path.IsPathRooted(filePath));
 
-                    if (!string.IsNullOrEmpty(rootDirectory))
+                    if (Path.IsPathRooted(filePath))
                     {
-                        string directoryPath = Path.GetDirectoryName(filePath);
-
-                        if (Directory.Exists(directoryPath))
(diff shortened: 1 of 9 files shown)


src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -210,8 +210,9 @@ public static bool AreFilesUpToDate(FileInfo firstFile, FileInfo secondFile)
         /// Deletes files 
         /// </summary>
         /// <param name="filePaths"></param>
+        /// <param name="rootDirectory"></param>
         /// <returns></returns>
-        public static bool DeleteFiles(IEnumerable<string> filePaths)
+        public static bool DeleteFiles(IEnumerable<string> filePaths, string rootDirectory = null)
         {
             HashSet<string> directories = new HashSet<string>();
```
prafullbhosale: You need to pass in a comparer for this hashset

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -245,21 +253,33 @@ public static bool DeleteFiles(IEnumerable<string> filePaths)
...
         /// <returns></returns>
-        public static bool DeleteEmptyFoldersFromDisk(IEnumerable<string> folderPaths)
+        public static bool DeleteEmptyFoldersFromDisk(IEnumerable<string> folderPaths, string rootDirectory)
         {
+            if (folderPaths.Count() == 0)
+            {
+                return true;
+            }
+
+            HashSet<string> newFolderPaths = new HashSet<string>();
```
prafullbhosale: Comparer for the hashset.

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -245,21 +253,33 @@ public static bool DeleteFiles(IEnumerable<string> filePaths)
...
+
             try
             {
                 foreach (string path in folderPaths)
                 {
-                    DirectoryInfo directory = new DirectoryInfo(path);
-                    if (directory.Exists && IsDirectoryEmpty(path))
+                    if (!path.Equals(rootDirectory, StringComparison.CurrentCultureIgnoreCase))
```
prafullbhosale: This block also needs to make sure that the directories being deleted are inside the `rootDirectory`

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -224,14 +225,21 @@ public static bool DeleteFiles(IEnumerable<string> filePaths)
                         DeleteFileFromDisk(filePath);
                     }
 
-                    string directoryPath = Path.GetDirectoryName(filePath);
-                    if (Directory.Exists(directoryPath))
+                    if (rootDirectory != null)
```
jodavis: >rootDirectory != null [](start = 24, length = 21)

Use String.IsNullOrEmpty

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -245,21 +253,33 @@ public static bool DeleteFiles(IEnumerable<string> filePaths)
...
             try
             {
                 foreach (string path in folderPaths)
                 {
-                    DirectoryInfo directory = new DirectoryInfo(path);
-                    if (directory.Exists && IsDirectoryEmpty(path))
+                    if (!path.Equals(rootDirectory, StringComparison.CurrentCultureIgnoreCase))
                     {
-                        directory.Delete(true);
+                        DirectoryInfo directory = new DirectoryInfo(path);
```
jimmylewis: I don't see a need to create a DirectoryInfo object here.  We've been using the File/Directory static methods elsewhere.  This segment could be rewritten as:

```
if (Directory.Exists(path) && Directory.GetFileSystemEntries(path).Any())
{
    Directory.Delete(path);
    newFolderPaths.Add(Path.GetDirectoryName(path));
}
```
(This also gets rid of the need for IsDirectoryEmpty from this method, I'm not sure if it's needed elsewhere).

src/LibraryManager.Contracts/FileHelpers.cs
YangLyu89: Adding the comparer for the local path, need to further support for network path later.

---
In reply to: [200727108](https://github.com/aspnet/LibraryManager/pull/286#discussion_r200727108) [](ancestors = 200727108)

src/LibraryManager.Contracts/FileHelpers.cs
YangLyu89: Use the full path here to compare.

---
In reply to: [200728323](https://github.com/aspnet/LibraryManager/pull/286#discussion_r200728323) [](ancestors = 200728323)

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -304,5 +324,28 @@ public static bool DeleteFileFromDisk(string filePath)
...
         }
+
+        private static bool IsUNCPath(string rootPath)
+        {
+            return new Uri(rootPath).IsUnc;
+        }
+
+        private class LocalPathComparer : IEqualityComparer<string>
+        {
+            public bool Equals(string path1, string path2)
```
prafullbhosale: >Equals [](start = 24, length = 6)

This does not account for paths that are different only in case. 

`D:\GIT\LIBRARYMANAGER` and `d:\git\librarymanager` need to return equal on windows. 

Also this does not handle trailing `\` or `/`

`D:\git\LibraryManager\` and `D:\git\LibraryManager` should return equal.

Take a look at [RelativePathEqualityComparer](https://github.com/aspnet/LibraryManager/blob/master/src/LibraryManager/RelativePathEqualityComparer.cs)

YangLyu89: make related change here as this function does.

---
In reply to: [201064752](https://github.com/aspnet/LibraryManager/pull/286#discussion_r201064752) [](ancestors = 201064752)

src/LibraryManager.Contracts/FileHelpers.cs
YangLyu89: added

---
In reply to: [200727652](https://github.com/aspnet/LibraryManager/pull/286#discussion_r200727652) [](ancestors = 200727652)

src/LibraryManager.Contracts/FileHelpers.cs
prafullbhosale: Can you file a separate issue on github to unify these two classes? 

---
In reply to: [201109962](https://github.com/aspnet/LibraryManager/pull/286#discussion_r201109962) [](ancestors = 201109962,201064752)

YangLyu89: Sure. Will do.

---
In reply to: [201131620](https://github.com/aspnet/LibraryManager/pull/286#discussion_r201131620) [](ancestors = 201131620,201109962,201064752)

src/LibraryManager.Contracts/FileHelpers.cs
prafullbhosale: This should be done using `LocalPathComparer`, to handle the cases we discussed for the comparer.



---
In reply to: [200867103](https://github.com/aspnet/LibraryManager/pull/286#discussion_r200867103) [](ancestors = 200867103,200728323)

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -245,21 +254,33 @@ public static bool DeleteFiles(IEnumerable<string> filePaths)
         /// Deletes folders if empty
         /// </summary>
         /// <param name="folderPaths"></param>
+        /// <param name="rootDirectory"></param>
         /// <returns></returns>
-        public static bool DeleteEmptyFoldersFromDisk(IEnumerable<string> folderPaths)
+        public static bool DeleteEmptyFoldersFromDisk(IEnumerable<string> folderPaths, string rootDirectory)
```
jimmylewis: >public [](start = 8, length = 6)

This should be private since it's only used in this class

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -245,21 +254,33 @@ public static bool DeleteFiles(IEnumerable<string> filePaths)
...
         {
+            if (folderPaths.Count() == 0)
+            {
+                return true;
+            }
+
+            LocalPathComparer comparer = new LocalPathComparer();
```
jimmylewis: >comparer [](start = 30, length = 8)

If you make the change above to ensure all of these are already under rootDirectory, then this comparer isn't needed.

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -210,10 +210,11 @@ public static bool AreFilesUpToDate(FileInfo firstFile, FileInfo secondFile)
         /// Deletes files 
         /// </summary>
         /// <param name="filePaths"></param>
+        /// <param name="rootDirectory"></param>
         /// <returns></returns>
-        public static bool DeleteFiles(IEnumerable<string> filePaths)
+        public static bool DeleteFiles(IEnumerable<string> filePaths, string rootDirectory = null)
```
jimmylewis: >filePaths [](start = 59, length = 9)

We should assert that all of these are absolute paths

src/LibraryManager.Contracts/FileHelpers.cs
YangLyu89: Yes it is

---
In reply to: [201165106](https://github.com/aspnet/LibraryManager/pull/286#discussion_r201165106) [](ancestors = 201165106)

src/LibraryManager.Contracts/FileHelpers.cs
YangLyu89: Will do

---
In reply to: [201168854](https://github.com/aspnet/LibraryManager/pull/286#discussion_r201168854) [](ancestors = 201168854)

src/LibraryManager.Contracts/FileHelpers.cs
YangLyu89: It's deleted.

---
In reply to: [201166856](https://github.com/aspnet/LibraryManager/pull/286#discussion_r201166856) [](ancestors = 201166856)

YangLyu89:
>                     if (File.Exists(filePath))

Will do.

---
In reply to: [403639955](https://github.com/aspnet/LibraryManager/pull/286#issuecomment-403639955) [](ancestors = 403639955)

---
Refers to: src/LibraryManager.Contracts/FileHelpers.cs:223 in 26d7942. [](commit_id = 26d7942b45deed7268ad57ccb97e5893e9b347dc, deletion_comment = False)

Issue #163, opened 2018-06-13
Autocomplete of 'jquerycarousel' becomes 'jquery_carousel' 


bhsubra:
### Functional impact
Autocomplete of 'jquerycarousel' becomes 'jquery_carousel'  for unPkg provider

### Minimal repro steps
 -  Select Unpkg provider, type jquery and pause. The 2nd option in the drop down is 'jquerycarousel' select that with mouse
-   The text in the library text box is 'jquery_carousel' instead of 'jquerycarousel'

### Expected result
Expected the text to be jquerycarousel

### Actual result
Actual text is jquery_carousel

I'm not sure if this is desired behavior or by design but we need to verify

jodavis:
This could be related to https://github.com/aspnet/LibraryManager/issues/198

alexgav:
Fixed with https://github.com/aspnet/LibraryManager/pull/290

PR #290, opened 2018-07-07
Fix missing underscores in library names in the UI

alexgav:
Fixing https://github.com/aspnet/LibraryManager/issues/163 - **Autocomplete of 'jquerycarousel' becomes 'jquery_carousel'**

The issue here was that ```<AccessText>``` was used in our UI instead of ```<TextBlock>``` for some reason. AccessText uses ```_``` character as a special meta-character to specify that the following character is the "access key", so the ```_``` is not normally visible (until you press Alt button). Pressing the access key will invoke OnAccessKey method of the control using AccessText. It doesn't look that functionality was being used at all. I looked through all of our xaml and replaced all uses of AccessText with TextBlock. None of the uses of AccessText were taking advantage of the access key functionality or handling OnAccessKey method.

For more info, see

https://stackoverflow.com/questions/4601801/wpf-listbox-skip-underscore-symbols-in-strings

and

https://stackoverflow.com/questions/35534223/wpf-uses-and-expected-results-of-accesstext-class


alexgav:
Thanks Bhavya!

Issue #197, opened 2018-06-21
[Menu items] "Add->Client-Side Libraries" needs to be added to project root.


justcla:
User should be able to right-click the project root and choose Add-.Client-Side LIbraries

bhsubra:
Fixed via https://github.com/aspnet/LibraryManager/commit/174e8351b18fa6b2b3988a1da1f84cc52721b927

PR #291, opened 2018-07-09
"Add->Client-Side Libraries" needs to be added to project root

bhsubra:
Fix for issue https://github.com/aspnet/LibraryManager/issues/197.

Fixes below issues:
- User should be able to right-click the project root and choose Add-.Client-Side library
- Set initial target location. If wwwroot exists, target location should be-  wwwroot/lib. If not, just- lib

Sending the pr against @YangLyu89 branch as we need to make sure we don't cause any regressions when merging both changes.


src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -60,52 +59,67 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
             button.Visible = button.Enabled = false;
 
             ProjectItem item = await VsHelpers.GetSelectedItemAsync();
```
ToddGrun: >; [](start = 69, length = 1)

ConfigureAwait here too?

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -60,52 +59,67 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
...
+            Project project = await VsHelpers.GetProjectOfSelectedItemAsync().ConfigureAwait(false);
```
ToddGrun: >await VsHelpers.GetProjectOfSelectedItemAsync(). [](start = 30, length = 48)

Is there a way to get this from the ProjectItem rather than calling an async VSHelper method?

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -60,52 +59,67 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
...
 
-            if (item?.ContainingProject == null)
+            if (item?.ContainingProject == null && project == null)
```
ToddGrun: >&& [](start = 48, length = 2)

Just making sure this isn't supposed to be an ||

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
bhsubra: ProjectItem is going to be null when command is invoked from project scope. I want to let is pass for both the scenarios: when command is invoked from folder and project

---
In reply to: [201045767](https://github.com/aspnet/LibraryManager/pull/291#discussion_r201045767) [](ancestors = 201045767)

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -60,52 +59,67 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
...
-                Project project = await VsHelpers.GetProjectOfSelectedItemAsync().ConfigureAwait(false);
-
-                if (project != null)
+                // Install command was invoked from a folder.
+                // So the initial target location should be name of the folder from which
+                // the command was invoked.
+                if (item != null)
                 {
-                    string rootFolder = await project.GetRootFolderAsync().ConfigureAwait(false);
+                    target = item.FileNames[1];
```
ToddGrun: >item.FileNames[1] [](start = 29, length = 17)

This includes the separator?

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
ToddGrun: >ProjectItem item [](start = 12, length = 16)

Do you need to verify this is a folder and not a file?

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
bhsubra: Yes, this already includes the separator

---
In reply to: [201047703](https://github.com/aspnet/LibraryManager/pull/291#discussion_r201047703) [](ancestors = 201047703)

Issue #307, opened 2018-07-13
[Add Files UI] [Web Site Project]: Target Location suggestion has an invalid forward slash at the front


justcla:
When using LibMan Add Files UI, the UI adds a suggested folder in the "Target Location" field.
In the case of a Web Site project, the suggested folder contains a forward slash at the start of the path.

This causes the LibMan install operation to fail, as it finds the path to be invalid.

Suggested fix: Amend the suggestion to not have a forward slash at the beginning.

PR #316, opened 2018-07-16
check for start of the path, trim if start with "/" or "\" when setting the destination folder.

YangLyu89:
This is to resolve https://github.com/aspnet/LibraryManager/issues/307

### DIFF
--- a/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
```diff
@@ -24,7 +24,7 @@ internal partial class InstallDialog : DialogWindow
 
         public InstallDialog(IDependencies dependencies, ILibraryCommandService libraryCommandService, string configFileName, string fullPath, string rootFolder)
         {
-            InstallationFolder.DestinationFolder = fullPath.Replace(rootFolder, "").Replace('\\', '/');
```
jodavis: >Replace [](start = 60, length = 7)

Use StringComparison.OrdinalIgnoreCase here

src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
```diff
@@ -24,7 +24,7 @@ internal partial class InstallDialog : DialogWindow
...
+            InstallationFolder.DestinationFolder = fullPath.Replace(rootFolder, "").TrimStart(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar).Replace('\\', '/');
```
jodavis: >fullPath.Replace(rootFolder, "").TrimStart(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar).Replace('\\', '/') [](start = 51, length = 122)

nit: Break this into multiple lines to make it easier to read.

src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
YangLyu89: use starswith for it

---
In reply to: [202842715](https://github.com/aspnet/LibraryManager/pull/316#discussion_r202842715) [](ancestors = 202842715)

src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
YangLyu89: Will do

---
In reply to: [202842892](https://github.com/aspnet/LibraryManager/pull/316#discussion_r202842892) [](ancestors = 202842892)

Issue #309, opened 2018-07-13
Project file not always updated when LibMan operations occur


justcla:
[P1] [ProjRef] Asp.NET Web Project (Framework) [Empty]: libman.json not added to the project when using Add Files UI
[P2] [ProjRef] ASP.NET (Core) [Empty]: Clean will delete files and folders, but does not remove them from the project.
[P1] [ProjRef] Asp.NET (Fraemwork): Restore operation does not add new folders to project

justcla:
Issue is with Framework projects.
Repro:
Create ASP.NET (Framework) project. (Empty)
Right-click the project root and Add Client Side Library.
Add latest jquery library.
Files get added to the file system, but they are not included in the Project.

PR #317, opened 2018-07-16
Make adding client-library work for .NET project

YangLyu89:
This is to resolve first part of https://github.com/aspnet/LibraryManager/issues/309

### DIFF
--- a/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
+using EnvDTE;
 using Microsoft.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
```diff
@@ -145,9 +145,11 @@ private void CloseDialog(bool res)
             Close();
         }
 
-        private void OnLoaded(object sender, RoutedEventArgs e)
+        private async void OnLoadedAsync(object sender, RoutedEventArgs e)
         {
-            ViewModel = new InstallDialogViewModel(Dispatcher, _libraryCommandService, _configFileName, _deps, _fullPath, CloseDialog);
+            EnvDTE.Project project = await VsHelpers.GetProjectOfSelectedItemAsync();
```
jodavis: >GetProjectOfSelectedItemAsync [](start = 53, length = 29)

Instead of doing this here, pass it into the constructor from InstallLibraryCommand.

YangLyu89: Will do

---
In reply to: [202841135](https://github.com/aspnet/LibraryManager/pull/317#discussion_r202841135) [](ancestors = 202841135)

PR #318, opened 2018-07-16
Remove usage of LibraryIdToNameAndVersionConverter

prafullbhosale:
- Adds API to the provider to convert libraryId to name and version (and also the other way).
- This allows providers to use their own naming schemes

Consequences:
- The `Name` and `Version` properties in `ILibraryInstallationState` can now only be used after the provider has done a `UpdateState` on the state. 
- The provider has to ensure to update the `Name` and `Version` properties of the `LibraryInstatllationState` during `UpdateStateAsync`


prafullbhosale:
Abandoning this in favor of #326

PR #323, opened 2018-07-17
Dev/bhsubra/install library command add check for file open

bhsubra:
Get manifest contents from buffer if file is open

### DIFF
--- a/src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
+++ b/src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
@@ -162,7 +162,6 @@ namespace Microsoft.Web.LibraryManager.Vsix
         private async Task<Manifest> GetManifestAsync(string configFilePath, IDependencies dependencies)
         {
             RunningDocumentTable rdt = new RunningDocumentTable(ServiceProvider.GlobalProvider);
-
             IVsTextBuffer textBuffer = rdt.FindDocument(configFilePath) as IVsTextBuffer;
 
             Manifest manifest = null;
--- a/src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
+++ b/src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
@@ -2,10 +2,15 @@
 using System.ComponentModel.Design;
 using System.IO;
 using System.Threading;
+using System.Threading.Tasks;
 using System.Windows.Interop;
 using EnvDTE;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Editor;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.TextManager.Interop;
 using Microsoft.Web.LibraryManager.Contracts;
 using Task = System.Threading.Tasks.Task;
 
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -88,7 +93,7 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
                 string configFilePath = Path.Combine(rootFolder, Constants.ConfigFileName);
                 IDependencies dependencies = Dependencies.FromConfigFile(configFilePath);
 
-                Manifest manifest = await Manifest.FromFileAsync(configFilePath, dependencies, CancellationToken.None).ConfigureAwait(false);
```
ToddGrun: >.ConfigureAwait(false); [](start = 118, length = 23)

Intentionally deleted?

bhsubra: I moved it inside GetManifestAsync(..) so I can check if file is open

---
In reply to: [203157458](https://github.com/aspnet/LibraryManager/pull/323#discussion_r203157458) [](ancestors = 203157458)

ToddGrun: but it seems like it should be here too, no?

---
In reply to: [203157935](https://github.com/aspnet/LibraryManager/pull/323#discussion_r203157935) [](ancestors = 203157935,203157458)

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -153,5 +158,34 @@ public static void Initialize(Package package, OleMenuCommandService commandServ
...
+                manifest = await Manifest.FromFileAsync(configFilePath, dependencies, CancellationToken.None).ConfigureAwait(false);
+            }
+            else
+            {
+                IComponentModel componentModel = ServiceProvider.GlobalProvider.GetService(typeof(SComponentModel)) as IComponentModel;
+                IVsEditorAdaptersFactoryService editorAdapterService = componentModel.GetService<IVsEditorAdaptersFactoryService>();
+
+                ITextBuffer documentBuffer = editorAdapterService.GetDocumentBuffer(textBuffer);
+
+                if (documentBuffer != null)
```
ToddGrun: >if (documentBuffer != null) [](start = 16, length = 27)

Is this expected to be possible? If yes, then maybe we should use the Manifest.FromFileAsync code above

bhsubra: I don't expect to hit this. I was only wondering if someone closed the document when operation was in progress.

I'll include the above code or refactor a bit

---
In reply to: [203158515](https://github.com/aspnet/LibraryManager/pull/323#discussion_r203158515) [](ancestors = 203158515)

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
bhsubra: Sorry didn't realize that you highlighted ConfigureAwait(..)
Updated in iteration 3.

---
In reply to: [203158277](https://github.com/aspnet/LibraryManager/pull/323#discussion_r203158277) [](ancestors = 203158277,203157935,203157458)

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
ToddGrun: Change per what we did together.

---
In reply to: [203158905](https://github.com/aspnet/LibraryManager/pull/323#discussion_r203158905) [](ancestors = 203158905,203158515)

bhsubra: @ToddGrun , I did not make the last change that we discussed, It would break this scenario:

1. libman.json file  is open and I remove a comma and cause syntax error
2. Manifest would be null in this case as can't parse correctly
3. The change that we made was to get the manifest using libman.json from disk which doesn't contain changes I made in first step. Manifest that I get using file from disk will parse file and show Add Client-Side libraries dialog instead of warning message box that something went wrong

I'll merge the changes as is and we can discuss tomorrow if you have any suggestion.

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
bhsubra: Updated with the changes we made offline.

PR #326, opened 2018-07-23
Change Provider APIs to use name and version instead of libraryId

prafullbhosale:
- Separates the data objects from Manifest and LibraryInstallationState
- Remove libraryId from from ILibraryInstallationState
- Adapt all clients and use cases to use Name and version instead of
libraryId

PR #338, opened 2018-08-02
Add jsDelivr provider

Drgy:
Adds jsDelivr as a new provider, which supports pulling files from npm and GitHub.

Features for npm packages:
 - name and version autocomplete (using existing NpmPackageSearch implementation)
 - file autocomplete (using [jsDelivr API](https://github.com/jsdelivr/data.jsdelivr.com))
 - support for [jsDelivr's on-demand minification feature](https://www.jsdelivr.com/features#minify) (a minified version is added to the list for any non-minified css/js file)

```json
{
  "version": "1.0",
  "defaultProvider": "jsdelivr",
  "libraries": [
    {
      "destination": "wwwroot/",
      "library": "jquery@3.3.1",
      "files": [
        "dist/jquery.min.js"
      ]
    }
  ]
}
```

Features for GitHub repos:
 - version autocomplete for releases (using [jsDelivr API](https://github.com/jsdelivr/data.jsdelivr.com))
 - any commit hash may also be used as a version
 - file autocomplete for both releases and commits (using [jsDelivr API](https://github.com/jsdelivr/data.jsdelivr.com))
 - support for [jsDelivr's on-demand minification feature](https://www.jsdelivr.com/features#minify) (a minified version is added to the list for any non-minified css/js file)

```json
{
  "version": "1.0",
  "defaultProvider": "jsdelivr",
  "libraries": [
    {
      "destination": "wwwroot/",
      "library": "jquery/jquery@3.3.1",
      "files": [
        "src/ajax.min.js"
      ]
    },
    {
      "destination": "wwwroot/",
      "library": "ractivejs/ractive@8d33e5c787f683f91bb260df1d1ba175e2bad52c",
      "files": [
        "src/Ractive.js"
      ]
    }
  ]
}
```

There were also a few things I'd have implemented differently but decided to do it the same way unpkg provider does for now. I'll be happy to change those in this or future PR if you think it's a good idea:
 - version autocomplete should probably use prefix-only matching
 - npm should support scoped packages
 - jsDelivr [file download stats](https://data.jsdelivr.com/v1/package/npm/jquery@3.2.1/stats) could be used to sort files - this would make autocomplete much more useful for packages with many files


Drgy:
It appears that the appveyor failure is not related to this PR.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -0,0 +1,286 @@
...
+
+            foreach (string version in versionsArray)
+            {
+                versions.Add(name + "@" + version);
+            }
+
+            return versions;
+        }
+
+        public static async Task<bool> IsGitHub(string libraryId)
```
prafullbhosale: >async [](start = 22, length = 5)

This doesn't need to be an async method.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -0,0 +1,286 @@
...
+        {
+            if (libraryId == null)
+            {
+                return false;
+            }
+
+            return libraryId.Contains('/');
```
prafullbhosale: This check would probably be incorrect for scoped npm packages like `@angular/cli`

https://cdn.jsdelivr.net/npm/@angular/cli@1.0.0/

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -0,0 +1,286 @@
...
+                }
+
+                // Make sure we don't list some minified files twice.
+                files = files.Distinct().ToList();
+
+                foreach(string file in files)
+                {
+                    result.Add(file);
+                }
+
```
prafullbhosale: nit: extra newline.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -0,0 +1,286 @@
...
+                foreach (var file in fileArray)
+                {
+                    JValue pathValue = file["name"] as JValue;
+                    string path = pathValue?.Value as string;
+                    
+                    if (path != null && path.Length > 0)
+                    {
+                        // Don't include the leading "/" in the file paths, so you get dist/jquery.js rather than /dist/jquery.js
+                        // We will want the user to always specify a relative path in the "Files" array of the library entry
+                        files.Add(path.Substring(1));
```
prafullbhosale: Instead of `path.Substring(1)` a better check would be to `path.TrimStart('/')`

prafullbhosale: This trimming can also be deferred to the `foreach` block on line 154.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -0,0 +1,286 @@
...
+            }
+            catch (Exception ex)
+            {
+                _provider.HostInteraction.Logger.Log(ex.ToString(), LogLevel.Error);
+            }
+
+            return libraryGroups;
+        }
+
+        public async Task<IEnumerable<string>> GetGithubLibraryVersions(string name)
```
prafullbhosale: Does this method need to be public? 
Seems to be used only in `JsDelivrCatalog` class.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -0,0 +1,286 @@
...
+        {
+            List<string> versions = new List<string>();
+            JObject versionsObject = await WebRequestHandler.Instance.GetJsonObjectViaGetAsync(string.Format(LatestLibraryVersionUrlGH, name), CancellationToken.None).ConfigureAwait(false);
+            JArray versionsArray = versionsObject["versions"] as JArray;
+
+            foreach (string version in versionsArray)
+            {
+                versions.Add(name + "@" + version);
```
prafullbhosale: Use `LibraryIdToNameAndVersionConverter` instead of concatenating.

src/LibraryManager/Providers/jsDelivr/JsDelivrLibraryGroup.cs
```diff
@@ -0,0 +1,48 @@
...
+            Description = description;
+        }
+        public string DisplayName { get; }
+
+        public string Description { get; }
+
+        public async Task<IEnumerable<string>> GetLibraryIdsAsync(CancellationToken cancellationToken)
+        {
+
+            if (JsDelivrCatalog.IsGitHub(DisplayName).Result)
```
prafullbhosale: Is this left empty intentionally (a TODO?) or is it meant to return empty enumerable for github?

src/LibraryManager/Providers/jsDelivr/JsDelivrLibraryGroup.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+
+            }
+            else
+            {
+                Microsoft.Web.LibraryManager.Providers.Unpkg.NpmPackageInfo npmPackageInfo = await Microsoft.Web.LibraryManager.Providers.Unpkg.NpmPackageInfoCache.GetPackageInfoAsync(DisplayName, CancellationToken.None);
```
prafullbhosale: Simplify?

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
MartinKolarik: It is but scoped packages are not supported by the current Npm/Unpkg/LibraryNaming implementations anyway so we thought support can be added later here as well.

src/LibraryManager/Providers/jsDelivr/JsDelivrLibraryGroup.cs
MartinKolarik: I think this is supposed to be empty because there's no library name autocomplete for GitHub (only versions and files). @Drgy?

Drgy:
I updated everything based on your feedback.

MartinKolarik:
@Drgy check the tests please.

justcla:
Hi Jakub

Thanks for the contribution.

Our team are considering whether to take the PR. But before accepting a major contribution such as this, we need to properly evaluate the value.
As such, I'm going to open an issue for discussion.

jimaek:
@justcla doesn't look like anyone wants to join the discussion there, maybe they all agree and don't have any questions?

MartinKolarik:
@jimmylewis I merged your changes into this PR with one small fix.

In addition to the issue with deserialization I noticed that if you hit Enter to autocomplete package version, it completely removes the package name, i.e., `react@16` gets completed to `16.6.0`. It worked correctly before and not sure what broke it yet.

Let me know if you can fix those issues or @Drgy should take a look.

jimmylewis:
The completion issue is because the InsertionText on the completion items needs to contain the full library ID (name@version) because it replaces the entire property value (JSON editor implementation).  We do this in the other providers, e.g:

```C#
var completion = new CompletionItem
{
    DisplayText = group.DisplayName,
    InsertionText = LibraryIdToNameAndVersionConverter.Instance.GetLibraryId(group.DisplayName, group.Version, _provider.Id),
    Description = group.Description,
};
```

It looks like the logic for getting the versions from GitHub was actually doing this in `GetGithubLibraryVersionsAsync`(returning the whole ID as the versions), whereas getting it from the library group was just returning versions.

Still investigating the serialization issue...

jimmylewis:
I think I've tracked down the serialization issue, it looks like it was a regression introduced in #326.  I'm going to get that fixed up, then get back to validating the JSDeliver experience.

jimmylewis:
I've fixed the serialization issue with #390.  Can you rebase these changes and fix the completion issue I mentioned above?

MartinKolarik:
I think the serialization is still broken? It doesn't include the `provider` property.

MartinKolarik:
You are right but it doesn't work correctly when there's no default provider set. Either way, I fixed the completion issue.

jimmylewis:
I see what you mean about the serialization.  I'll take another look shortly.

jimmylewis:
I'm good with this change.  @raquelsa or @prafullbhosale, are there any changes you'd like to see, or can we get this PR in?

nickalbrecht:
I tried to use jsDelivr today after I ran into an issue with the unpkg service being down, but quickly realized that although this has been merged, it is not included in the released version of Library Manager. When can we expect to see this make it into a release?

jimmylewis:
It's available in CLI and MSBuild packages as of this morning (1.0.172).  Unfortunately, it won't be supported in VS until 2019 Preview 2.

You can build a private version of the VSIX to use it in VS2017, just make sure you build from commit `2a33cb3eeb98acdf85` (changes after that are for VS2019 compatibility).

ole1986:
I ave no idea how this works for jsdelivr using github and i also could not find any example

MartinKolarik:
@ole1986 please check Drgy's initial PR comment, there are code samples for both npm and GitHub.

ole1986:
Oh, I have mistakenly wrote the "<user>/<projectname>.git@<hash>" which does not seem to work.
Removing the ".git" suffix resolves it.

Thank you

PR #339, opened 2018-08-03
Loc changes for CLI

prafullbhosale:


src/LibraryManager.Vsix/Commands/CommandTable/LocalizationFiles/VSCommandTable.cs.xlf
```diff
@@ -1,33 +1,33 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd" version="1.2">
-  <file original="E:\LibraryManager\src\LibraryManager.Vsix\Commands\CommandTable\VSCommandTable.en.vsct" source-language="en" datatype="xml" target-language="cs">
+  <file original="C:\Users\bhsubra\Desktop\LibraryManager\src\LibraryManager.Vsix\Commands\CommandTable\VSCommandTable.en.vsct" source-language="en" target-language="cs" datatype="xml">
```
bhsubra: Can you revert these?

prafullbhosale: This fullpath change seems to keep coming back. We should fix it in the loc repo as well. 
And also the positioning of `datatype="xml"` attribute. The attribute maybe a mismatch in settings/ version of the toolkit we use vs the loc team uses.

PR #340, opened 2018-08-03
Cli rename

prafullbhosale:
Renames the CLI nuget package to `Microsoft.Web.LibraryManager.Cli`
Also fixes up NuGet properties for `Copyright` for all packages and `GitCommitId` for the CLI package.


prafullbhosale:
Installing the locally built package and testing out the libman command:

image

PR #341, opened 2018-08-06
Add logic to invoke clean client side libraries command in test clean…

bhsubra:
Added logic to invoke clean client side libraries command during test cleanup phase


test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
```diff
@@ -38,13 +41,51 @@ protected override void DoHostTestInitialize()
...
+            }
+
             base.DoHostTestCleanup();
         }
 
+        private void CleanClientSideLibraries()
+        {
+            Guid guid = Guid.Parse("44ee7bda-abda-486e-a5fe-4dd3f4cefac1");
+            uint commandId = 0x0200;
+            SolutionExplorerItemTestExtension libmanConfigNode = SolutionExplorer.FindItemRecursive(_libman);
```
jimmylewis: >SolutionExplorerItemTestExtension libmanConfigNode = SolutionExplorer.FindItemRecursive(_libman); [](start = 12, length = 97)

This could be fragile if we ever add other projects to the test solution.  Maybe you should pass in `webProject[_libman]` from the caller instead of doing another search here?

It's not perfect, but makes this function more reusable in the future.

jimmylewis: Oh, I guess `webProject[_libman]` is a ProjectItemTestExtension, not a SolutionExplorerTestExtension.  Sigh...

---
In reply to: [208029266](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208029266) [](ancestors = 208029266)

bhsubra: Yes, I thought I could do that. But I don't see an option to Select(..) in ProjectItemTestExtension 

---
In reply to: [208029571](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208029571) [](ancestors = 208029571,208029266)

test/LibraryManager.IntegrationTest/AddClientSideLibrariesFromUITests.cs
```diff
@@ -12,20 +9,17 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
     [TestClass]
     public class AddClientSideLibrariesFromUITests : VisualStudioLibmanHostTest
     {
-        private string _initialLibmanFileContent;
         private string _pathToLibmanFile;
```
jimmylewis: It would be swell if these private fields were inherited from the parent class.  They don't seem specific to this class at all.

test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
```diff
@@ -17,9 +18,11 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
     public class VisualStudioLibmanHostTest : VisualStudioHostTest
     {
         // Solution consts
-        const string _testSolutionName = @"TestSolution.sln";
+        const string _libman = "libman.json";
```
jimmylewis: put access modifiers on all of these.  Probably `protected`?

jimmylewis: That's on the new ones... I think the existing privates should probably remain so.

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
```diff
@@ -52,25 +52,15 @@ public void FileSaveRestore_AddDeleteLibrary()
...
-  ""libraries"": []
-}";
 
             ReplaceFileContent(addingLibraryContent);
             Helpers.FileIO.WaitForRestoredFiles(pathToLibrary, expectedFiles, caseInsensitive: true);
-
-            ReplaceFileContent(deletingLibraryContent);
-            Helpers.FileIO.WaitForDeletedFiles(pathToLibrary, expectedFiles, caseInsensitive: true);
-
-            ReplaceFileContent(_libmanFileContent);
```
YangLyu89: As discussed offline. This is to validate the uninstall libraries can work when changing the libman.json file. Can you add it back?

test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
```diff
@@ -38,13 +41,51 @@ protected override void DoHostTestInitialize()
 
             Solution.Open(_solutionPath);
             Solution.WaitForFullyLoaded(); // This will get modified after bug 627108 get fixed
+
+            string pathToLibmanFile = Path.Combine(SolutionRootPath, _projectName, _libman);
+            _initialLibmanFileContent = File.ReadAllText(pathToLibmanFile);
         }
 
         protected override void DoHostTestCleanup()
```
YangLyu89: >DoHostTestCleanup() [](start = 32, length = 19)

Do we want to have this clean up after each test running or just have TestCleanup() in individual test class so that it will not getting executed for no library changed tests like LibmanCompletionTests.

jimmylewis: Cleaning up after every test ensures that no test can leave it in a dirty state.  Any test which opens/types in this file has a chance of leaving it dirty if something fails or throws during execution.

bhsubra: I don't want to include these in every test. We could override in LibmanCompletionTests. But clean won't do much here since the initial contents has no libraries. So its not expensive. Let me know if you disagree



---
In reply to: [208043077](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208043077) [](ancestors = 208043077)

test/LibraryManager.IntegrationTest/FileSaveRestoreTests.cs
bhsubra: Add the logic back in iteration 2

---
In reply to: [208039258](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208039258) [](ancestors = 208039258)

test/LibraryManager.IntegrationTest/AddClientSideLibrariesFromUITests.cs
bhsubra: updated in iteration 2 to inherit from parent class

---
In reply to: [208032751](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208032751) [](ancestors = 208032751)

test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
bhsubra: Updated in iteration 2

---
In reply to: [208033555](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208033555) [](ancestors = 208033555)

bhsubra: Updated in iteration 2

---
In reply to: [208033443](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208033443) [](ancestors = 208033443)

test/LibraryManager.IntegrationTest/VisualStudioLibmanHostTests.cs
bhsubra: If this wasn't in response to Yang's comment, let me know

---
In reply to: [208044697](https://github.com/aspnet/LibraryManager/pull/341#discussion_r208044697) [](ancestors = 208044697)

Issue #359, opened 2018-08-23
Fix the logic to get the latest version


felipepessoto:
### Functional impact
It downloads the incorrect latest version of some libraries.

### Minimal repro steps
libman install prism

### Expected result
Install prism 1.15.0, the version that is shown when I open the library url: https://cdnjs.com/libraries/prism or when calling the API, the object root has a **Version** property that should be considered, https://api.cdnjs.com/libraries/prism:
![image](https://user-images.githubusercontent.com/1336227/44548302-0b807880-a6f4-11e8-8c8c-391633e4865b.png)

### Actual result
It is installing the first version that it found in array: 9000.0.1.
https://github.com/aspnet/LibraryManager/blob/45474d37ed6977b744786e102c397b4f558a8096/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs#L194

### Further technical details
The CDNJS API has a property to flag the current version, so we should use it.
I'll try to send a fix via pull request and create a unit test to cover this scenario.

Could you assign to me?

jodavis:
@fujiy thanks for identifying this and offering to fix it.

felipepessoto:
@jodavis I'm having problem with AppVeyor, each time it runs, a different test fails, none of them related with my change. In my machine the tests are OK

jimmylewis:
Thanks @felipepessoto for submitting the fix!

PR #360, opened 2018-08-24
Get the latest version of a library in CDNJS from Version property

felipepessoto:
Fix https://github.com/aspnet/LibraryManager/issues/359

### DIFF
--- a/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
+++ b/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
@@ -190,13 +190,9 @@ namespace Microsoft.Web.LibraryManager.Providers.Cdnjs
                 return null;
             }
 
-            var versions = (await GetLibraryVersionsAsync(group.DisplayName, cancellationToken).ConfigureAwait(false)).ToList();
-            string first = versions[0];
-
-            if (!includePreReleases)
-            {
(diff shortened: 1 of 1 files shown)


fujiy:
How can I retry de CI tests?

jodavis:
Do the tests fail when run locally, or did they only fail in the AppVeyor build?

When you open the solution locally, Visual Studio's Test Explorer window should detect all the tests, and they can be executed from there.

jimmylewis:
FWIW, rebasing the PR should cause AppVeyor to run the tests again as well.

felipepessoto:
Some tests are failing because unpkg.com is returning 503

felipepessoto:
If I change the urls do https, it works:

        public const string LibraryFileListUrlFormat = "https://unpkg.com/{0}@{1}/?meta"; // http://unpkg.com/jquery@3.3.1/?meta
        public const string LatestLibraryVersonUrl = "https://unpkg.com/{0}/package.json"; // http://unpkg.com/jquery/package.json

felipepessoto:
I think unpkg is unstable today. I already rebased, so I'll need to create another PR to trigger the tests.

The tests are OK in my machine.

Do you have something like .NET team to trigger the tests manually? For example, if I sent a message like this: "@dotnet-bot test outerloop linux x64 debug build" the tests run again.

felipepessoto:
@jodavis, the tests worked on PR https://github.com/aspnet/LibraryManager/pull/429

PR #390, opened 2018-11-02
Fix serialization when using UI

jimmylewis:
This addresses a regression from #326.  When we add a library through the
UI *and* the libman.json file is open in a text buffer, we serialize the
LibraryInstallationState directly.  LibraryInstallationState is no longer annotated for JSON serialization, so this causes us to insert invalid JSON into the manifest file.

### DIFF
--- a/src/LibraryManager/Json/LibraryInstallationStateOnDisk.cs
+++ b/src/LibraryManager/Json/LibraryInstallationStateOnDisk.cs
@@ -10,11 +10,11 @@ namespace Microsoft.Web.LibraryManager.Json
 {
     internal class LibraryInstallationStateOnDisk
     {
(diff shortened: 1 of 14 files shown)


src/LibraryManager/Json/LibraryInstallationStateOnDisk.cs
```diff
@@ -10,12 +10,12 @@ namespace Microsoft.Web.LibraryManager.Json
 {
     internal class LibraryInstallationStateOnDisk
     {
-        [JsonProperty(ManifestConstants.Library)]
```
jimmylewis: The re-ordering here is so that the serialized object matches the order that one would type in the editor (provider has to be first in order to get completion for remaining properties).  This also matches the screenshots we have in the wiki documentation.

Issue #378, opened 2018-09-27
[CLI] libman init: accept defaults


scottaddie:
I propose an enhancement to the `libman init` CLI command, such that a `-y` option is supported. By running `libman init -y`, all default settings are accepted and I'm no longer prompted for a DefaultProvider value. An example of where this is possible today is npm: `npm init -y`.

justcla:
Great suggestion, Scott. Thanks. We will look to add something like this in a future release.

PR #410, opened 2018-12-17
Disables interactive mode on consoles with redirected input

prafullbhosale:
Fixes the issue where using the interactive commands (`libman init`/ `libman install`) appear to be hung if the console's input is redirected.
Visual studio's Package manager console is one such case.

### DIFF
--- a/src/libman/Commands/BaseCommand.cs
+++ b/src/libman/Commands/BaseCommand.cs
@@ -170,6 +170,7 @@ namespace Microsoft.Web.LibraryManager.Tools.Commands
         protected async Task<Manifest> CreateManifestAsync(string defaultProvider,
             string defaultDestination,
             EnvironmentSettings settings,
+            string providerOptionString,
             CancellationToken cancellationToken)
         {
             if (File.Exists(Settings.ManifestFileName))
@@ -181,9 +182,13 @@ namespace Microsoft.Web.LibraryManager.Tools.Commands
             manifest.AddVersion(Manifest.SupportedVersions.Last().ToString());
             manifest.DefaultDestination = string.IsNullOrEmpty(defaultDestination) ? null : defaultDestination;
 
-            defaultProvider = string.IsNullOrEmpty(defaultProvider)
-                ? HostEnvironment.InputReader.GetUserInputWithDefault(nameof(settings.DefaultProvider), settings.DefaultProvider)
-                : defaultProvider;
+            if (string.IsNullOrEmpty(defaultProvider))
+            {
+                defaultProvider = GetUserInputWithDefault(
+                    fieldName: nameof(settings.DefaultProvider),
+                    defaultFieldValue: settings.DefaultProvider,
+                    optionLongName: providerOptionString);
+            }
 
             if (ManifestDependencies.GetProvider(defaultProvider) == null)
             {
@@ -214,6 +219,21 @@ namespace Microsoft.Web.LibraryManager.Tools.Commands
             foreach (IError error in errors)
             {
                 Logger.Log(string.Format("[{0}]: {1}", error.Code, error.Message), LogLevel.Error);
+            }
+        }
+
+        protected string GetUserInputWithDefault(string fieldName, string defaultFieldValue, string optionLongName)
+        {
+            try
+            {
+                // If the console's input is being redirected, we cannot read user input interactively.
+                // All the inputs for this case should come from the commandline directly.
+                return HostEnvironment.InputReader.GetUserInputWithDefault(fieldName, defaultFieldValue);
+            }
+            catch (InvalidOperationException)
+            {
+                HostEnvironment.Logger.Log(string.Format(Resources.Text.SpecifyFieldUsingOption, fieldName, optionLongName), LogLevel.Error);
+                throw;
             }
         }
     }
--- a/src/libman/Contracts/ConsoleLogger.cs
+++ b/src/libman/Contracts/ConsoleLogger.cs
@@ -28,7 +28,9 @@ namespace Microsoft.Web.LibraryManager.Tools.Contracts
         /// <returns></returns>
         public string GetUserInput(string fieldName)
         {
-            lock(_syncObject)
+            ThrowIfInputIsRedirected();
+
+            lock (_syncObject)
             {
                 Console.Out.Write($"{fieldName}: ");
                 return Console.ReadLine();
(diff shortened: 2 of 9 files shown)


src/libman/Resources/Text.resx
```diff
@@ -132,6 +132,9 @@
   <data name="DefaultProviderOptionDesc" xml:space="preserve">
     <value>The provider to use if no provider is defined for a given library. (eg. “cdnjs”, “filesystem”)</value>
   </data>
+  <data name="NonInteractiveConsoleMessage" xml:space="preserve">
+    <value>Cannot start an interactive session. Please specify all required arguments on the commandline.</value>
```
jimmylewis: This message is not actionable: what are the arguments I have to specify?  I can see this being a frustrating user experience.

Can you add a list of arguments?  Or I'd even be fine if you rephrased to the one argument at a time, which might be easier depending on how the command is processed, e.g. "Cannot start an interactive session. Please specify a value for the {--foo} argument on the command line."  The user would have to iterate if multiple were required, but each one would be clearly actionable.

src/libman/Commands/BaseCommand.cs
```diff
@@ -216,5 +217,18 @@ protected void LogErrors(IEnumerable<IError> errors)
                 Logger.Log(string.Format("[{0}]: {1}", error.Code, error.Message), LogLevel.Error);
             }
         }
+
+        protected string GetUserInputWithDefault(string fieldName, string defaultFieldValue)
```
jimmylewis: This seems like it belongs on the `HostEnvironment` or the `IInputReader` implementation.  Whether the input is redirected is an environmental condition, right?  Feels a little odd to push that up to this level.

Why not have `ConsoleLogger` handle this within itself, since it is console specific?

src/libman/Resources/Text.resx
prafullbhosale: This message is shown after `ShowHelp()` which will show the arguments and options for the command being run.

src/libman/Commands/BaseCommand.cs
prafullbhosale: The only reason for doing the check here is to be able to show help message to the user before showing the error message. 

Other way is to throw an exception from the ConsoleLogger and catch that here and then show help.

jimmylewis: Are there other possible cases (aside from `CreateManifestAsync`) where we might hit the same issue with redirected input?  I.e. will we have to make this fix later in multiple places?

prafullbhosale: Right now the only two possible scenarios are Init and Install command. 
Both are addressed.

The only other possibility is if in future libman allows multiple libraries with the same library name to be installed in the manifest (to resolve the library by asking the user to provide a choice).

src/libman/Resources/Text.resx
jimmylewis: It doesn't seem clear to me from `libman init --help` which parameters are required (mostly because none of them are required).  I don't think relying upon `ShowHelp()` gives the best experience here.

src/libman/Commands/BaseCommand.cs
jimmylewis: Along the lines of my comment below... what if I don't want a default provider?  Doesn't this just throw now?

prafullbhosale: The current experience is designed to not allow manifests without a default provider. 
User can go ahead and delete it from the file, but the next install command will set whatever provider it uses to be the default one.

src/libman/Resources/Text.resx
jodavis: What if it was "Please specify all arguments on the command line"? Does that read as "all the arguments you want to specify", or does it make you think that every single argument is required whether you want it or not?

---
In reply to: [242353043](https://github.com/aspnet/LibraryManager/pull/410#discussion_r242353043) [](ancestors = 242353043)

jodavis: >interactive session [](start = 27, length = 19)

The phrase "interactive session" seems too technical. Maybe something like, "Unable to read input in this console" or something that explains what the problem is from the user's perspective.

src/libman/Commands/BaseCommand.cs
jodavis: I think we should update the disambiguating codepath to use this as well, even if it can't be hit today. If we do re-enable it in the future, it should work right.

prafullbhosale: The disambiguation code path now handles this case and shows the below message: 
```
Cannot prompt for user input in this console.
Please specify the complete library id and the also the provider using '--provider' option.
```

src/libman/Resources/Text.resx
prafullbhosale: Changed the messaging to be: 
```
libman init
libman : Cannot prompt for user input in this console.
Please specify 'DefaultProvider' using the '--default-provider' option.
```

The field name and option string are provided by the caller.

src/libman/Commands/BaseCommand.cs
prafullbhosale: Even if we implement #378, users will still face this issue if they don't pass in `-y` (or whatever we decide to call the option) on an input redirected console. 

I have moved the check to be handled by the `ConsoleLogger` which will throw if the input is redirected and the exception is being handled by the callers.

src/libman/Resources/Text.resx
prafullbhosale: Changed the messaging to show the field name and option string.

For instance on running libman init the user will see the below: 
```
Cannot prompt for user input in this console.
Please specify 'DefaultProvider' using the '--default-provider' option.
```

src/libman/Contracts/ConsoleLogger.cs
```diff
@@ -84,5 +88,13 @@ public void Log(string message, LogLevel level)
                 }
             }
         }
+
+        private void ThrowIfInputIsRedirected()
+        {
+            if (Console.IsInputRedirected)
+            {
+                throw new InvalidOperationException();
```
jimmylewis: nit: maybe have a message in the exception, so that if someone does catch this while debugging, they know what it's from?

prafullbhosale: Fixed. 
Also squashed the changes to a single commit.

PR #429, opened 2019-01-10
Get the latest version of a library in CDNJS from Version property

### LINKED ISSUE #359: Fix the logic to get the latest version

felipepessoto:
Fix #359

### DIFF
--- a/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
+++ b/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
@@ -190,13 +190,9 @@ namespace Microsoft.Web.LibraryManager.Providers.Cdnjs
                 return null;
             }
 
-            var versions = (await GetLibraryVersionsAsync(group.DisplayName, cancellationToken).ConfigureAwait(false)).ToList();
-            string first = versions[0];
-
-            if (!includePreReleases)
-            {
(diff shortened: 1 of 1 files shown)


felipepessoto:
Finally :) I got the tests running without problem

felipepessoto:
In which VS version is this fix?

jimmylewis:
@felipepessoto based on history, this would have been in the VS 16.0 release or the 2.0.48 version of the nuget packages.

felipepessoto:
The issue is still happening on update menu:

image

felipepessoto:
I created a new issue: https://github.com/aspnet/LibraryManager/issues/633

PR #437, opened 2019-02-01
Add Fxcop Analyzers and fix errors/warnings

jimmylewis:
Added the FxCop analyzer and a copy of the SDL required ruleset.  I had to remove a couple of the rules as they come from modules that I can't find; however, they looked like areas that are generally not applicable to LibMan (cryptography and XML parsing mostly).


src/LibraryManager.Contracts/CompletionSet.cs
```diff
@@ -33,13 +37,65 @@ public struct CompletionSet
         /// The type of the completion item sorting.
         /// </summary>
         public CompletionSortOrder CompletionType;
+
+        /// <summary>
+        /// Returns whether the objects are equal.
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
```
bhsubra: nit: Can we update description here and the next one?

PR #440, opened 2019-02-01
Add retry loop when downloading content from providers.

jimmylewis:
This attempts to address severe flakiness we've seen from all providers,
as well as in executing our unit tests.  Many of the failures we see are
intermittent 404s, so the loop gives us better resiliency when CDNs are
being flaky.

### DIFF
--- a/src/LibraryManager/CacheService/CacheService.cs
+++ b/src/LibraryManager/CacheService/CacheService.cs
@@ -53,15 +53,37 @@ namespace Microsoft.Web.LibraryManager
         /// <summary>
         /// Downloads a resource from specified url to a destination file
         /// </summary>
-        /// <param name="url"></param>
-        /// <param name="fileName"></param>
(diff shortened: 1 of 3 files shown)


src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -53,15 +53,37 @@ public async Task<string> GetMetadataAsync(string url, string cacheFile, Cancell
...
+                    using (Stream libraryStream = await _requestHandler.GetStreamAsync(url, cancellationToken))
+                    {
+                        await FileHelpers.SafeWriteToFileAsync(fileName, libraryStream, cancellationToken);
+                        break;
+                    }
+                }
+                catch (ResourceDownloadException)
+                {
+                    // rethrow last exception
+                    if(i == attempts - 1)
```
bhsubra: nit: space between if and (

Issue #348, opened 2018-08-09
Localize the text in "Check for Updates" lightbulb action (Stable/Pre-release)


justcla:
**Repro Steps:**
When editing within with a libman.json file, right-click on a library (that has newer versions available).
Expand the Updates menu.

**Expected:**
localized version of text for "Stable" and "Pre-release" will show.

**Actual:**
American English version is shown.

**Links to hard-coded text strings:**

"Stable"
https://github.com/aspnet/LibraryManager/blob/45474d37ed6977b744786e102c397b4f558a8096/src/LibraryManager.Vsix/Json/SuggestedActions/UpdateSuggestedActionSet.cs#L87

"Pre-release"
https://github.com/aspnet/LibraryManager/blob/45474d37ed6977b744786e102c397b4f558a8096/src/LibraryManager.Vsix/Json/SuggestedActions/UpdateSuggestedActionSet.cs#L97

PR #444, opened 2019-02-04
Refactor hardcoded strings into resources for Update lightbulb action

jimmylewis:
This addresses #348 after the subsequent loc handback.

### DIFF
--- a/src/LibraryManager.Vsix/Resources/Text.resx
+++ b/src/LibraryManager.Vsix/Resources/Text.resx
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
     Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
-    various data types are done through the TypeConverter classes 
+
+    The primary goals of this format is to allow a simple XML format
+    that is mostly human readable. The generation and parsing of the
+    various data types are done through the TypeConverter classes
     associated with the data types.
-    
+
     Example:
-    
+
     ... ado.net/XML headers & schema ...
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Vsix/Resources/Text.resx
```diff
@@ -235,4 +235,14 @@ Do you want to continue?</value>
   <data name="Install" xml:space="preserve">
     <value>_Install</value>
   </data>
+  <data name="SuggestedAction_Update_Prerelease" xml:space="preserve">
+    <value>Pre-release: {0}</value>
+  </data>
+  <data name="SuggestedAction_Update_Stable" xml:space="preserve">
+    <value>Stable: {0}</value>
+  </data>
+  <data name="SuggestedAction_Update_Title" xml:space="preserve">
```
bhsubra: Should we call it SuggestedAction_Update_Library to keep it consistent with value and other entries above?

src/LibraryManager.Vsix/Resources/Text.resx
```diff
@@ -235,4 +235,14 @@ Do you want to continue?</value>
   <data name="Install" xml:space="preserve">
     <value>_Install</value>
   </data>
+  <data name="SuggestedAction_Update_Prerelease" xml:space="preserve">
```
bhsubra: I'm guessing this is not for preview 4 as we need another loc pass.

src/LibraryManager.Vsix/Resources/Text.resx
jimmylewis: I wanted to differentiate the use case for this one.  It's the title for the SuggestedActionSet, and isn't actually visible in the UI (though the SuggestedAction API might expose it somewhere).  Since context matters for translations, I didn't want this to seem like a string that could be reused elsewhere.

src/LibraryManager.Vsix/Resources/Text.resx
jimmylewis: That's true.  I can hold this one back for now.

PR #446, opened 2019-02-05
Fixing permissions issue in local IIS webs

alexgav:


src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -169,7 +169,11 @@ public static async Task<bool> MoveFileAsync(string sourceFile, string destinati
                         string directoryName = Path.GetDirectoryName(destinationFile);
                         Directory.CreateDirectory(directoryName);
 
-                        File.Move(sourceFile, destinationFile);
+                        // Do not use File.Move, as it causes destination not to inherit security permissions.
+                        // Our IIS projects set ACL to give AppPool user permissions, and it gets inherited by project files.
+                        // See https://stackoverflow.com/questions/2929918/file-move-does-not-inherit-permissions-from-target-directory
```
jimmylewis: I find this comment misleading, as the solution in the StackOverflow link is to do this instead:

```
var fs = File.GetAccessControl(destination);
fs.SetAccessRuleProtection(false, false);
File.SetAccessControl(destination, fs);
```

Is there any guarantee that `File.Copy` doesn't copy the source ACLs as well?  Or could you add a unit test to show that the ACL matches the destination folder?

alexgav: Changed the comment to point to https://support.microsoft.com/en-us/help/310316/how-permissions-are-handled-when-you-copy-and-move-files-and-folders

By default, an object inherits permissions from its parent object, either at the time of creation or when it is copied or moved to its parent folder. The only exception to this rule occurs when you move an object to a different folder on the same volume. In this case, the original permissions are retained.

jimmylewis:
Please remember to squash the commits when you merge (or before).

alexgav:
Thanks Jimmy!

PR #448, opened 2019-02-20
Fixing 790461 - Libman Install Dialog Install button crashes VS

bzhu94:
Based on [Bug 790461](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/790461) it seems like `            bool isLibraryInstallationStateValid = await IsLibraryInstallationStateValidAsync().ConfigureAwait(false);
` on line 218 is switching to a background thread after it returns. In the `else` clause below we are accessing `ViewModel` which is a dependency property that requires us to be on the UI thread to access, which crashes VS.

### DIFF
--- a/src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
+++ b/src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
@@ -143,17 +143,11 @@ namespace Microsoft.Web.LibraryManager.Vsix
                     }
                 }
 
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
                 UI.InstallDialog dialog = new UI.InstallDialog(dependencies, _libraryCommandService, configFilePath, target, rootFolder, project);
 
-                var dte = (DTE)Package.GetGlobalService(typeof(SDTE));
-                int hwnd = dte.MainWindow.HWnd;
-                WindowInteropHelper windowInteropHelper = new WindowInteropHelper(dialog);
-
-                // Set visual studio window's handle as the owner of the dialog.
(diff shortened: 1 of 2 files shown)


jimmylewis:
The 2nd commit ought to be in a separate PR, as it addresses a separate issue, no?

src/LibraryManager.Vsix/Commands/InstallLibraryCommand.cs
```diff
@@ -143,17 +143,11 @@ private async Task ExecuteAsync(object sender, EventArgs e)
                     }
                 }
 
-                UI.InstallDialog dialog = new UI.InstallDialog(dependencies, _libraryCommandService, configFilePath, target, rootFolder, project);
-
-                var dte = (DTE)Package.GetGlobalService(typeof(SDTE));
-                int hwnd = dte.MainWindow.HWnd;
-                WindowInteropHelper windowInteropHelper = new WindowInteropHelper(dialog);
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
BillHiebert: Since the if (project == null) case and this one both switch to the UI thread, move one of them above the if and get rid of the other one.

bzhu94: Looks like this SwitchToMainThreadAsync isn't even needed since we're in the (project != null) block which has SwitchToMainThreadAsync already

bzhu94:
> The 2nd commit ought to be in a separate PR, as it addresses a separate issue, no?

bzhu94:
The 2nd commit addresses a bug that appeared from the 1st commit, so keeping 1 PR should be fine.

PR #450, opened 2019-02-26
Remove : from labels

bhsubra:
Remove : from labels as we don't want that to be read by the narrator.

### DIFF
--- a/src/LibraryManager.Vsix/Resources/Text.Designer.cs
+++ b/src/LibraryManager.Vsix/Resources/Text.Designer.cs
@@ -178,7 +178,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Resources {
         }
         
         /// <summary>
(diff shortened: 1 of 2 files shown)


jimmylewis:
Labels for field should end with a colon, as called out in the VS design standards (https://vsuiguide.azurewebsites.net/topics/07%20Interaction%20Patterns/Basic%20control%20patterns/07.10.04%20Text%20controls.html)

bhsubra:
> Labels for field should end with a colon, as called out in the VS design standards (https://vsuiguide.azurewebsites.net/topics/07%20Interaction%20Patterns/Basic%20control%20patterns/07.10.04%20Text%20controls.html)

@jimmylewis , thanks for sharing the document. Looks like we don't need it for radio button and checkbox. Let me know if you disagree.

bhsubra:
Had an offline discussion with @jimmylewis . This is by design and we don't need to change anything here.

PR #453, opened 2019-02-27
Fix to make TreeViewItems non tabbable

bhsubra:
As a part of this change verified below:
- Made sure tree view items are non tabbable. User can tab into tree view and use up/down arrow keys to navigate
- Home and end key press events work as expected


bhsubra:
Closing this pr. Sent out another pr which includes changes in this pr plus fixes for few other accessibility issues in PackageContentsTreeView control.

PR #454, opened 2019-02-28
Changes to remove _ from library, target location and provider names when narrator is on

bhsubra:
_ was added when access keys were implemented. This seems to interfere with narrator. Made changes to strip _ for automation names.

How was this tested:
- Manual testing
- Unit and integration tests

### DIFF
--- a/src/LibraryManager.Vsix/UI/Extensions/RemoveSubstringExtension.cs
+++ b/src/LibraryManager.Vsix/UI/Extensions/RemoveSubstringExtension.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Extensions
         public RemoveSubstringExtension(object text, string remove)
         {
             _text = text;
-            _remove = remove;
+            _remove = remove ?? "_";
         }
 
         public override object ProvideValue(IServiceProvider serviceProvider)
@@ -27,7 +27,6 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Extensions
                 return null;
             }
 
-            string remove = _remove ?? "_";
             string s = null;
 
             MarkupExtension m = _text as MarkupExtension;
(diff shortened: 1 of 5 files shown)


src/LibraryManager.Vsix/UI/Extensions/RemoveSubstringExtension.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Windows.Markup;
+
+namespace Microsoft.Web.LibraryManager.Vsix.UI.Extensions
+{
+    [MarkupExtensionReturnType(typeof(string))]
+    internal class RemoveSubstringExtension : MarkupExtension
```
jimmylewis: I'm wondering if we should rename this and also tweak the design.  Right now it removes a substring (by default, a single character, but it can take an arbitrary string).  Based on our offline conversation, it seems like it would more useful in the case we need if it would remove any character in that string.  E.g. if I wanted to remove the angle brackets in `<foo>`, the current implementation would require that I call this separately for `<` and `>`.

What do you think about renaming to `RemoveCharacterExtension` instead, and change the `string.Replace` call on line 46 with a loop over each of the characters in `_remove`?  `RemoveSubstring` could be a useful operation, but it's not serving our current need.

src/LibraryManager.Vsix/UI/Extensions/RemoveSubstringExtension.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public override object ProvideValue(IServiceProvider serviceProvider)
+        {
+            if (_text == null)
+            {
+                return null;
+            }
+
+            string remove = _remove ?? "_";
```
jimmylewis: I would do this in the constructor instead of here.

bhsubra: Updated here:
https://github.com/aspnet/LibraryManager/commit/fb24f8124cac6e0f58dd12ded43b8c1b99029fbb

src/LibraryManager.Vsix/UI/Extensions/RemoveSubstringExtension.cs
bhsubra: I'll be sending out another pr with the suggested design changes.

PR #456, opened 2019-03-04
Fix accessibility issue around PackageContentsTreeView control

bhsubra:
Fixed following issues with this pr:
1. Made sure tree view items are non tabbable. User can tab into tree view and use up/down arrow keys to navigate
2. Home and end key press events work as expected
3. When treeview items are selected(via keyboard commands), screen reader should narrate checkbox status, folder status(expanded/collapsed), file name
4. When in choose specific mode , hitting right arrow key should expand node and left arrow should collapse node
5. When selected, treeview item text foreground color should be white
6. Hitting space should toggle treeview item selection

### DIFF
--- a/src/LibraryManager.Vsix/UI/Controls/PackageContentsTreeView.xaml
+++ b/src/LibraryManager.Vsix/UI/Controls/PackageContentsTreeView.xaml
@@ -10,11 +10,11 @@
              d:DesignHeight="300" d:DesignWidth="300" d:DataContext="{d:DesignInstance IsDesignTimeCreatable=False, Type=models:InstallDialogViewModel}">
     <UserControl.Resources>
         <converters:BoldingConverter x:Key="BoldingConverter" />
+        <converters:CheckBoxAutomationNameConverter x:Key="CheckBoxAutomationNameConverter" />
         <converters:PackageItemIconConverter x:Key="IconConverter" />
         <converters:VisibleIfNonNullConverter x:Key="VisibleIfNonNullConverter" />
     </UserControl.Resources>
-    <TreeView AutomationProperties.Name="{x:Static resources:Text.Files}"
-              ItemsSource="{Binding Path=DisplayRoots, Mode=OneWay}"
+    <TreeView ItemsSource="{Binding Path=DisplayRoots, Mode=OneWay}"
               VirtualizingPanel.IsVirtualizing="True"
               KeyboardNavigation.TabNavigation="Continue"
               PreviewKeyUp="OnPreviewKeyUp">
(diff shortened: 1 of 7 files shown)


src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+            if (values == null || values.Length != 2 || !(values[0] is string) || !(values[1] is bool))
+            {
+                return null;
+            }
+
+            bool isChecked = (bool)values[1];
+
+            if (isChecked)
+            {
+                return "Checked" + (string)values[0];
```
jimmylewis: Do these strings need to be localized? (honestly not sure)

src/LibraryManager.Vsix/UI/Controls/PackageContentsTreeView.xaml
```diff
@@ -10,19 +10,38 @@
...
         <TreeView.ItemContainerStyle>
             <Style TargetType="{x:Type TreeViewItem}" d:DataContext="{d:DesignInstance Type=TreeViewItem, IsDesignTimeCreatable=False}">
                 <Setter Property="IsExpanded" Value="{Binding Path=IsExpanded, Mode=TwoWay}" />
-                <Setter Property="Focusable" Value="False" />
-                <Setter Property="KeyboardNavigation.TabNavigation" Value="Continue"/>
-                <Setter Property="AutomationProperties.Name" Value="{Binding Name}"/>
+                <Setter Property="KeyboardNavigation.TabNavigation" Value="Once"/>
+                <Setter Property="AutomationProperties.Name">
+                    <Setter.Value>
+                        <MultiBinding Converter="{StaticResource CheckBoxAutomationNameConverter}">
```
jimmylewis: It seems weird that the state of the checkbox is bound into the name.  Is this a hack to workaround the control not being focusable?

src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
bhsubra: I am not sure. Tagging @jodavis

src/LibraryManager.Vsix/UI/Controls/PackageContentsTreeView.xaml
bhsubra: The issue here is the treeview item has a checkbox and label bound to it. We can't give focus to checkbox as it would take additional key press event and every keyUp, keyDown would take two key press instead of one. Let me know if you have any suggestions.

src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
jodavis: Yes I think they do, since they'll be read. I'm trying to see if there's a way to access the localized values that would come from the Checkbox by default. (Can you bind value[1] to the Checkbox's automation name, instead of its IsChecked property?)

Otherwise, let's make it localizable, but we won't wait on the loc handback to push the changes.

jodavis: At a minimum, there needs to be a space after "Checked" and "Unchecked", or else they become "Checkedbootstrap"

PR #466, opened 2019-03-30
Set http client to use default proxy settings

### LINKED ISSUE #399: How to use behind proxy that require authentication?
### Minimal repro steps
Run the following command in console

`libman install bootstrap -d wwwroot/lib/bootstrap --files dist/css/bootstrap.css --files dist/css/bootstrap.min.css --files dist/js/bootstrap.js --files dist/js/bootstrap.min.js`

### Expected result
Download the library files

### Actual result
Microsoft.Web.LibraryManager.Contracts.ResourceDownloadException: Failed to download resource from "http://unpkg.com/bootstrap/package.json"
   at Microsoft.Web.LibraryManager.WebRequestHandler.GetStreamAsync(String url, CancellationToken cancellationToken)
   at Microsoft.Web.LibraryManager.Extensions.GetJsonObjectViaGetAsync(IWebRequestHandler webRequestHandler, String url, CancellationToken cancellationToken)   at Microsoft.Web.LibraryManager.Providers.Unpkg.UnpkgCatalog.GetLatestVersion(String libraryName, Boolean includePreReleases, CancellationToken cancellationToken)
The "bootstrap" library could not be resolved by the "unpkg" provider

### Further technical details

- I am behind corporate proxy that require  authentication
- The solution that I am currently using is activating traffic capture in Fiddler while I am using libman

jimmylewis:
By default, HttpMessageHandler seems to not inherit the default proxy settings from the OS.  Setting them to the default from WebRequest seems to fix #399 for me.

I tested this using Fiddler as a proxy with basic auth, but I don't have a configuration to test with NTLM proxy authentication.  Hopefully this will "Just Work" like others have claimed it did for them.

### DIFF
--- a/src/LibraryManager/CacheService/WebRequestHandler.cs
+++ b/src/LibraryManager/CacheService/WebRequestHandler.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
@@ -19,6 +20,7 @@ namespace Microsoft.Web.LibraryManager
         private WebRequestHandler()
         {
             HttpClientHandler httpMessageHandler = new HttpClientHandler();
+            httpMessageHandler.Proxy = WebRequest.DefaultWebProxy;
             _httpClient = new HttpClient(httpMessageHandler);
         }
 


Roman-Tarasiuk:
Hi, @jimmylewis!
I tried code you suggest in the pull request:
<code>httpMessageHandler.Proxy = WebRequest.DefaultWebProxy;</code>
It does not work for me.
So you have to decide to use also another option for manual settings of the proxy cridentials as described in https://github.com/aspnet/LibraryManager/issues/399#issuecomment-472760343

jimmylewis:
After futher consideration, and comparing with now other tools handle this, I think it would actually be better to go down a route of letting users configure the proxy settings and save them.  It looks like NPM and NuGet both do this, which seems like decent prior art.  I'll try again with that approach.

PR #471, opened 2019-04-13
Code cleanup

jimmylewis:
1. Addressing all IDE style suggestions and warnings that show up in code files
2. Added a code style to .editorconfig for static members and made all
statics consistent (previously we split between two different styles)
3. Added copyright headers everywhere that I found them missing
4. Addressed any analyzer warnings that were present
5. Cleaned up extraneous and unsorted usings


jimmylewis:
Closing this as it was too broad in a single change.

Issue #306, opened 2018-07-12
Uninstall of filesystem libraries with UNC paths does not work


raquelsa:
We are failing to compare escaped directory delimeters in a UNC path due to differences on how Newtonsoft and Microsoft.JSON.Core.Parser parses the escaped chars. 

Manifest.UninstallAsync fails at line367: 
ILibraryInstallationState library = Libraries.FirstOrDefault(l => l.LibraryId == libraryId); 

Comparisson below fails: 
\\FileShare\folder\
\\\\FileShare\\folder\\

PR #477, opened 2019-05-05
Fix deserialization of backslashes from VS editor

### LINKED ISSUE #363: Uninstall of filesystem library fails in VS
### Functional impact
In Visual Studio, I'm unable to uninstall a library that's using the _filesystem_ provider.

### Minimal repro steps

1. Create a new ASP.NET Core Razor Pages app targeting .NET Core 2.1.
2. Add a *libman.json* file to the project root with the following contents:

```json
{
  "version": "1.0",
  "defaultProvider": "cdnjs",
  "libraries": [
    {
      "provider": "filesystem",
      "library": "C:\\tmp\\contosoCalendar",
      "files": [
        "calendar.js",
        "calendar.css"
      ],
      "destination": "wwwroot/lib/contosoCalendar/"
    }
  ]
}
```

3. With the *libman.json* file open in the editor, position the caret in the left margin. Click the light bulb icon that appears, and select the *Uninstall C:\\tmp\\contosoCalendar* option.
4. Notice that the object literal inside of *libman.json*'s `libraries` property is removed. The files in the project aren't removed.

### Expected result
The object literal inside of the `libraries` property of *libman.json* should be removed. The library files should be deleted from the configured location in the project.

### Actual result
The **Library Manager** feed of Visual Studio's **Output** window displays the following:

```console
Uninstall of library C:\\tmp\\contosoCalendar started...
Uninstall of library C:\\tmp\\contosoCalendar failed
Object reference not set to an instance of an object.
```

### Further technical details
* VS 2017 Enterprise v15.8.1
* .NET Core SDK v2.1.401
* LibMan CLI v1.0.163+g45474d37ed

jimmylewis:
When we deserialize these values using VS' JSON parser, it adds an extra
level of escaping to the backslashes.  We need to compensate for this when we
compare with values from the LibMan manifest.

This fixes #363 and #306.

### DIFF
--- a/src/LibraryManager.Vsix/Json/JsonHelpers.cs
+++ b/src/LibraryManager.Vsix/Json/JsonHelpers.cs
@@ -94,21 +94,32 @@ namespace Microsoft.Web.LibraryManager.Vsix
 
             SortedNodeList<Node> children = GetChildren(parent);
 
+            string GetCanonicalizedValue(BlockChildNode m)
+            {
+                if (m.Value is TokenNode value)
+                {
+                    return value.GetCanonicalizedText();
+                }
+                return m.UnquotedValueText;
+            }
+
             foreach (MemberNode child in children.OfType<MemberNode>())
             {
+                // note: the Json parser escapes backslashes in the node's Value text,
+                // so we need to unescape those so that they match the value from the manifest.
                 switch (child.UnquotedNameText)
                 {
                     case ManifestConstants.Provider:
-                        state.ProviderId = child.UnquotedValueText;
+                        state.ProviderId = GetCanonicalizedValue(child);
                         break;
                     case ManifestConstants.Library:
-                        state.LibraryId = child.UnquotedValueText;
+                        state.LibraryId = GetCanonicalizedValue(child);
                         break;
                     case ManifestConstants.Destination:
-                        state.DestinationPath = child.UnquotedValueText;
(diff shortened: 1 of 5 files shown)


src/LibraryManager.Vsix/Json/JsonHelpers.cs
```diff
@@ -96,19 +96,21 @@ public static bool TryGetInstallationState(ObjectNode parent, out ILibraryInstal
 
             foreach (MemberNode child in children.OfType<MemberNode>())
             {
+                // note: the Json parser escapes backslashes in the node's Value text,
+                // so we need to unescape those so that they match the value from the manifest.
                 switch (child.UnquotedNameText)
                 {
                     case ManifestConstants.Provider:
-                        state.ProviderId = child.UnquotedValueText;
+                        state.ProviderId = UnescapeSlashes(child.UnquotedValueText);
```
alexgav: >UnescapeSlashes(child.UnquotedValueText) [](start = 43, length = 40)

Have you looked at using GetCanonicalizedText on the value token here instead? I'm pretty sure it should give you the unescaped text we want.

jimmylewis: Sounds perfect, except that it's not exposed publicly from the JSON parser code.

jimmylewis: Figure out how to call it, thanks for the pointer.

PR #484, opened 2019-06-04
[UI] Fix issue with cancel button for unpkg provider

bhsubra:
Issue description:
- Invoke libman UI
- Select ‘unpkg’ in the provider combo
- Type ‘@’ character in the ‘Library’ text box.
- Hit cancel button

You will notice that dialog doesn't close. Made a change as a part of this pr to not throw null ref exception and close the dialog when cancel button is hit.

How was this tested?
- Manual testing
- Ran existing unit and integration tests
- Added unit test for this scenario

### DIFF
--- a/test/LibraryManager.Test/Providers/Unpkg/UnpkgCatalogTest.cs
+++ b/test/LibraryManager.Test/Providers/Unpkg/UnpkgCatalogTest.cs
@@ -154,7 +154,7 @@ namespace Microsoft.Web.LibraryManager.Test.Providers.Unpkg
         }
 
         [TestMethod]
-        public async Task GetLibraryCompletionSetAsync_WithNullNpmPackageInfoVersions()
+        public async Task GetLibraryCompletionSetAsync_WithNullNpmPackageInfoVersions_ReturnsNoCompletions()
         {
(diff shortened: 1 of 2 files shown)


jimmylewis:
Also, can you add tests for this scenario?  It might take some refactoring, but we should be able to test this scenario pretty easily.

test/LibraryManager.Test/Providers/Unpkg/UnpkgCatalogTest.cs
```diff
@@ -154,5 +152,16 @@ public async Task GetLatestVersion_PreRelease()
                 Assert.AreNotEqual(existing[1], result);
             }
         }
+
+        [TestMethod]
+        public async Task GetLibraryCompletionSetAsync_WithNullNpmPackageInfoVersions()
```
jimmylewis: just a nit: I've been using the naming scheme `<MethodBeingTested>_<DescribeInputScenario>_<DescribeExpectedResult>` where the last portion usually begins with a verb like Return or Should.  That way it's clear simply from the test name what the intent of all of the Assert statements are, and you can often tell which scenario is broken without even looking at the test implementation.

PR #487, opened 2019-06-06
[UI] Fix accessibility issues in treeView

bhsubra:
Changes in this pr include:
- Tabbing into treeView will not set focus and select the first item 
- Updated CheckBoxNameAutomationConverter to include Intermediate state of checkBox. Also made changes so that narrator specifies whether folder/file is selected
- Reduce border thickness
- TreeView should is not accessible when IncludeAllLibraryFiles option is selected
- Added new resource strings
- Added unit tests for CheckBoxNameAutomationConverter

### DIFF
--- a/src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
+++ b/src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
@@ -9,10 +9,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Converters
     {
         public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
         {
-            if (values == null || values.Length != 3 || !(values[0] is string))
-            {
-                return null;
-            }
+            ValidateInputs(values);
 
             if (values[2] == null)
             {
@@ -28,6 +25,24 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Converters
             else
             {
                 return string.Format(Text.UnChecked, (string)values[0], (string)values[1]);
+            }
+        }
+
+        private void ValidateInputs(object[] values)
+        {
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            if (values.Length != 3)
+            {
+                throw new ArgumentOutOfRangeException(nameof(values), $"{nameof(values)} length must be 3");
+            }
+
+            if (!(values[0] is string))
+            {
+                throw new ArgumentException(string.Format("{0} is not a string", values[0]));
             }
         }
 
--- a/src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
+++ b/src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
@@ -9,9 +9,14 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Converters
     {
         public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
         {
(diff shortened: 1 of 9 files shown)


test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Converters/CheckBoxAutomationNameConverterTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.Web.LibraryManager.Vsix.Resources;
+using Microsoft.Web.LibraryManager.Vsix.UI.Converters;
+
+namespace Microsoft.Web.LibraryManager.Vsix.Test.UI.Converters
+{
+    [TestClass]
+    public class CheckBoxAutomationNameConverterTests
+    {
```
jodavis: All the happy paths are tested here, but this could use some more tests of unexpected inputs.

src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
```diff
@@ -62,16 +62,6 @@ private void OnLoaded(object sender, RoutedEventArgs e)
             FocusManager.SetFocusedElement(LibrarySearchBox, LibrarySearchBox);
         }
 
-        private void IncludeAllLibraryFilesRb_Checked(object sender, RoutedEventArgs e)
```
jodavis: What code is responsible now for making sure the checkboxes remain checked when "IncludeAllLibraryFiles" is selected?

jimmylewis: I was going to flag this too - seems like it should be a binding, not an event handler.

src/LibraryManager.Vsix/UI/Converters/CheckBoxAutomationNameConverter.cs
```diff
@@ -9,20 +9,25 @@ internal class CheckBoxAutomationNameConverter : IMultiValueConverter
     {
         public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
         {
-            if (values == null || values.Length != 2 || !(values[0] is string) || !(values[1] is bool))
+            if (values == null || values.Length != 3 || !(values[0] is string))
             {
                 return null;
```
jimmylewis: Converters should return `DependencyObject.UnsetValue` when the inputs are invalid in an anticipated way, or throw otherwise (see [here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.data.imultivalueconverter.convert?view=netframework-4.8#remarks)).  Returning null doesn't signal that there was anything wrong with the converter inputs, so it would be harder to track down that's something's going wrong.

jimmylewis: I kinda lean towards throwing if the arguments are not in the right order or missing, but you could go either way with it.  It's handy to see the exceptions being thrown when running under a debugger if a control doesn't behave as expected.

bhsubra: I like throwing exceptions better. Made changes here:
https://github.com/aspnet/LibraryManager/commit/0d5e1914b17d0fe7853d0b50b666ee02aede15c4

src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
bhsubra: IsChecked is bound to treeViewItem checkbox. And the value is set here base on selection type:
https://github.com/aspnet/LibraryManager/blob/285984f683c213c7f50dfcdb6542fac761e82cf3/src/LibraryManager.Vsix/UI/Models/PackageItem.cs#L41

PR #488, opened 2019-06-07
Ensures focus is in the Library Search text box on Install dialog load

alexgav:
Fixes internally tracked accessibility issue 893985.

### DIFF
--- a/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI
 
         private void OnLoaded(object sender, RoutedEventArgs e)
         {
-            FocusManager.SetFocusedElement(LibrarySearchBox, LibrarySearchBox.SearchTextBox);
+            FocusManager.SetFocusedElement(ProviderComboBox, ProviderComboBox);
         }
 
         private void IncludeAllLibraryFilesRb_Checked(object sender, RoutedEventArgs e)
--- a/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI
 
         private void OnLoaded(object sender, RoutedEventArgs e)
         {
-            FocusManager.SetFocusedElement(LibrarySearchBox, LibrarySearchBox);
+            FocusManager.SetFocusedElement(LibrarySearchBox, LibrarySearchBox.SearchTextBox);
         }
 
         private void IncludeAllLibraryFilesRb_Checked(object sender, RoutedEventArgs e)


src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
```diff
@@ -59,7 +59,7 @@ private void CloseDialog(bool res)
 
         private void OnLoaded(object sender, RoutedEventArgs e)
         {
-            FocusManager.SetFocusedElement(LibrarySearchBox, LibrarySearchBox);
+            FocusManager.SetFocusedElement(LibrarySearchBox, LibrarySearchBox.SearchTextBox);
```
bhsubra: I guess I added this last year, but I'm not sure if this is the right thing to do. Someone using the narrator might not know what the provider is(default one). Does it make sense to set focus to provider combo box instead so that when narrator is on it would announce provider and user can tab and move on to type required library?

bhsubra: If you think it's outside the scope of this review, I can look into it once this changes is checked in.

bhsubra:
I guess I added this last year, but I'm not sure if this is the right thing to do. Someone using the narrator might not know what the provider is(default one). Does it make sense to set focus to provider combo box instead so that when narrator is on it would announce provider and user can tab and move on to type required library?

If you think it's outside the scope of this review, I can look into it once this change is checked in.

src/LibraryManager.Vsix/UI/InstallDialog.xaml.cs
alexgav: Hmm, that's an interesting idea, but as a user, I think I would still prefer initial focus in the search box. I might or might not want to change the provider. Most of the time I would probably stick with the same provider. But I would want to do the library search every time. So I think the library id search box is right place. Related to this, we should probably remember last user selected provider so it doesn't reset to default every time. Assuming most people use the same provider for their library, they would know the behavior of us remembering their provider, and simply start typing library name.

jimmylewis: Focus should start at the first element of the dialog.  IMO there's no point in checking in a change that's still wrong.

jimmylewis: As a blind user you'll never know about the provider control.  If you tab through all the controls and then get to Install and Cancel at the bottom, you're not going to cycle through to see if you missed anything.

alexgav: "Focus should start at the first element of the dialog." - is that the official accessibility guideline or your opinion? I don't have a problem with changing it, other than, I think it would hurt the usability. If I were a blind user, I would probably explorer an unknown dialog for the first time - unless "Focus should start on the first element" is a rule that's always followed for accessibility.

jimmylewis: So far it's my opinion, but I'm looking for an explicit guideline.  In the meantime, I have found one that would indicate that Provider should receive focus first:

MAS 2.4.3 quotes [EN 301 549](https://www.etsi.org/deliver/etsi_en/301500_301599/301549/02.01.02_60/en_301549v020102p.pdf) section 11.2.4.3, which states:
> If software can be navigated sequentially and the navigation sequences affect meaning or operation, focusable components receive focus in an order that preserves meaning and operability.

Since the Provider selection affects operation of the search text box, it should receive focus first.

Issue #395, opened 2018-11-10
JSDelivr: First time typing / in Library textbox in UI sends focus back to main VS window


jimmylewis:
### Functional impact
The behavior becomes erratic to an unsuspecting user, as focus is no longer in the textbox that they were typing in.

### Minimal repro steps

1. Open the Add Client Side Library wizard
2. Select jsdelivr as the provider
3. For the provider name, start typing a GitHub library name up to the slash, e.g. `jquery/`

### Expected result
Focus should still be in that text box

### Actual result
Focus goes back to the main VS window (probably Solution Explorer).  However ,funny things continue to happen:
- if you set focus to the VS editor, then to the library name textbox in the wizard, typing happens in the wizard (TYPECHAR).  However, other commands get routed to the editor first, like backspace, undo, or tab.
- The cancel button no longer works.  I tried the Install button and it crashed VS.  Closing the dialog with the `[x]` button worked fine.

PR #493, opened 2019-06-25
Fix for #395 (internally tracked as Bug 936856)

alexgav:
Apprently throwing unhandled exception from event handlers (at least from this one) leave WPF in a bad state. Not sure what's a good generic way of handling it, but this fix addresses the forward slash issue (in that case we were getting CompletionSet struct with Completion list set to null). Also removing a check for null of the struct itself. It's a struct. That method will always return false.

### DIFF
--- a/src/LibraryManager.Vsix/UI/Controls/TextBoxWithSearch.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/Controls/TextBoxWithSearch.xaml.cs
@@ -231,7 +231,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Controls
                 {
                     CompletionSet completionSet = await ViewModel.GetCompletionSetAsync(SearchTextBox.CaretIndex);
 
-                    if (completionSet.Equals(null) || !completionSet.Completions.Any())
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Vsix/UI/Controls/TextBoxWithSearch.xaml.cs
```diff
@@ -231,7 +231,7 @@ private void SearchTextBox_TextChanged(object sender, TextChangedEventArgs e)
                 {
                     CompletionSet completionSet = await ViewModel.GetCompletionSetAsync(SearchTextBox.CaretIndex);
 
-                    if (completionSet.Equals(null) || !completionSet.Completions.Any())
+                    if (completionSet.Completions == null || !completionSet.Completions.Any())
```
jimmylewis: Should we still have a null check on `completionSet`?  Maybe do `if (completionSet?.Completions is null || ...)`?

jimmylewis: never mind, saw that it's a struct, so null check isn't needed/valid.

PR #494, opened 2019-06-26
Fixing npm search

alexgav:
Switching search to use registry database and API since skimdb was retired.

### DIFF
--- a/src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
+++ b/src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
@@ -97,14 +97,7 @@ namespace Microsoft.Web.LibraryManager.Providers.Unpkg
 
         private static async Task<IEnumerable<string>> GetPackageNamesFromSimpleQueryAsync(string searchTerm, CancellationToken cancellationToken)
         {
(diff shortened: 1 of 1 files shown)


src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
```diff
@@ -96,52 +97,98 @@ private static async Task<IEnumerable<string>> GetPackageNamesWithScopeAsync(str
 
         private static async Task<IEnumerable<string>> GetPackageNamesFromSimpleQueryAsync(string searchTerm, CancellationToken cancellationToken)
         {
-            List<string> packageNames = new List<string>();
+            string searchUrlFormat = GetCustomNpmRegistryUrl() ?? NpmPackageSearchUrl;
```
jimmylewis: LibMan queries NPM to get package info that comes from CDNs (unpkg, jsdelivr) - they won't be able to serve up packages from custom feeds.  Since the GetCustomNpmRegistryUrl() method is basically a noop, should we just remove it?  It doesn't feel like it makes sense in this code base.

alexgav: Yeah, I agree, I removed that.

Issue #475, opened 2019-05-01
file permissions incorrect after adding signalR to new webapp on ubuntu?


rakkarage:
### Functional impact
file permissions incorrect after adding signalR to new webapp on ubuntu?

### Minimal repro steps
after adding signalR using 

`libman install @aspnet/signalr -p unpkg -d wwwroot/lib/signalr --files dist/browser/signalr.js --files dist/browser/signalr.min.js`

from

https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr?view=aspnetcore-2.2&tabs=visual-studio

the file permissions are like this

![Untitled](https://user-images.githubusercontent.com/962340/57001292-e3e88680-6b85-11e9-980c-e9e24c35e310.png)

so it works locally but when run on server get error because public cannot see it

![222](https://user-images.githubusercontent.com/962340/57001323-1befc980-6b86-11e9-9e71-d2279151a8ff.png)

`sudo chmod 664 filename`

fixes it 

### Expected result

even a new sln or project file is generated with more permissions then these files? should or could it start with default 664 like most other generated files? since these permissions create an 'error' without a 'good' error message

or maybe i am just missing a setting or something
or maybe it is something with those permission and dotnet publish
thanks

jimmylewis:
This should be fixed in the CLI or Build NuGet packages version 2.0.76 or higher.

PR #495, opened 2019-06-28
Fixes file permission issues on Linux and MacOS

alexgav:
Fixes issue #475. The problem here was that we first write downloaded library to a temp file, and then copy it to the final destination. That by itself is fine, but on Linux and MacOS Path.GetTempFileName() create a temp file with permission 600, which causes issues on publishing (since read permission is missing for everyone other than the owner). Working around it by creating and writing to a slightly modified temp file name that will still be unique, but wasn't originally created as a temp file by mkstemps, so it won't have 600 permissions and instead will have the desired 664 permissions.

### DIFF
--- a/src/LibraryManager.Contracts/FileHelpers.cs
+++ b/src/LibraryManager.Contracts/FileHelpers.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -42,6 +43,18 @@ namespace Microsoft.Web.LibraryManager.Contracts
             }
             else
             {
+                string originalTempFileName = tempFileName;
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    // Temp files created on Linux and MacOS by Path.GetTempFileName() will have 600 permissions.
+                    // We want 664 permissions (read permission for all). So rather than using the orignial temp file,
+                    // we will append suffix to its name, which in practical terms still guarantees a unique temp file but
+                    // will have the default 644 permissions when we write to that file.
+                    // See issue https://github.com/aspnet/LibraryManager/issues/475
+                    tempFileName += ".txt";
+                }
+
                 result = await WriteToFileAsync(tempFileName, sourceStream, cancellationToken).ConfigureAwait(false);
 
                 if (result)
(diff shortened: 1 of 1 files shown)


src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -42,6 +43,18 @@ public static async Task<bool> SafeWriteToFileAsync(string fileName, Stream sour
...
+                string originalTempFileName = tempFileName;
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    // Temp files created on Linux and MacOS by Path.GetTempFileName() will have 600 permissions.
+                    // We want 664 permissions (read permission for all). So rather than using the orignial temp file,
+                    // we will append suffix to its name, which in practical terms still guarantees a unique temp file but
+                    // will have the default 644 permissions when we write to that file.
+                    // See issue https://github.com/aspnet/LibraryManager/issues/475
+                    tempFileName += ".txt";
```
jimmylewis: .temp seems like a more appropriate extension than .txt

src/LibraryManager.Contracts/FileHelpers.cs
```diff
@@ -52,21 +65,31 @@ public static async Task<bool> SafeWriteToFileAsync(string fileName, Stream sour
...
+                if (tempFileName != originalTempFileName)
+                {
+                    SafeDeleteTempFile(originalTempFileName);
                 }
             }
 
             return result;
         }
 
+        private static void SafeDeleteTempFile(string tempFileName)
```
jimmylewis: I don't see why Temp is relevant in this.  Why not just delete the file directly?  It should be apparent from the usage that you're deleting the temp files
```
DeleteFileFromDisk(tempFileName); <-- pretty clear that you're deleting the temp file
```

If the Debug.Fail is the only value this adds: a) it doesn't do anything on non-Windows (where there aren't non-fatal Debug listeners by default), and b) why not move that to `DeleteFileFromDisk`, which already has a try-catch.  This one puts us (I think) 3 levels deep of try-catching around the actual call to `File.Delete()`... I don't think it will actually have anything to catch.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: Sure, sounds good.

src/LibraryManager.Contracts/FileHelpers.cs
alexgav: That's fair, didn't realize we already had DeleteFile that did try/catch. I'll leave the try/catch in DeleteFile though just in case. In my experience, I/O operations **always** throw, earlier or later.

Issue #352, opened 2018-08-14
Install Library dialog does not handle scoped package names (starting with '@') well


kellcomnet2:
### Functional impact
I am unable to add a client side library for fontawesome due to an error.

### Minimal repro steps
Open Add Client-Side Library
Select "unpkg" as the provider
Search for "@fortawesome/fontawesome-free"
Select the package
No files appear in the list and if you choose install it fails with the "@fortawesome/fontawesome-free" library could not be resolved by the unpkg provider

### Expected result
You would be able to add the package

### Actual result
No files appear in the list and if you choose install it fails with the "@fortawesome/fontawesome-free" library could not be resolved by the unpkg provider

### Additional Details
If you manually type a second @ it will give you the version list and you can choose a specific version and it works. It appears to be an issue caused by the @ in the package path/name

nickalbrecht:
This might be related to #329

kellcomnet2:
Yes it does appear to be related, I didn't notice the significance of the @ until after I posted the issue originally. I then amended the details.

biehlermi:
Is there a workaround available for this bug?

Altafard:
I was be able to find the package by full name and version '@fortawesome/fontawesome-free@5.5.0' but seems that it cannot be installed. What's problem?

Altafard:
There is some update - it's all ok for me

![default](https://user-images.githubusercontent.com/28362919/47967251-86f39400-e06c-11e8-84a1-316583adb035.png)

ASKemp:
Not working for me

robinwilson16:
It looks like this is fixed now.
I have had to comment Font Awesome out from my libman.json files for to get it to compile for the last two weeks but today the package resolves correctly and everything was downloaded to the correct folder.

This is what I use:
```
{
  "provider": "unpkg",
  "library": "@fortawesome/fontawesome-free@5.6.3",
  "destination": "wwwroot/lib/font-awesome/"
},
```
Maybe someone else can confirm?

jimmylewis:
I think this is separate from #329 because the repro steps for this issue are when there is only a single @ sign.  I can still repro with the original steps, and it seems like the error returned should better indicate that the version is missing from the library ID.

robinwilson16:
Since my post on 6th Jan this has been an intermittent problem but more recently it has consistently failed every time.
My projects stall on build so I need to stop and comment out the reference to fontawesome in libman.

Errors are that files failed to download:
`
[...]

PR #496, opened 2019-07-08
Allowing for Unpkg to search for scoped packages

bzhu94:
Fixes [Issue #352](https://github.com/aspnet/LibraryManager/issues/352)


jimmylewis:
I'm going to assume JSDelivr has the same problem.  Can you double check?

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -175,7 +175,7 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
             try
             {
                 // library name completion
-                if (caretPosition < name.Length + 1)
+                if (caretPosition < name.Length + 1 || string.IsNullOrEmpty(name) && libraryNameStart.StartsWith("@", StringComparison.Ordinal))
```
alexgav: I'd Trim() libraryNameStart at the beginning of this method to make this code a bit more resilient to possible leading or trailing spaces.

jimmylewis: I don't think this is the right fix.  We're getting the wrong data back from the LibraryIdToNameAndVersionConverter (which in this case is using VersionedLibraryNamingScheme) when a scoped package doesn't have a 2nd @ to delineate the version number.

bzhu94:
@jimmylewis JSDeliver completion for scoped packages seems to be fine

test/LibraryManager.Test/VersionedLibraryNamingSchemeTest.cs
```diff
@@ -11,7 +11,10 @@ public class VersionedLibraryNamingSchemeTest
     {
         [DataTestMethod]
         [DataRow("jquery@3.3.1", "jquery", "3.3.1")]
-        [DataRow("@angular/cli@1.0.0", "@angular/cli", "1.0.0")]
+        [DataRow("@angular/cli@1.0.0", "cli", "1.0.0")]
```
jimmylewis: These are wholly wrong.  The expected name is "@angular/cli".

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
```diff
@@ -1,19 +1,20 @@
...
     {
         private const char Separator = '@';
 
         /// <summary>
         /// Splits libraryId into name and version using '@' as the split char.
         /// Only the last appearance of '@' is used to split the libraryId.
-        /// Note: If libraryId starts with '@' and has no other occurences of '@',
-        /// the entire libraryId is considered to be the name and version is considered to be
-        /// empty.
+        /// Note: If libraryId starts with '@', a (name + version) substring will be split after the first '/',
```
jimmylewis: The / is part of the name, but not a delimiter.  For example, there's a big difference `@angular/cli` and `@react/cli`, you can't trim the name of both of these to `cli`.

test/LibraryManager.Test/VersionedLibraryNamingSchemeTest.cs
jimmylewis: Resolving this in favor of other comment closer to the code change.  Better place to hash it out.

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
jimmylewis: (I think this largely invalidates the commit, so I'm going to hold off looking at the other changes in this commit)

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -18,7 +18,10 @@ public class LibraryIdViewModelTests
         private readonly ILibraryCatalog _testCatalog = new LibraryCatalog()
             .AddLibrary(new Library { Name = "test", Version = "1.0" })
             .AddLibrary(new Library { Name = "test", Version = "1.2" })
-            .AddLibrary(new Library { Name = "test", Version = "2.0" });
+            .AddLibrary(new Library { Name = "test", Version = "2.0" })
+            .AddLibrary(new Library { Name = "@types/node", Version = "1.0" })
```
jimmylewis: Please try to use obvious garbage data in unit tests.  It makes it more clear that this isn't hitting a live endpoint, and that the test output is not affected by external changes.

src/LibraryManager.Vsix/UI/Models/LibraryIdViewModel.cs
```diff
@@ -23,7 +23,15 @@ public LibraryIdViewModel(ISearchService packageSearchService, string initialSea
         protected override Task<CompletionSet> FilterCompletions(CompletionSet input)
         {
             CompletionSet result = input;
-            int atIndex = SearchText.IndexOf('@');
+            int atIndex = SearchText.LastIndexOf('@');
+
+            //For scoped packages
```
jimmylewis: Finish the explanation:
`// For scoped packages, the version segment starts after an @ sign which follows a /`

src/LibraryManager.Vsix/UI/Models/LibraryIdViewModel.cs
```diff
@@ -23,7 +23,15 @@ public LibraryIdViewModel(ISearchService packageSearchService, string initialSea
...
+            if (SearchText.StartsWith("@", StringComparison.Ordinal))
+            {
+                int indexOfFirstSlash = SearchText.IndexOf('/');
+                atIndex = indexOfFirstSlash > 0 && SearchText.LastIndexOf('@') > indexOfFirstSlash ?
```
jimmylewis: You already calculated `SearchText.LastIndexOf('@')` above, but it's repeated twice again here.

src/LibraryManager.Vsix/UI/Models/LibraryIdViewModel.cs
```diff
@@ -23,7 +23,15 @@ public LibraryIdViewModel(ISearchService packageSearchService, string initialSea
         protected override Task<CompletionSet> FilterCompletions(CompletionSet input)
         {
             CompletionSet result = input;
-            int atIndex = SearchText.IndexOf('@');
+            int atIndex = SearchText.LastIndexOf('@');
```
jimmylewis: If we wanted to simplify this a bit by making *a large and scary assumption*, we could just assume that an `@` at the beginning is for a scoped package, and there should only ever be a single `@` after that.  If we go on this assumption, we could just do `int atIndex = SearchText.LastIndexOf('@', 1);` and then skip the if block.

This is a technically correct (if inflexible - but it is simple!) option for all valid scenarios today.  There are currently no  scenarios (AFAIK?) that allow @ anywhere in the name except at the beginning for a scoped package.  If you have multiple after that, there isn't a strictly defined behavior.

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -41,7 +44,7 @@ public async Task FilterCompletions_SearchTextDoesNotContainAtSign_ReturnAllMatc
         {
             var testObj = new LibraryIdViewModel(GetTestSearchService(), "test");
 
-            CompletionSet result = await testObj.GetCompletionSetAsync(caretIndex: 0);
+            CompletionSet result = await testObj.GetCompletionSetAsync(caretIndex: 4);
```
jimmylewis: Why do these need to change?  When I wrote the tests originally, the caretIndex was actually kind of meaningless on this code path, which is why they were set to 0.

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -51,18 +54,49 @@ public async Task FilterCompletions_SearchTextEndsWithAt_ReturnAllMatchingComple
...
+        }
+
+        [TestMethod]
+        public async Task FilterCompletions_ScopedPackageEndsWithAfterAt_ReturnAllMatchingCompletions()
+        {
+            var testObj = new LibraryIdViewModel(GetTestSearchService(), "@types/node@2");
+
+            CompletionSet result = await testObj.GetCompletionSetAsync(caretIndex: 13);
+
+            Assert.AreEqual(1, result.Completions.Count());
```
jimmylewis: Does this filter differently?  The unscoped case expects 2 results when the version string is "2".

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
jimmylewis: Ah, this was addressed later, but the comment wasn't restored.

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -170,7 +170,8 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
 
             // Typing '@' after the library name should have version completion.
             int at = name.LastIndexOf('@');
-            name = at > -1 ? name.Substring(0, at) : name;
+            bool isScopedName = libraryNameStart.StartsWith("@", StringComparison.Ordinal);
```
jimmylewis: It feels like if we've already converted the library Id to `(name, version)` we shouldn't have to check for scopes again.  I think the original code here was handling a weird case from VersionedLibraryNamingScheme, in that for a partial name e.g. `jquery@` it would return `("jquery@", "")`.  

Note the @ is included in the name *only when there is no version segment*, whereas in a case where there is any version portion, it skips the at: jquery@1 would return `("jquery", "1")` and the `@` is dropped from both (this also grants a fun check - if `name.Length + version.Length == libraryNameStart.Length` then VLNS didn't skip over the @ sign and considers it a partial match.

I think there's a better ways to address this than re-dealing with scopes here.
1.  Change VLNS to not leave the @ appended in this "partial match" case.  I don't know why it does this, but it's just causing more work elsewhere (namely here and in JsDelivrCatalog).  I don't know if this behavior is relied upon elsewhere, we would need to check other uses of VLNS to see if there's any other paths where this might change a decision branch.
2. Do a smarter check instead of IndexOf tricks.  We're just looking to eliminate @ from the end, so:
```
string nonPartialName = name.TrimEnd("@");
if (caretPosition < name.Length + 1) // or caretPosition <= name.Length
```
should do the trick.

The JSDelivrCatalog has the same idea, but so much more succinct.  This one's a little harder to read, and I don't think it's correct when the caret position is not at the end of the `libraryNameStart`, but it probably works in the common case (though we should probably make them the same).

    if (caretPosition < name.Length + 1 && name[name.Length - 1] != '@')

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -180,14 +180,14 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
...
+                    IEnumerable<Tuple<string, string>> packageNamesAndCurrentVersions = await NpmPackageSearch.GetPackageNamesAndCurrentVersionsAsync(libraryNameStart, CancellationToken.None);
 
-                    foreach (string packageName in packageNames)
+                    foreach (Tuple<string, string> nameAndCurrentVersion in packageNamesAndCurrentVersions)
                     {
                         CompletionItem completionItem = new CompletionItem
                         {
-                            DisplayText = packageName,
-                            InsertionText = packageName,
+                            DisplayText = nameAndCurrentVersion.Item1,
```
jimmylewis: Item1/Item2 syntax suffers from a complete loss of semantic information.  Instead of using them, use tuple deconstruction so you can have names or create a type with members.

src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
```diff
@@ -7,6 +7,7 @@
 using System.Threading.Tasks;
 using System.Web;
 using Newtonsoft.Json.Linq;
+using NameAndCurrentVersion = System.Tuple<string, string>;
```
jimmylewis: A struct is better than this any day of the week.

src/LibraryManager/Providers/Unpkg/NpmPackageSearch.cs
```diff
@@ -179,9 +181,11 @@ private static async Task<IEnumerable<string>> GetPackageNamesFromSimpleQueryAsy
                             if (packageEntry != null)
                             {
                                 string currentPackageName = packageEntry["name"].ToString();
-                                if (!String.IsNullOrWhiteSpace(currentPackageName))
+                                string currentPackageVersion = packageEntry["version"].ToString();
```
jimmylewis: This seems really close to what we do in CreatePackageInfoAsync.  Maybe we can consolidate with that and return NpmPackageInfo instead of this tuple hack?

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -180,14 +180,14 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
                     libraryNameStart.StartsWith("@", StringComparison.Ordinal) &&
                     at < 0 || caretPosition < at) 
                 {
-                    IEnumerable<string> packageNames = await NpmPackageSearch.GetPackageNamesAsync(libraryNameStart, CancellationToken.None);
+                    IEnumerable<Tuple<string, string>> packageNamesAndCurrentVersions = await NpmPackageSearch.GetPackageNamesAndCurrentVersionsAsync(libraryNameStart, CancellationToken.None);
```
jimmylewis: Instead of changing NpmPackageSearch so drastically, can we get the desired result by using SearchAsync() for package name completion?  That's how the CDNJS catalog does it, and that gives back an ILibraryGroup, so you can use group.Versions.Max() to get the latest version.

bzhu94:
Going to break this into multiple PR's to address:

1. Fixing Issue 352 specifically
2. Fixing version completion for Unpkg / JsDelivr
3. Fixing up the error shown when package version is missing when users hit "install"

PR #499, opened 2019-07-16
Fixing Issue 352

bzhu94:


test/LibraryManager.Test/VersionedLibraryNamingSchemeTest.cs
```diff
@@ -12,10 +12,15 @@ public class VersionedLibraryNamingSchemeTest
         [DataTestMethod]
         [DataRow("jquery@3.3.1", "jquery", "3.3.1")]
         [DataRow("@angular/cli@1.0.0", "@angular/cli", "1.0.0")]
+        [DataRow("@angular/cli@", "@angular/cli", "")]
+        [DataRow("@angular/cli", "@angular/cli", "")]
```
jimmylewis: `@angular/cli`, `@angular/`, and `@angular` are all redundant with other cases already specified in this list.  Arguably, so is `@angular/cli@`.

test/LibraryManager.Test/Providers/Unpkg/UnpkgCatalogTest.cs
```diff
@@ -113,6 +113,100 @@ public async Task GetLibraryCompletionSetAsync_Names()
...
+        {
+            CompletionSet result = await _catalog.GetLibraryCompletionSetAsync("@types/node@1.0.2", 8);
+
+            Assert.AreEqual(0, result.Start);
+            Assert.AreEqual(17, result.Length);
+            Assert.AreEqual(1, result.Completions.Count());
+            Assert.IsTrue(result.Completions.First().InsertionText.StartsWith("@types/node"));
+        }
+
+        public async Task GetLibraryCompletionSetAsync_LibraryNameWithLeadingAndTrailingWhitespace()
```
jimmylewis: This behavior is inherent to the search API being used, not to the logic of the catalog class.  What you're really testing is not your code.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -155,31 +156,33 @@ private void GetFiles(JObject fileObject, List<string> result)
 
         public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryNameStart, int caretPosition)
         {
-            CompletionSet completionSet = new CompletionSet
+            libraryNameStart = libraryNameStart == null ? string.Empty : libraryNameStart;
```
jimmylewis: `libraryNameStart = libraryNameStart ?? string.Empty;` is more concise.

jimmylewis: Although, it could also be avoided altogether with 

```
var completionSet = new CompletionSet { ... };
if (string.IsNullOrEmpty(libraryNameStart)
{
    // no point in doing the rest of the work, we know it's going to be an empty completion set anyway
    return completionSet; 
}
```

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
```diff
@@ -26,15 +28,16 @@ class VersionedLibraryNamingScheme : ILibraryNamingScheme
                 return (name, version);
             }
```
jimmylewis: add this to simplify handing a single-character libraryId (simplifies below):
```
else if(libraryId.Length == 1)
{
    return (libraryId, version);
}
```

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
```diff
@@ -26,15 +28,16 @@ class VersionedLibraryNamingScheme : ILibraryNamingScheme
...
 
-            int indexOfAt = libraryId.LastIndexOf(Separator);
+            int indexOfLastAt = libraryId.LastIndexOf(Separator);
```
jimmylewis: with the else-if suggested above, you can make this `libraryId.Substring(1).LastIndexOf(Separator) + 1;`.  This removes the need to further calculate `indexOfAtToSplitNameAndVersion`.  (`indexOfLastAt` is now 0 for a non-match, but you're already checking for > 0 below).

Actually, I think it can even further simplify below, back to a nice and easy:

```
if (indexOfAt > 0)
 {
    name = libraryId.Substring(0, indexOfAt);
    version = libraryId.SubString(indexOfAt + 1);
}
else
{
    name = libraryId;
}
```

I don't think the `libraryId.TrimEnd().Length - 1` check is relevant for a non-whitespace separator, and the else block replaces the `name = indexOfLastAt > 0 ? libraryId.TrimEnd('@') : libraryId;` line.  The interesting test case for this one is "@@@", which should return ("@@", "") I believe - trim the final @ as the separator, but TrimEnd('@') would remove all of them.

bzhu94: Resolved in commit 2

src/LibraryManager/LibraryNaming/VersionedLibraryNamingScheme.cs
bzhu94: Fixed in commit 2

test/LibraryManager.Test/VersionedLibraryNamingSchemeTest.cs
bzhu94: Fixed in commit 2. I think I'll leave the `@angular/cli@` as it proves that this case has been fixed, even though it is technically redundant

jimmylewis: Sure, that sounds good.

PR #504, opened 2019-07-30
Fixing version completion for scoped packages

bzhu94:
This fixes LibraryIdViewModel.FilterCompletions not returning any completions for scoped packages.

### DIFF
--- a/test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
+++ b/test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
@@ -15,93 +15,106 @@ namespace Microsoft.Web.LibraryManager.Vsix.Test.UI.Models
     [TestClass]
     public class LibraryIdViewModelTests
     {
-        private readonly ILibraryCatalog _testCatalog = new LibraryCatalog()
-            .AddLibrary(new Library { Name = "test", Version = "1.0" })
-            .AddLibrary(new Library { Name = "test", Version = "1.2" })
-            .AddLibrary(new Library { Name = "test", Version = "2.0" })
-            .AddLibrary(new Library { Name = "@types/node", Version = "1.0" })
-            .AddLibrary(new Library { Name = "@types/node", Version = "1.2" })
-            .AddLibrary(new Library { Name = "@types/node", Version = "2.0" })
-
-            //Add garbage data 
-            .AddLibrary(new Library { Name = "@Bad@Package@", Version = "abc/2.junk" })
-            .AddLibrary(new Library { Name = "", Version = "2.0" })
-            .AddLibrary(new Library { Name = "Blah", Version = "" })
-            .AddLibrary(new Library { Name = "", Version = "" });
-
-
-        private IProvider GetTestProvider()
+        private ISearchService GetTestSearchService(ILibraryCatalog libraryCatalog)
         {
-            IProvider testProvider = new Provider(new HostInteraction())
-            {
-                Catalog = _testCatalog
-            };
-
-            return testProvider;
+            IProvider testProvider = new Provider(new HostInteraction()) { Catalog = libraryCatalog };
+            return new ProviderCatalogSearchService(() => testProvider);
         }
 
-        private ISearchService GetTestSearchService()
+        private ILibraryCatalog CreateLibraryCatalogWithUnscopedLibrary()
         {
-            IProvider provider = GetTestProvider();
-            return new ProviderCatalogSearchService(() => provider);
+            return new LibraryCatalog()
+                .AddLibrary(new Library { Name = "test", Version = "1.0" })
+                .AddLibrary(new Library { Name = "test", Version = "1.2" })
+                .AddLibrary(new Library { Name = "test", Version = "2.0" });
         }
 
         [TestMethod]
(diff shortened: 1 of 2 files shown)


test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -41,7 +51,7 @@ public async Task FilterCompletions_SearchTextDoesNotContainAtSign_ReturnAllMatc
         {
             var testObj = new LibraryIdViewModel(GetTestSearchService(), "test");
 
-            CompletionSet result = await testObj.GetCompletionSetAsync(caretIndex: 0);
```
bzhu94: FYI, changing the caretIndex for the existing tests here is necessary.

GetCompletionSetAsync calls LibraryCatalog.GetLibraryCompletionSetAsync, which returns libraries which match the substring left of the caretPosition. For the tests which I've changed the caretIndex for, caretIndex = 0, which means that the substring is an empty string and therefore will return all library results (which coincidentally worked since all the added libraries were "test")

So I've changed the caretIndex in order to properly enable the filtering on these tests.

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -18,7 +18,17 @@ public class LibraryIdViewModelTests
         private readonly ILibraryCatalog _testCatalog = new LibraryCatalog()
             .AddLibrary(new Library { Name = "test", Version = "1.0" })
             .AddLibrary(new Library { Name = "test", Version = "1.2" })
-            .AddLibrary(new Library { Name = "test", Version = "2.0" });
+            .AddLibrary(new Library { Name = "test", Version = "2.0" })
+            .AddLibrary(new Library { Name = "@types/node", Version = "1.0" })
+            .AddLibrary(new Library { Name = "@types/node", Version = "1.2" })
+            .AddLibrary(new Library { Name = "@types/node", Version = "2.0" })
+
```
jimmylewis: nit: don't have the empty line.  It breaks the reader from the fluent syntax.

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -18,7 +18,17 @@ public class LibraryIdViewModelTests
         private readonly ILibraryCatalog _testCatalog = new LibraryCatalog()
```
jimmylewis: Unless the additional data being added is relevant to all of the tests, I don't like having it all in one place (the original dataset was shared because it was broadly applicable).

The reason for this is that a unit test should contain the minimal set of data to express the scenario.  This also makes it easier to tell during debugging which part of the test data is relevant to the issue (ideally, all of it); otherwise, you might have to spend time debugging if it's because of the "garbage data", when that data might not be relevant to the failure or scenario being tested at all.

What if we changed these tests to each create the catalog with data they need, then pass that to a helper factory method to get the LibraryIdViewModel to test?  Something along the lines of:
```
var stubCatalog = new LibraryCatalog()
                      .AddLibrary(new Library { Name = "test", Version = "1.0" })
                      .AddLibrary(new Library { Name = "@scope/test", Version = "1.2" } );
LibraryIdViewModel testObj = CreateLibraryIdViewModel(stubCatalog);
```

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
```diff
@@ -18,7 +18,17 @@ public class LibraryIdViewModelTests
...
             .AddLibrary(new Library { Name = "test", Version = "1.0" })
             .AddLibrary(new Library { Name = "test", Version = "1.2" })
-            .AddLibrary(new Library { Name = "test", Version = "2.0" });
+            .AddLibrary(new Library { Name = "test", Version = "2.0" })
+            .AddLibrary(new Library { Name = "@types/node", Version = "1.0" })
```
jimmylewis: Names like "@types/node" are real package name, so it can be confusing as to whether something is going to hit a live endpoint or not to get data about this.  There's no benefit to using a real package name over using a fake name (unless you're doing an integration test).

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
jimmylewis: Thanks for explaining

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
bzhu94: Resolved in commit 2

test/Microsoft.Web.LibraryManager.Vsix.Test/UI/Models/LibraryIdViewModelTests.cs
bzhu94: Resolved in commit 2

jimmylewis:
Please add a description in the PR, so that it has something in the commit when merged.

PR #507, opened 2019-08-01
Add latest version to name completion insertion text for JsDelivr + Unpkg 

bzhu94:
This change will auto-fill package name + latest version in the search text box when users select a package in the search text box dropdown.


src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -185,14 +188,16 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
...
 
                     foreach (string packageName in packageNames)
                     {
+                        NpmPackageInfo packageInfo = await _packageInfoCache.GetPackageInfoAsync(packageName, CancellationToken.None);
+
                         var completionItem = new CompletionItem
                         {
                             DisplayText = packageName,
-                            InsertionText = packageName,
+                            InsertionText = LibraryIdToNameAndVersionConverter.Instance.GetLibraryId(packageName, packageInfo.LatestVersion, _providerId)
```
jimmylewis: use `_libraryNamingScheme.GetLibraryId`

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -37,7 +37,9 @@ public UnpkgProvider(IHostInteraction hostInteraction)
 
         public ILibraryCatalog GetCatalog()
         {
-            return _catalog ?? (_catalog = new UnpkgCatalog(Id, LibraryNamingScheme, HostInteraction.Logger, WebRequestHandler.Instance));
+            var packageSearch = new NpmPackageSearch();
+            var infoCache = new NpmPackageInfoCache(packageSearch);
```
jimmylewis: Hm, so this cache will be separate from the one for the jsdelivr provider?

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -185,14 +189,16 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
...
 
                     foreach (string packageName in packageNames)
                     {
+                        NpmPackageInfo packageInfo = await _packageInfoCache.GetPackageInfoAsync(packageName, CancellationToken.None);
+
                         var completionItem = new CompletionItem
                         {
                             DisplayText = packageName,
-                            InsertionText = packageName
+                            InsertionText = LibraryIdToNameAndVersionConverter.Instance.GetLibraryId(packageName, packageInfo.LatestVersion, _providerId)
```
jimmylewis: Use `_libraryNamingScheme` here too

test/LibraryManager.Test/Providers/JsDelivr/JsDelivrCatalogTest.cs
```diff
@@ -291,6 +299,46 @@ public async Task GetLatestVersion_PreRelease()
...
+        public async Task GetLibraryCompletionSetAsync_ReturnsCompletionWithLatestVersion()
+        {
+            //Arrange
+            var dependencies = new Dependencies(null);
+            var provider = new Provider(null) { Id = JsDelivrProvider.IdText, SupportsLibraryVersions = true };
+
+            dependencies.AllProviders = new List<IProvider>() { provider };
+
+            //Reinitialize is necessary because any other tests could have already initialized before
+            LibraryIdToNameAndVersionConverter.Instance.Reinitialize(dependencies);
```
jimmylewis: This becomes unnecessary by using the instance field above.  (That was part of #498, to avoid this level of state-dependent setup.)

test/LibraryManager.Test/Providers/Unpkg/UnpkgCatalogTest.cs
```diff
@@ -265,6 +271,46 @@ public async Task GetLibraryCompletionSetAsync_ScopedPackageNameisSingleAt_Retur
...
+        public async Task GetLibraryCompletionSetAsync_ReturnsCompletionWithLatestVersion()
+        {
+            //Arrange
+            var dependencies = new Dependencies(null);
+            var provider = new Provider(null) { Id = UnpkgProvider.IdText, SupportsLibraryVersions = true };
+
+            dependencies.AllProviders = new List<IProvider>() { provider };
+
+            //Reinitialize is to necessary because any other tests could have already initialized before
+            LibraryIdToNameAndVersionConverter.Instance.Reinitialize(dependencies);
```
jimmylewis: as above, this should become unnecessary

src/LibraryManager/Providers/Unpkg/INpmPackageInfoCache.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Web.LibraryManager.Providers.Unpkg
+{
+    internal interface INpmPackageInfoCache
```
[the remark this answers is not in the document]
jimmylewis: It also would make the type reusable by additional providers not in this library/not already included in the IVT list.  That reusability might be nice if someone tries to implement their own provider based on the NPM catalog.

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
jodavis: How often is GetCatalog() called? If it's more than once, the cache will get recreated a lot.

On the other hand, how long do we want to maintain a cache of data that might never be used again?

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -185,14 +189,16 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
...
                     }
 
-                    IEnumerable<string> packageNames = await NpmPackageSearch.GetPackageNamesAsync(libraryNameStart, CancellationToken.None);
+                    IEnumerable<string> packageNames = await _packageSearch.GetPackageNamesAsync(libraryNameStart, CancellationToken.None);
 
                     foreach (string packageName in packageNames)
                     {
+                        NpmPackageInfo packageInfo = await _packageInfoCache.GetPackageInfoAsync(packageName, CancellationToken.None);
+
                         var completionItem = new CompletionItem
```
jodavis: This used to do one request for package names, now it's going to do lots of requests for the info for every package. Is there a way InsertionText can be computed on-demand? The majority of InsertionText values that are created are going to be discarded, so ideally we would save those requests until they're actually necessary.

(This applies to the lines above. Apparently I can't aim GitHub comments.)

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
jodavis: Now that I'm paying attention... packageSearch and infoCache are created every time this method is called, but _catalog is cached in the line below. The objects above are created for no reason if _catalog != null.

(Same applies to jsDeliver)

src/LibraryManager/Providers/Unpkg/INpmPackageInfoCache.cs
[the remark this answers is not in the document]
bzhu94: Resolved in commit 5

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
bzhu94: Resolved in commit 4

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
bzhu94: Resolved in commit 4

test/LibraryManager.Test/Providers/JsDelivr/JsDelivrCatalogTest.cs
bzhu94: Resolved in commit 4

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
jimmylewis: I think it's possible now that we're using the NPM registry for the name queries.  The response body we get back includes the latest version; the search API we used prior to #494 didn't include any version info.  We'll have to change where we parse the response info to do this, but it should now be possible without multiple requests.

test/LibraryManager.Test/Providers/JsDelivr/JsDelivrCatalogTest.cs
```diff
@@ -17,10 +20,15 @@ public class JsDelivrCatalogTest
     {
         private static JsDelivrCatalog SetupCatalog(IWebRequestHandler webRequestHandler = null)
         {
+            var packageSearch = new NpmPackageSearch();
```
jimmylewis: These should probably follow the same pattern as the `webRequestHander` - optionally pass in a test fake, otherwise create a default.  That way they're easily used with an isolation framework (to generate a fake) and still able to reuse this method so that individual tests don't need to know/fake all the rest of the dependencies as well.

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
bzhu94: Resolved with commit 8

test/LibraryManager.Test/Providers/JsDelivr/JsDelivrCatalogTest.cs
bzhu94: Fixed with commit 9

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
bzhu94: Fixed with commit 7. NpmPackageSearch methods will now return NpmPackageInfo rather than just a package name, which I take advantage of here

src/LibraryManager/Providers/Unpkg/NpmPackageInfoFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+                packageInfo = await GetPackageInfoForUnscopedPackageAsync(packageName, cancellationToken).ConfigureAwait(false);
+            }
+
+            return packageInfo;
+        }
+
+        private static async Task<NpmPackageInfo> GetPackageInfoForScopedPackageAsync(string packageName, CancellationToken cancellationToken)
+        {
+            Debug.Assert(packageName.StartsWith("@", StringComparison.Ordinal));
+            string searchName = "@" + HttpUtility.UrlEncode(packageName.Substring(1));
```
jimmylewis: Might be worth adding a comment here that we're trying to avoid encoding the @ sign.  At a glance it may seem weird to do the concatenation with `packageName.Substring(1)` until one realizes the actual intent.

PR #509, opened 2019-08-06
Having LibMan use MEF instead of reflection for .NET Standard 2.0 to pull in IProviderFactory

bzhu94:
Currently LibMan has 2 methods to import IProviderFactory: .NET Standard 2.0 uses reflection while .NET 472 uses MEF. This change have both frameworks use MEF to do so.

### DIFF
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
@@ -14,6 +14,7 @@
     <PackageReference Include="Microsoft.Build.Framework" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" />
     <PackageReference Include="System.Runtime.Loader" />
+    <PackageReference Include="Microsoft.VisualStudio.Composition" Version="16.1.8" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
@@ -13,7 +13,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Framework" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" />
-    <PackageReference Include="System.Runtime.Loader" Condition="'$(TargetFramework)' == 'netstandard2.0'"/>
+    <PackageReference Include="System.Runtime.Loader" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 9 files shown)


src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
```diff
@@ -13,7 +13,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Framework" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" />
-    <PackageReference Include="System.Runtime.Loader" Condition="'$(TargetFramework)' == 'netstandard2.0'"/>
```
bzhu94: This is to resolve the new use of `AssemblyLoadContext` in `src/LibraryManager.Build/Contracts/Dependencies.cs`

src/LibraryManager/Microsoft.Web.LibraryManager.csproj
```diff
@@ -11,10 +11,10 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="System.ComponentModel.Composition" Version="4.5.0" />
```
bzhu94: @alexgav I couldn't find any good use case for the Microsoft.VisualStudio.Composition Nuget package so I just stuck with using this.

jimmylewis:
Just to play devil's advocate a bit:

Using MEF across all 3 systems will simplify things a bit, but do we even need MEF at all right now?  It makes sense to use it once we figure out how LibMan will be extensible, but right now we have 4 providers - always the same 4 providers.  We could just initialize those providers directly, e.g.

```
var providerFactories = new IProviderFactory [] {
    new FileSystemProviderFactory(),
    new CdnjsProviderFactory(),
    new UnpkgProviderFactory(),
    new JsDelivrProviderFactory(),
};

foreach(IProviderFactory factory in providerFactories)
{ /* this would be the same loop */ }
```

and skip all the MEF initialization for now.  Once we have an extensibility story, we can revisit and use an extensibility framework.

bzhu94:
@jimmylewis currently `IProviderFactory` is a public type. If we go with initializing the providers directly, would ignoring implementations from 3rd party devs be an issue?

jimmylewis:
I'm not aware of any other providers, so it's a moot point today.  We haven't finalized a way to add new providers, so anything that works now is coincidental.  This was discussed a bit in #67, but to recap, we need to spec out how to extend across:
* VS (e.g. is a VSIX required?)
* MSBuild (e.g. is there a way to hook in via a .props or .targets file?)
* CLI (e.g. is there a way for a separate package to install itself to the libman CLI tool folder?)
* And ultimately, how do we bridge across these so that each custom provider doesn't have to ship 3 different ways.

bzhu94:
From what you describe, it seems reasonable to construct the 4 providers directly, so I'll close this PR.

PR #510, opened 2019-08-08
Adding more diagnostics for malformed manifest error

### LINKED ISSUE #111: LIB004 The manifest file contains syntax errors
### Functional impact
LibMan with VS 15.8.0 Preview 2.0 has stopped working for me, parsing issue with filesystem providers' "files" array. 

### Minimal repro steps

1. Create New VS ASP.Net Web Application, SPA starter.
2. Enable Client-side Library Management
3. Add something. eg:

{
  "version": "1.0",
  "defaultProvider": "filesystem",
  "libraries": [
    {
      "library": "node_modules/jquery/dist/",
      "destination": "Scripts/lib/jquery",
      "files": [ "jquery.min.js", "jquery.min.map" ]
    }
   ]
}

Saving this results in:

LIB004 The manifest file contains syntax errors in the Error List pad

### Expected result

This used to work last week ??

### Actual result

LIB004 The manifest file contains syntax errors in the Error List pad

### Further technical details
??

alexgav:
This partially addresses #111 . I'm still unable to repro this, but I'm adding more diagnostics so we should be able to investigate this issue better.

### DIFF
--- a/src/LibraryManager/Manifest.cs
+++ b/src/LibraryManager/Manifest.cs
@@ -3,15 +3,17 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.Web.LibraryManager.Json;
 using Microsoft.Web.LibraryManager.LibraryNaming;
+
 using Newtonsoft.Json;
 
 namespace Microsoft.Web.LibraryManager
@@ -72,7 +74,7 @@ namespace Microsoft.Web.LibraryManager
         /// <param name="dependencies">The host provided dependencies.</param>
         /// <param name="cancellationToken">A token that allows for cancellation of the operation.</param>
         /// <returns>An instance of the <see cref="Manifest"/> class.</returns>
-        public static async Task<Manifest> FromFileAsync(string fileName, IDependencies dependencies, CancellationToken cancellationToken)
+        public static async Task<(Manifest manifest, string diagnostics)> FromFileAsync(string fileName, IDependencies dependencies, CancellationToken cancellationToken)
         {
             if (File.Exists(fileName))
             {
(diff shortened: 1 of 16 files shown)


src/LibraryManager/Manifest.cs
```diff
@@ -102,11 +106,21 @@ internal static Manifest FromJson(string json, IDependencies dependencies)
...
             }
-            catch (Exception)
+            catch (Exception e)
             {
-                return null;
+                string fullDiagnostics = e.ToString() + Environment.NewLine +
+                    "---" + Environment.NewLine +
+                    Thread.CurrentThread.Name + Environment.NewLine +
+                    "---" + Environment.NewLine +
+                    Environment.StackTrace + Environment.NewLine +
```
jimmylewis: Is this stack trace useful?  Doesn't `e.ToString()` contain a stack trace of where the exception was thrown (which would include this frame).

src/LibraryManager/Manifest.cs
```diff
@@ -83,14 +85,16 @@ public static async Task<Manifest> FromFileAsync(string fileName, IDependencies
             return FromJson("{}", dependencies);
         }
 
+
```
jimmylewis: nit: unnecessary blank line

src/LibraryManager/Manifest.cs
```diff
@@ -83,14 +85,16 @@ public static async Task<Manifest> FromFileAsync(string fileName, IDependencies
...
         /// <summary>
         /// Creates an instance of the <see cref="Manifest"/> class based on
         /// the provided JSON string.
         /// </summary>
         /// <param name="json">A string of JSON in the correct format.</param>
         /// <param name="dependencies">The host provided dependencies.</param>
         /// <returns></returns>
-        internal static Manifest FromJson(string json, IDependencies dependencies)
+        [SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Capturing exception detail")]
+        internal static (Manifest manifest, string diagnostics) FromJson(string json, IDependencies dependencies)
```
jimmylewis: How useful is it to pass up the diagnostics string?  We're logging details in the catch block here, and then almost immediately logging it again - well, `e.Message`, so not as much details but it's still redundant with the `e.ToString()` - and to the same loggers.  It'll be a verbose dump followed by a less verbose report of the same failure.

Can we keep the API simple (just return Manifest) and get enough info from what we're logging here?

PR #522, opened 2019-09-24
Fix code warnings from analyzers

jimmylewis:
This addresses the remaining code analysis failures.  There are still a few MSBuild warnings left, but this significantly improves our code cleanliness in the editor/IDE.

### DIFF
--- a/src/LibraryManager/SemanticVersion/SemanticVersion.cs
+++ b/src/LibraryManager/SemanticVersion/SemanticVersion.cs
@@ -136,6 +136,13 @@ namespace Microsoft.Web.LibraryManager
             return ver;
         }
 
+        /// <summary>
+        /// Compares this object to the other and returns a result indicating sort order.
+        /// </summary>
+        /// <remarks>
+        /// This comparison does take build metadata into account for the comparison.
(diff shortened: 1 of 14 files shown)


src/LibraryManager/SemanticVersion/SemanticVersion.cs
```diff
@@ -186,24 +193,92 @@ public int CompareTo(SemanticVersion other)
...
+        {
+            return !(left == right);
+        }
+
+        /// <summary>
+        /// Less-than operator
+        /// </summary>
+        public static bool operator <(SemanticVersion left, SemanticVersion right)
+        {
+            return ReferenceEquals(left, null) ? !ReferenceEquals(right, null) : left.CompareTo(right) < 0;
```
tlmii: Out of curiosity, why `ReferenceEquals(left/right, null)` and not `left/right is null`? Same on the other lt/gt operators.

jimmylewis: That was generated from the suggested code fix, but I can change it to use `is null` instead.

Issue #523, opened 2019-09-26
Publishing .Net Core 2.2 app broken after upgrading to VS2019 16.3.1


euang:
After upgrading to VS2019 16.3.1, I can no longer publish my .Net Core 2.2 app.
I'm using 2.0.76

It throws the following error:
```
Restore operation started...
C:\Users\euan\.nuget\packages\microsoft.web.librarymanager.build\2.0.76\build\Microsoft.Web.LibraryManager.Build.targets(35,9): Error MSB4018: The "Microsoft.Web.LibraryManager.Build.RestoreTask" task failed unexpectedly.
System.MissingMethodException: No parameterless constructor defined for this object.
   at System.RuntimeTypeHandle.CreateInstance(RuntimeType type, Boolean publicOnly, Boolean noCheck, Boolean& canBeCached, RuntimeMethodHandleInternal& ctor, Boolean& bNeedSecurityCheck)
   at System.RuntimeType.CreateInstanceSlow(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at Microsoft.Web.LibraryManager.Build.Dependencies.<>c__DisplayClass10_0.<GetProvidersFromReflection>b__1(TypeInfo fac)
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at Microsoft.Web.LibraryManager.Build.Dependencies.GetProvidersFromReflection()
   at Microsoft.Web.LibraryManager.Build.Dependencies.Initialize()
   at Microsoft.Web.LibraryManager.Build.Dependencies..ctor(IHostInteraction hostInteraction, IEnumerable`1 assemblyPaths)
   at Microsoft.Web.LibraryManager.Build.RestoreTask.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
```

[...]

PR #524, opened 2019-09-27
Make the restore-on-build task run in its own app-domain

jimmylewis:
This addresses a recurring issue where doing a WebDeploy-based publish from within Visual Studio encounteres MissingMethodExceptions when running LibMan's Restore-on-build.  The issue is that VS Publish runs MSBuild in-proc, so it loads the build task DLL from the NuGet package, but the other LibMan assemblies are already loaded into the VS appdomain from the VSIX (often a newer version).  Any incompatibilities between these cause the build task to throw.

This fix causes the build task to be executed in its own appdomain, and as such it resolves all of its dependent assemblies from the nuget package.  The appdomain gets cleared when the build is finished, so we're not leaving duplicates of all the assmeblies loaded into VS.  Note: the Build task DLL itself is loaded into VS' primary appdomain, and will linger - this isn't a change from how it works today.

Even though AppDomainIsolatedTask is only available on net472, that's the build type that VS uses for the in-proc build, even for projects that only target netcoreapp*.

### DIFF
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
@@ -13,6 +13,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Framework" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" />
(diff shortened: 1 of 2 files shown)


jimmylewis:
Here's an image showing how the assemblies get loaded with this change (the 2.0.78 version is a private build based on the last published NuGet package but with this fix; 2.0.95 is the latest tip from master):
image

This is a close representation of #523 to show that this fix works.

PR #525, opened 2019-10-22
Enable VSSDK analyzers and fix analyzer issues

jimmylewis:
3 types of issues were found by analyzers:
- use of async void methods
- not checking for null on GetService calls
- awaiting foreign task

This PR supercedes #261.

### DIFF
--- a/src/LibraryManager.Vsix/Commands/RestoreOnBuildCommand.cs
+++ b/src/LibraryManager.Vsix/Commands/RestoreOnBuildCommand.cs
@@ -1,18 +1,18 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Design;
+using System.Linq;
 using EnvDTE;
-using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.VisualStudio.ComponentModelHost;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
-using NuGet.VisualStudio;
-using System;
-using System.ComponentModel.Design;
-using System.Linq;
-using System.Collections.Generic;
-using Task = System.Threading.Tasks.Task;
+using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.Web.LibraryManager.Vsix.Contracts;
+using NuGet.VisualStudio;
+using Task = System.Threading.Tasks.Task;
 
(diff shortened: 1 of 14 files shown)


src/LibraryManager.Vsix/Commands/RestoreOnBuildCommand.cs
```diff
@@ -1,37 +1,38 @@
...
             _package = package;
             _componentModel = VsHelpers.GetService<SComponentModel, IComponentModel>();
             _dependenciesFactory = dependenciesFactory;
 
             var cmdId = new CommandID(PackageGuids.guidLibraryManagerPackageCmdSet, PackageIds.RestoreOnBuild);
-            var cmd = new OleMenuCommand(ExecuteHandlerAsync, cmdId);
-            cmd.BeforeQueryStatus += BeforeQueryStatusHandlerAsync;
+            var cmd = new OleMenuCommand((s, e) => package.JoinableTaskFactory.RunAsync(() => ExecuteAsync(s, e)),
+                                         cmdId);
+            cmd.BeforeQueryStatus += (s, e) => _package.JoinableTaskFactory.RunAsync(() => BeforeQueryStatusAsync(s, e));
```
tlmii: Is this an intentional switch to `_package` (to avoid the capture?) since its being stored in the object? And if so, why not on the previous lambda on line 33?

src/LibraryManager.Vsix/Commands/RestoreSolutionCommand.cs
```diff
@@ -15,38 +15,21 @@ namespace Microsoft.Web.LibraryManager.Vsix
...
         {
             _package = package;
             _libraryCommandService = libraryCommandService;
 
             var cmdId = new CommandID(PackageGuids.guidLibraryManagerPackageCmdSet, PackageIds.RestoreSolution);
-            var cmd = new OleMenuCommand(ExecuteHandlerAsync, cmdId);
-            cmd.BeforeQueryStatus += BeforeQueryStatusHandlerAsync;
+            var cmd = new OleMenuCommand((s, e) => package.JoinableTaskFactory.RunAsync(() => ExecuteAsync(s, e)),
+                                         cmdId);
+            cmd.BeforeQueryStatus += (s, e) => _package.JoinableTaskFactory.RunAsync(() => BeforeQueryStatusAsync(s, e));
```
tlmii: Same comment about `_package` as above.

src/LibraryManager.Vsix/Commands/RestoreOnBuildCommand.cs
jimmylewis: oversight - the previous line should also be `_package`.  Nice catch

PR #534, opened 2019-11-07
Add support for file patterns in libman.json

jimmylewis:
This change adds support for file patterns (`*.js`, or `**/*.css`, or even negations like `!*.map`) in the libman.json file.  This has been a top requested feature since day 1.

File pattern support is a pretty big behavioral change (even if the diff is small).  While it is technically a breaking change (e.g. a file named `*.js` will now be expanded), I doubt it will actually impact anyone negatively.  It is easier though with the version bump to say that file patterns are supported on 2.1+ instead of 2.0.113+ (or whatever the actual release version is).

### DIFF
--- a/src/LibraryManager/Providers/BaseProvider.cs
+++ b/src/LibraryManager/Providers/BaseProvider.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.Web.LibraryManager.LibraryNaming;
+using Microsoft.Web.LibraryManager.Utilities;
 
 namespace Microsoft.Web.LibraryManager.Providers
 {
@@ -90,7 +91,7 @@ namespace Microsoft.Web.LibraryManager.Providers
             {
                 return LibraryOperationResult.FromCancelled(desiredState);
             }
-            
+
             string libraryId = LibraryNamingScheme.GetLibraryId(desiredState.Name, desiredState.Version);
             try
             {
@@ -104,7 +105,20 @@ namespace Microsoft.Web.LibraryManager.Providers
 
                 if (desiredState.Files != null && desiredState.Files.Count > 0)
                 {
-                    return CheckForInvalidFiles(desiredState, libraryId, library);
+                    // expand any potential file patterns
+                    IEnumerable<string> updatedFiles = FileGlobbingUtility.ExpandFileGlobs(desiredState.Files, library.Files.Keys);
+                    var processedState = new LibraryInstallationState
+                    {
+                        Name = desiredState.Name,
+                        Version = desiredState.Version,
+                        ProviderId = desiredState.ProviderId,
+                        DestinationPath = desiredState.DestinationPath,
+                        IsUsingDefaultDestination = desiredState.IsUsingDefaultDestination,
+                        IsUsingDefaultProvider = desiredState.IsUsingDefaultProvider,
+                        Files = updatedFiles.ToList(),
+                    };
+
+                    return CheckForInvalidFiles(processedState, libraryId, library);
                 }
 
                 desiredState = new LibraryInstallationState
--- a/src/LibraryManager/Providers/BaseProvider.cs
+++ b/src/LibraryManager/Providers/BaseProvider.cs
@@ -1,0 +1,298 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Web.LibraryManager.Contracts;
+using Microsoft.Web.LibraryManager.LibraryNaming;
+
+namespace Microsoft.Web.LibraryManager.Providers
+{
+    /// <summary>
+    /// Default implenentation for a provider, since most provider implementations are very similar.
+    /// </summary>
+    internal abstract class BaseProvider : IProvider
+    {
+        private readonly CacheService _cacheService;
+        private string _cacheFolder;
+
+        public BaseProvider(IHostInteraction hostInteraction, CacheService cacheService)
+        {
+            HostInteraction = hostInteraction;
+            _cacheService = cacheService;
+        }
+
+        #region IProvider implementation
+
+        /// <inheritdoc />
+        public abstract string Id { get; }
+
+        /// <inheritdoc />
+        public virtual string NuGetPackageId => "Microsoft.Web.LibraryManager.Build";
+
+        /// <inheritdoc />
+        public abstract string LibraryIdHintText { get; }
+
+        /// <inheritdoc />
+        public IHostInteraction HostInteraction { get; }
+
+        /// <inheritdoc />
+        public virtual bool SupportsLibraryVersions => true;
+
+        /// <inheritdoc />
+        public abstract ILibraryCatalog GetCatalog();
+
+        /// <inheritdoc />
+        public abstract string GetSuggestedDestination(ILibrary library);
+
+        /// <inheritdoc />
+        public virtual async Task<ILibraryOperationResult> InstallAsync(ILibraryInstallationState desiredState, CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return LibraryOperationResult.FromCancelled(desiredState);
+            }
+
+            //Expand the files property if needed
+            ILibraryOperationResult updateResult = await UpdateStateAsync(desiredState, cancellationToken);
+            if (!updateResult.Success)
+            {
+                return updateResult;
+            }
+
+            desiredState = updateResult.InstallationState;
+
+            // Refresh cache if needed
+            ILibraryOperationResult cacheUpdateResult = await RefreshCacheAsync(desiredState, cancellationToken);
+            if (!cacheUpdateResult.Success)
+            {
+                return cacheUpdateResult;
+            }
+
+            // Check if Library is already up to date
+            if (IsLibraryUpToDate(desiredState))
+            {
+                return LibraryOperationResult.FromUpToDate(desiredState);
+            }
+
+            // Write files to destination
+            return await WriteToFilesAsync(desiredState, cancellationToken);
+        }
+
+        /// <inheritdoc />
+        public virtual async Task<ILibraryOperationResult> UpdateStateAsync(ILibraryInstallationState desiredState, CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return LibraryOperationResult.FromCancelled(desiredState);
+            }
+            
+            string libraryId = LibraryNamingScheme.GetLibraryId(desiredState.Name, desiredState.Version);
+            try
+            {
+                ILibraryCatalog catalog = GetCatalog();
+                ILibrary library = await catalog.GetLibraryAsync(desiredState.Name, desiredState.Version, cancellationToken).ConfigureAwait(false);
+
+                if (library == null)
+                {
+                    return new LibraryOperationResult(desiredState, PredefinedErrors.UnableToResolveSource(desiredState.Name, desiredState.ProviderId));
+                }
+
+                if (desiredState.Files != null && desiredState.Files.Count > 0)
+                {
+                    return CheckForInvalidFiles(desiredState, libraryId, library);
+                }
+
+                desiredState = new LibraryInstallationState
+                {
+                    ProviderId = Id,
+                    Name = desiredState.Name,
+                    Version = desiredState.Version,
+                    DestinationPath = desiredState.DestinationPath,
+                    Files = library.Files.Keys.ToList(),
+                    IsUsingDefaultDestination = desiredState.IsUsingDefaultDestination,
+                    IsUsingDefaultProvider = desiredState.IsUsingDefaultProvider
+                };
+            }
+            catch (InvalidLibraryException)
+            {
+                return new LibraryOperationResult(desiredState, PredefinedErrors.UnableToResolveSource(libraryId, desiredState.ProviderId));
+            }
+            catch (UnauthorizedAccessException)
+            {
+                return new LibraryOperationResult(desiredState, PredefinedErrors.PathOutsideWorkingDirectory());
+            }
+            catch (Exception ex)
+            {
+                HostInteraction.Logger.Log(ex.ToString(), LogLevel.Error);
+                return new LibraryOperationResult(desiredState, PredefinedErrors.UnknownException());
+            }
+
+            return LibraryOperationResult.FromSuccess(desiredState);
+        }
+
+        #endregion
+
+        protected virtual ILibraryOperationResult CheckForInvalidFiles(ILibraryInstallationState desiredState, string libraryId, ILibrary library)
+        {
+            IReadOnlyList<string> invalidFiles = library.GetInvalidFiles(desiredState.Files);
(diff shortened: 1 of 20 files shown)


src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+            try
+            {
+                var librariesMetadata = new List<CacheFileMetadata>();
+                foreach (string sourceFile in state.Files)
+                {
+                    string cacheFile = Path.Combine(libraryDir, sourceFile);
+                    string url = GetDownloadUrl(state, sourceFile);
+
+                    var newEntry = new CacheFileMetadata(url, cacheFile);
+                    if (!librariesMetadata.Contains(newEntry))
```
alexgav: I don't think this Contains would every return, would it? CacheFileMetadata doesn't override Equals, and it's class - not a struct. So we are going to do reference equality checking here, right? Since it's a newly created object, that would never exist. Unless we override Equals().

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                await _cacheService.RefreshCacheAsync(librariesMetadata, cancellationToken);
+            }
+            catch (ResourceDownloadException ex)
+            {
+                HostInteraction.Logger.Log(ex.ToString(), LogLevel.Error);
+                return new LibraryOperationResult(state, PredefinedErrors.FailedToDownloadResource(ex.Url));
+            }
+            catch (OperationCanceledException)
+            {
+                return LibraryOperationResult.FromCancelled(state);
```
alexgav: We just swallow these? Do they happen a lot during normal execution? Is it basically a timeout? If it means we didn't refresh successfully, should we log this anyway?

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -8,52 +8,40 @@
...
+        public override bool SupportsLibraryVersions => false;
 
         /// <summary>
         /// Gets the <see cref="Microsoft.Web.LibraryManager.Contracts.ILibraryCatalog" /> for the <see cref="Microsoft.Web.LibraryManager.Contracts.IProvider" />. May be <code>null</code> if no catalog is supported.
         /// </summary>
         /// <returns></returns>
-        public ILibraryCatalog GetCatalog()
+        public override ILibraryCatalog GetCatalog()
         {
             return new FileSystemCatalog(this);
```
alexgav: Do we need a new instance every time? Other providers cache their catalogs

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -257,5 +195,12 @@ private static async Task<Stream> GetRemoteResourceAsync(string sourceUrl)
                 throw new ResourceDownloadException(sourceUrl);
             }
         }
+
+        protected override ILibraryNamingScheme LibraryNamingScheme { get; } = new SimpleLibraryNamingScheme();
+
+        protected override string GetDownloadUrl(ILibraryInstallationState state, string sourceFile)
+        {
+            throw new NotImplementedException();
```
alexgav: This is a little weird. No other way of doing it? I'm assuming this will never get called for the FileSystemProvider, still looks pretty screwy / bad smell. Maybe use "file:///" url, or re-design base class (create intermediate base class?)

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -1,98 +1,41 @@
...
 
         /// <summary>
         /// Returns the UnpkgLibrary's name.
         /// </summary>
         /// <param name="library"></param>
         /// <returns></returns>
-        public string GetSuggestedDestination(ILibrary library)
+        public override string GetSuggestedDestination(ILibrary library)
         {
             if (library != null && library is UnpkgLibrary unpkgLibrary)
```
alexgav: Can this ever get called on UpkgProvider with non-UnpkgLibrary? Same for other providers? I would guess it shouldn't, in which case we cane use use library.Name instead of casting to unpkgLibrary and move this method to base provider as well, no? cdnjs, jsdeliver and unpkg all do the same thing. File system could override.

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
```diff
@@ -8,52 +8,40 @@
...
-        /// <remarks>
-        /// If the provider doesn't have a NuGet package, then return <code>null</code>.
-        /// </remarks>
-        public string NuGetPackageId { get; } = "Microsoft.Web.LibraryManager.Build";
-
-        /// <summary>
-        /// An object specified by the host to interact with the file system etc.
-        /// </summary>
-        public IHostInteraction HostInteraction { get; }
+        public override string Id => "filesystem";
```
alexgav: Should we make these constants (in all providers)? Tests use those strings if nothing else. In this case, I noticed one test uses "fileSystem" instead of "filesystem", so I it seems it would fail on Linux/Mac?

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+        public abstract string LibraryIdHintText { get; }
+
+        /// <inheritdoc />
+        public IHostInteraction HostInteraction { get; }
+
+        /// <inheritdoc />
+        public virtual bool SupportsLibraryVersions => true;
+
+        /// <inheritdoc />
+        public abstract ILibraryCatalog GetCatalog();
```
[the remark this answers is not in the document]
jimmylewis: `GetCatalog()` is expected to be an expensive operation, typically requiring a network call.  It should not be a property.

src/LibraryManager/Utilities/FileGlobbingUtility.cs
```diff
@@ -0,0 +1,45 @@
...
+                if (potentialGlob.StartsWith("!", StringComparison.Ordinal))
+                {
+                    // Remove matches from the files list
+                    var filesToRemove = finalSetOfFiles.Where(f => Minimatcher.Check(f, potentialGlob, negatedOptions)).ToList();
+                    foreach (string file in filesToRemove)
+                    {
+                        finalSetOfFiles.Remove(file);
+                    }
+                }
+                else if (potentialGlob.Contains("*"))
```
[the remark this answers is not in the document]
jimmylewis: This isn't looking for disallowed characters (! is a valid file name character on Windows, and invalid path characters are [mostly a Windows-only limitation](https://stackoverflow.com/questions/1311037/are-there-any-invalid-linux-filenames)).  This is just looking for characters indicating that this is a file pattern, so that we know whether to attempt to expand it or to preserve it as a literal file name.  `?` would make sense to add here, as well as `[`.  There's some more esoteric patterns that minimatch also supports, but I don't know if they'd actually get used so I didn't look for them here.

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
jimmylewis: This isn't used by FileSystemProvider (it's only used in `RefreshCacheAsync` and FileSystemProvider doesn't use the cache), and doesn't really make sense to get a download URL in FS scenarios.  The other options here would be to either not make FileSystemProvider derive from BaseProvider, or - like you say - introduce an intermediate class for the InstallAsync implementation (which is the root of the code path where this comes into play).  Since FSProvider already overrides the InstallAsync implementation, this doesn't feel like a dirty architecture to me.

That said, the docs recommend using `NotSupportedException` in this case, so I'll change it to that.  Semantically, that's what I meant to say.

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
jimmylewis: Technically it would probably be fine, but I don't see why giving up the type check is critical.  This is not on a hot path, and if anything ought to change, it feels like we should be throwing when the `library is ProviderLibrary` check fails (assuming `library != null`) because then we've somehow breached sane expectations.

src/LibraryManager/Providers/FileSystem/FileSystemProvider.cs
jimmylewis: The bad test is a good catch, I'll fix that.

There's a reasonable argument against exposing the constants: if you exclusively rely on the constants in tests, you'll only be testing that the constant field is used, not that the value of that field is correct.  There is an equally reasonable counter proposal to this: add a single unit test asserting the value of the constant, then use the constant in all the other tests.

It looks like there's a lot of places in tests where we've just hardcoded the provider names out of convenience.  Fixing these to all be consistent doesn't seem to provide value worth the time to go fix up all the uses.  I'll leave this as a future exercise if we do that level of code cleanup; it doesn't seem to add any value in this PR.

(TL;DR: I agree, but not going to address it in this PR)

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                            return LibraryOperationResult.FromCancelled(state);
+                        }
+
+                        if (string.IsNullOrEmpty(file))
+                        {
+                            return new LibraryOperationResult(state, PredefinedErrors.CouldNotWriteFile(file));
+                        }
+
+                        string destinationPath = Path.Combine(state.DestinationPath, file);
+                        var sourceStream = new Func<Stream>(() => GetStreamAsync(state, file, cancellationToken).Result);
```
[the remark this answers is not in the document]
jimmylewis: That would require changing the HostInteraction.WriteFileAsync to take a `Func<Stream, Task>`.  *That* is the code which blocks when invoking this (even though it is also async in all 3 implementations).  I think we should change it, but in a separate PR.

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                    foreach (string file in state.Files)
+                    {
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            return LibraryOperationResult.FromCancelled(state);
+                        }
+
+                        if (string.IsNullOrEmpty(file))
+                        {
+                            return new LibraryOperationResult(state, PredefinedErrors.CouldNotWriteFile(file));
```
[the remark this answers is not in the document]
jimmylewis: I think I might split this into a separate PR, since it does also have loc impact (adding new string).

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+            return LibraryOperationResult.FromSuccess(state);
+        }
+
+        private async Task<Stream> GetStreamAsync(ILibraryInstallationState state, string sourceFile, CancellationToken cancellationToken)
+        {
+            if (!string.IsNullOrEmpty(state.Name) && !string.IsNullOrEmpty(state.Version))
+            {
+                string absolute = Path.Combine(CacheFolder, state.Name, state.Version, sourceFile);
+
+                if (File.Exists(absolute))
```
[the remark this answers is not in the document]
jimmylewis: That currently gets handled by the caller (`HostInteractions.WriteFileAsync`) returning false, which informs the provider to write an error message.

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+            try
+            {
+                if (!string.IsNullOrEmpty(state.Name) && !string.IsNullOrEmpty(state.Version))
+                {
+                    string cacheDir = Path.Combine(CacheFolder, state.Name, state.Version);
+                    string destinationDir = Path.Combine(HostInteraction.WorkingDirectory, state.DestinationPath);
+
+                    foreach (string sourceFile in state.Files)
+                    {
+                        var destinationFile = new FileInfo(Path.Combine(destinationDir, sourceFile).Replace('\\', '/'));
```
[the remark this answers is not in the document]
jimmylewis: This is normalizing to use '/', which is the only path separator we should be using.  (We don't even allow `\` in the `"destination"` value through the manifest schema, and I'm surprised that we do allow it on the `"files"` items.  Probably an oversight there, since it allows a mix.)

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                        var cacheFile = new FileInfo(Path.Combine(cacheDir, sourceFile).Replace('\\', '/'));
```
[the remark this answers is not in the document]
jimmylewis: Why?  Not following on this...

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                        if (!destinationFile.Exists || !cacheFile.Exists || !FileHelpers.AreFilesUpToDate(destinationFile, cacheFile))
+                        {
+                            return false;
+                        }
+                    }
+                }
+            }
+            catch
+            {
+                return false;
```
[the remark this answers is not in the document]
jimmylewis: Maybe; I didn't check if this leads to an error reported elsewhere (though I suspect it does).  There's a ton of places where we blindly catch exceptions in this code.  I'm going to call won't-fix on this for now; if we are going to address them, then we should do it across the board and re-enable CA1031.

src/LibraryManager/Providers/BaseProvider.cs
jimmylewis: This is the type thrown when a task is cancelled - so if we have cancel-able operations (like background tasks in VS), this will catch it here.

I don't know if they happen often, since we don't test cancellation heavily.

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                var librariesMetadata = new List<CacheFileMetadata>();
+                foreach (string sourceFile in state.Files)
+                {
+                    string cacheFile = Path.Combine(libraryDir, sourceFile);
+                    string url = GetDownloadUrl(state, sourceFile);
+
+                    var newEntry = new CacheFileMetadata(url, cacheFile);
+                    if (!librariesMetadata.Contains(newEntry))
+                    {
+                        librariesMetadata.Add(new CacheFileMetadata(url, cacheFile));
```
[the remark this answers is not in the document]
jimmylewis: I filed #535 to track a few of the items you called out here, since it's on the existing code.

src/LibraryManager/Providers/BaseProvider.cs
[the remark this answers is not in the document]
jimmylewis: Although, looking at those implementations, getting the catalog is not the expensive part, it's doing anything with it.  This might be worth considering.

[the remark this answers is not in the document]
jimmylewis: Since this is a change to the contract API, punting on it for this PR.

src/LibraryManager/Providers/BaseProvider.cs
jimmylewis: No, these would happen if the user cancels the operation, not when the operation times out.  And we do log it (see `LogMessageGenerator.GetOperationSummaryString()`).

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
jimmylewis: The other methods that take in an `ILibrary` are not part of an `IProvider` implementation.  They use the `ProviderId` property to retrieve the correct provider.  In that sense, there isn't a comparable situation - this is the only place where one could pass an `ILibrary` to the wrong `Provider`.

I know it's redundant, but it is a unique check (though similar) in each implementation.  I prefer to leave it as is.

Issue #512, opened 2019-08-12
Restoring library never completes


ManuelHaas:
### Functional impact
Restore never finishes

### Minimal repro steps
Try this config with VS 16.3.0:

{
  "version": "1.0",
  "defaultProvider": "cdnjs",
  "libraries": [
    {
      "library": "metisMenu@3.0.4",
      "destination": "wwwroot/lib/metisMenu/"
    },
    {
      "provider": "unpkg",
      "library": "devextreme@19.1.5",
      "destination": "wwwroot/lib/devextreme/"
    }
  ]
}

I already tried to clean the libs before restore, restarted VS and tried again. Always the same - it never ends.

### Actual result
Restore never finishes. Stopped restore after 1200 seconds. 
First libarary was restored, second not.
Not a single file was in the directory afterwards!

### Further technical details
Here is the log of stopping the restore after 20! minutes:

Restore operation started...
Restoring libraries for project ToSec.Presentation
Restoring library metisMenu@3.0.4... (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.css written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.css.map written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.js written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.js.map written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.min.css written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.min.css.map written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.min.js written to destination (ToSec.Presentation)
wwwroot/lib/metisMenu/metisMenu.min.js.map written to destination (ToSec.Presentation)
Restoring library devextreme@19.1.5... (ToSec.Presentation)
Restore operation completed
1 libraries restored in 1219,11 seconds
1 libraries failed to restore

ddweber:
I can confirm this issue. It seems as if UNPKG as library provider is the cause of this problem.
The GUI of Visual Studio sometimes even freezed/crashed when i tried searching for a lib on UNPKG via LibMan.

ddweber:
I tested the same libman.json via the CLI and it worked. It took ages but it executed without any errors.

dmitry-pavlov:
[...]

Issue #389, opened 2018-10-29
Installation of some libraries dramatically slower than alternative tooling


nwoolls:
### Functional impact
*Does the bug result in any actual functional issue, if so, what?*  

If the library is cached, installation is quick enough. However, without libraries in cache, the performance of installing / restoring a library is dramatically slower than alternatives such as NPM and Bower.

### Minimal repro steps
*What is the smallest, simplest set of steps to reproduce the issue. If needed, provide a project that demonstrates the issue.*  

1. `mkdir libman-speed-test`
1. `cd libman-speed-test/`
1. `npm cache clean --force`
1. `bower cache clean`
1. `libman cache clean`
1. `time npm install ckeditor@4.6.2`
1. `time bower install ckeditor#4.6.2 --save`
1. `time libman install ckeditor@4.6.2 --provider cdnjs --destination libman_libs/ckeditor`

### Expected result
*What would you expect to happen if there wasn't a bug*

LibMan should install / restore the specified library in an amount of time comparable to alternatives such as NPM and Bower.

### Actual result
*What is actually happening*

With the above steps, installing the specified library takes:
* NPM: 3.63s
* Bower: 8.49s
* LibMan: 2m47.92s

### Further technical details
*Optional, details of the root cause if known*

nwoolls:
FWIW jQuery UI is another package where this is reproducible:

* NPM: 3.57s
* Bower: 4.58s
* LibMan: 27.95s

AdamWillden:
This is killing productivity because it restores client libraries after every c# code change, if it didn't take so long this wouldn't be such an issue.

```
1>Restore operation started...
1>Restoring library @fortawesome/fontawesome-free@5.4.1...
1>Restoring library gasparesganga-jquery-loading-overlay@2.1.6...
1>Restoring library jquery@3.3.1...
1>Restoring library jquery-ajax-unobtrusive@3.2.6...
1>Restoring library jquery-validation@1.18.0...
1>Restoring library jquery-validation-unobtrusive@3.2.11...
1>
1>6 libraries restored in 59.33 seconds
```

Is there a workaround? I'm using `unpkg`, would using a different provider help?

nwoolls:
@AdamWillden IME the issue happens even with the CDNJS. That being said, IME UnPkg is even worse, including random timeout errors trying to restore packages (on a different file each time). The issue(s) with UnPkg are being discussed to some degree here: https://github.com/aspnet/LibraryManager/issues/344

[...]

PR #538, opened 2019-12-04
Download files in parallel

jimmylewis:
Attempting to address poor performance when restoring libraries (such as #389 and #512).  There are two parts to this change:

1. Introducing parallel downloads when populating the file cache.  Until now, we have downloaded files in serial order, which is very slow.  With these changes, we will allow up to 10 simultaneous requests for files (slightly more than most browsers).  Keeping the limit is important to avoid DDoSing a server, and to avoid queuing too many TPL tasks or exploding the load on the thread pool.
2. Adding new output messages showing when files are being downloaded.  This mitigates the perception that libman is hung by reporting progress more regularly throughout the operation.

### DIFF
--- a/src/LibraryManager/CacheService/CacheService.cs
+++ b/src/LibraryManager/CacheService/CacheService.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading;
@@ -16,7 +17,7 @@ namespace Microsoft.Web.LibraryManager
     /// </summary>
     public class CacheService
     {
-        // TO DO: Move these expirations to the provider 
+        // TO DO: Move these expirations to the provider
         private readonly int _catalogExpiresAfterDays = 1;
         private readonly int _metadataExpiresAfterDays = 1;
         private readonly int _libraryExpiresAfterDays = 30;
@@ -34,7 +35,7 @@ namespace Microsoft.Web.LibraryManager
         }
 
         /// <summary>
-        /// Defines the root cache directory.  This should be consistent across all LibMan tools. 
+        /// Defines the root cache directory.  This should be consistent across all LibMan tools.
         /// </summary>
         public static string CacheFolder
         {
@@ -124,12 +125,9 @@ namespace Microsoft.Web.LibraryManager
         }
 
         /// <summary>
-        /// Refreshes the cache for the given set of files if expired 
+        /// Refreshes the cache for the given set of files if expired
         /// </summary>
-        /// <param name="librariesCacheMetadata"></param>
-        /// <param name="cancellationToken"></param>
-        /// <returns></returns>
-        public async Task RefreshCacheAsync(IEnumerable<CacheFileMetadata> librariesCacheMetadata, CancellationToken cancellationToken)
+        public async Task RefreshCacheAsync(IEnumerable<CacheFileMetadata> librariesCacheMetadata, ILogger logger, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
@@ -139,7 +137,12 @@ namespace Microsoft.Web.LibraryManager
             {
                 if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-_libraryExpiresAfterDays))
                 {
-                    Task readFileTask = DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
+                    var readFileTask = Task.Run(async () =>
+                    {
+                        logger.Log(string.Format(Resources.Text.DownloadingFile, metadata.Source), LogLevel.Operation);
+                        await DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
+                    });
+
                     refreshTasks.Add(readFileTask);
                 }
             }
--- a/src/LibraryManager/CacheService/CacheService.cs
+++ b/src/LibraryManager/CacheService/CacheService.cs
@@ -19,7 +19,6 @@ namespace Microsoft.Web.LibraryManager
         // TO DO: Move these expirations to the provider
         private const int CatalogExpiresAfterDays = 1;
         private const int MetadataExpiresAfterDays = 1;
-        private const int LibraryExpiresAfterDays = 30;
         private const int MaxConcurrentDownloads = 10;
 
         private readonly IWebRequestHandler _requestHandler;
@@ -134,7 +133,7 @@ namespace Microsoft.Web.LibraryManager
 
             async Task DownloadFileIfNecessaryAsync(CacheFileMetadata metadata)
             {
-                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-LibraryExpiresAfterDays))
+                if (!File.Exists(metadata.DestinationPath))
                 {
                     logger.Log(string.Format(Resources.Text.DownloadingFile, metadata.Source), LogLevel.Operation);
                     await DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
--- a/src/LibraryManager/CacheService/CacheService.cs
+++ b/src/LibraryManager/CacheService/CacheService.cs
@@ -3,12 +3,11 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Web.LibraryManager.Contracts;
+using Microsoft.Web.LibraryManager.Utilities;
 
 namespace Microsoft.Web.LibraryManager
 {
@@ -18,9 +17,11 @@ namespace Microsoft.Web.LibraryManager
     public class CacheService
     {
         // TO DO: Move these expirations to the provider
-        private readonly int _catalogExpiresAfterDays = 1;
-        private readonly int _metadataExpiresAfterDays = 1;
-        private readonly int _libraryExpiresAfterDays = 30;
+        private const int CatalogExpiresAfterDays = 1;
+        private const int MetadataExpiresAfterDays = 1;
+        private const int LibraryExpiresAfterDays = 30;
+        private const int MaxConcurrentDownloads = 10;
+
         private readonly IWebRequestHandler _requestHandler;
 
         private static string CacheFolderValue;
@@ -60,7 +61,7 @@ namespace Microsoft.Web.LibraryManager
         /// <returns></returns>
         public async Task<string> GetCatalogAsync(string url, string cacheFile, CancellationToken cancellationToken)
         {
-            return await GetResourceAsync(url, cacheFile, _catalogExpiresAfterDays, cancellationToken);
+            return await GetResourceAsync(url, cacheFile, CatalogExpiresAfterDays, cancellationToken);
         }
 
         /// <summary>
@@ -72,7 +73,7 @@ namespace Microsoft.Web.LibraryManager
         /// <returns></returns>
         public async Task<string> GetMetadataAsync(string url, string cacheFile, CancellationToken cancellationToken)
         {
-            return await GetResourceAsync(url, cacheFile, _metadataExpiresAfterDays, cancellationToken);
+            return await GetResourceAsync(url, cacheFile, MetadataExpiresAfterDays, cancellationToken);
         }
 
         /// <summary>
@@ -129,25 +130,16 @@ namespace Microsoft.Web.LibraryManager
         /// </summary>
         public async Task RefreshCacheAsync(IEnumerable<CacheFileMetadata> librariesCacheMetadata, ILogger logger, CancellationToken cancellationToken)
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            await ParallelUtility.ForEachAsync(DownloadFileIfNecessaryAsync, MaxConcurrentDownloads, librariesCacheMetadata, cancellationToken);
 
-            List<Task> refreshTasks = new List<Task>();
-
-            foreach (CacheFileMetadata metadata in librariesCacheMetadata)
+            async Task DownloadFileIfNecessaryAsync(CacheFileMetadata metadata)
             {
-                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-_libraryExpiresAfterDays))
+                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-LibraryExpiresAfterDays))
                 {
-                    var readFileTask = Task.Run(async () =>
-                    {
-                        logger.Log(string.Format(Resources.Text.DownloadingFile, metadata.Source), LogLevel.Operation);
-                        await DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
-                    });
-
-                    refreshTasks.Add(readFileTask);
+                    logger.Log(string.Format(Resources.Text.DownloadingFile, metadata.Source), LogLevel.Operation);
+                    await DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
                 }
             }
-
-            await Task.WhenAll(refreshTasks).ConfigureAwait(false);
         }
     }
 }
--- a/src/LibraryManager/CacheService/CacheService.cs
+++ b/src/LibraryManager/CacheService/CacheService.cs
@@ -140,7 +140,6 @@ namespace Microsoft.Web.LibraryManager
                 if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-_libraryExpiresAfterDays))
                 {
                     Task readFileTask = DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
-                    await readFileTask;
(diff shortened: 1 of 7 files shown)


jimmylewis:
Some measurements I took comparing before and after.  I used the CLI tool to minimize overhead (e.g. from VS project system) to focus on the performance of restoring the files.  For each iteration, I deleted the files from disk, cleared the libman cache, and then re-reran restore.

Libraries used to test:

Test Size | Provider | Library used | # files | Size
-- | -- | -- | -- | --
Medium | JSDelivr | jquery@3.4.1 | 234 | 1,416,263
Large | JSDelivr | @fortawesome/fontawesome-free@5.11.2 | 1631 | 10,636,759
X-large | JSDelivr | devextreme@19.1.5 | 2018 | 144,183,992
Large | Unpkg | devextreme@19.1.5 | 1095 | 77,290,879

(Note: JSDelivr automatically creates minimized files for .js and .css, which is why the file count is higher than Unpkg)

Before changes (using 2.0.96):

  | Test 1 | Test 2 | Test 3 | Test 4 | Test 5 | Average (seconds)
-- | -- | -- | -- | -- | -- | --
JSDelivr Medium | 6.647 | 6.601 | 6.718 | 7.707 | 6.59 | 6.8526
JSDelivr   Large | 46.255 | 50.846 | 47.543 | 43.883 | 43.506 | 46.4066
JSDelivr X-Large | 243.241 | 239.685 | 241.075 | 572.358 | 226.768 | 304.6254
Unpkg Large | 138.025 | 148.368 | 143.815 | 122.61 | 128.367 | 136.237

With changes (private build):

  | Test 1 | Test 2 | Test 3 | Test 4 | Test 5 | Average (seconds)
-- | -- | -- | -- | -- | -- | --
JSDelivr Medium | 4.712 | 4.53 | 5.097 | 4.294 | 4.766 | 4.6798
JSDelivr Large | 27.712 | 27.611 | 27.402 | 28.546 | 25.911 | 27.4364
JSDelivr X-large | 249.437 | 217.413 | 109.197 | 168.489 | 135.379 | 175.983
Unpkg Large | 94.286 | 70.937 | 72.426 | 75.201 | 87.531 | 80.0762

Overall, we are seeing approximately a 40% decrease in the time waiting for restore to complete.

After taking these measurements, I realized that there's a large chunk of time being spent copying files from the downloaded cache to the restore location on disk.  In the Unpkg Large and JSDelivr X-Large scenarios, it's about 30 and 70 seconds respectively.  I'll look into that in a later, separate change.

Also not shown here is that the very first time I downloaded each of these libraries had a significantly longer delay.  My assumption is that they were being loaded into each provider's CDN cache, as I could only repro the first time I downloaded the library every two days or so.  I skipped these data points above as they seem to be external to libman (but for what they're worth: the first time I downloaded the Unpkg Large test was 3831 seconds, and the JSDelivr X-Large was 692 seconds using the 2.0.96 build).

src/LibraryManager/Utilities/ParallelUtility.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            var semaphore = new SemaphoreSlim(degreeOfParallelism, degreeOfParallelism);
+            var allTasks = new List<Task>();
+
+            foreach (T item in items)
+            {
+                await semaphore.WaitAsync();
+                cancellationToken.ThrowIfCancellationRequested();
+
+                Task task = DoAction(action, item, semaphore);
```
alexgav: Nit: I personally would've preferred to see Release call in this method (since WaitAsync is in this method) rather than in the helper method so that it's clear we are properly releasing. Or else I would've renamed your helper into something like DoActionAndRelease, again, so it's clear we are always releasing. So my suggestion would to either get rid of helper or rename it to make it clearer that it releases the semaphore.

test/LibraryManager.Test/CacheServiceTest.cs
```diff
@@ -109,24 +109,23 @@ public async Task GetCatalogAsync_Throws_OperationCanceled_WhenCancelled()
...
+            var desiredCasheFiles = new List<CacheFileMetadata>()
             {
                 new CacheFileMetadata(validUrl, libraryFile1_Path),
                 new CacheFileMetadata(validUrl, libraryFile2_Path)
             };
 
             tokenSource.Cancel();
-            await _cacheService.RefreshCacheAsync(desiredCasheFiles, null, tokenSource.Token);
+            await Assert.ThrowsExceptionAsync<OperationCanceledException>(async () =>
+                await _cacheService.RefreshCacheAsync(desiredCasheFiles, logger, tokenSource.Token));
```
alexgav: Nit: (existing) typo in desiredCasheFiles

test/LibraryManager.Test/Utilities/ParallelUtilityTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                    return Task.CompletedTask;
+                }, 1, Enumerable.Range(1, 3))
+            );
+
+            Assert.AreEqual("Expected Failure", exception.Message);
+            Assert.AreEqual(2, counter); // both 1 and 3 ran, 2 threw the exception
+        }
+
+        [TestMethod]
+        public async Task ForEachAsync_CancellationTokenSet_DoesNotRunAllTasks()
```
alexgav: I think at least theoretically this test can fail. If you read the doc at https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim?view=netframework-4.8, it says
```
 If multiple threads are blocked, there is no guaranteed order, such as FIFO or LIFO, that controls when threads enter the semaphore.
```

That would mean that after executing the Func for "1", it could execute Func for "3" before executing it for "2".

src/LibraryManager/Utilities/ParallelUtility.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            await Task.WhenAll(allTasks);
+        }
+
+        private static async Task DoAction<T>(Func<T, Task> act, T input, SemaphoreSlim s)
+        {
+            try
+            {
+                await act(input);
```
alexgav: Should this be Task.Run or are we relying on the Func to actually start a new thread? Our current usage is probably fine because async download file code will end up creating a new thread under the covers, but it seems that in general case this would be running on the same thread if your action Func didn't create a new thread, no? As in

https://stackoverflow.com/questions/27265818/does-the-use-of-async-await-create-a-new-thread

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -129,25 +130,16 @@ private async Task<string> GetResourceAsync(string url, string localFile, int ex
...
-            List<Task> refreshTasks = new List<Task>();
+            await ParallelUtility.ForEachAsync(DownloadFileIfNecessaryAsync, MaxConcurrentDownloads, librariesCacheMetadata, cancellationToken);
 
-            foreach (CacheFileMetadata metadata in librariesCacheMetadata)
+            async Task DownloadFileIfNecessaryAsync(CacheFileMetadata metadata)
             {
-                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-_libraryExpiresAfterDays))
+                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-LibraryExpiresAfterDays))
                 {
-                    var readFileTask = Task.Run(async () =>
```
alexgav: Should Task.Run have moved inside the new ForEachAsync utility code for general case to be correct? See my other comment in the next commit.

src/LibraryManager/Utilities/ParallelUtility.cs
jimmylewis: I can't release here.  There isn't a point in this method that I know the task has completed until after the `await Task.WhenAll()` below.  It's an implicit handshake between this method (using `semaphore.WaitAsync()`) to know when it can queue another job, and each job itself notifying when it's complete (`sempaphore.Release()`).  And I don't want to await each task as it's queued to be able to `Release` here without losing the parallel aspect (or, I end up chunking jobs, but that could mean that 9 files of <1KB are in the same chunk as a 1MB resource file, so I still lose parallel execution while waiting for the one big file; this is a realistic case with some larger libraries).  Also, if I call `WaitAsync()` in the `DoAction` helper, I'll end up queueing all the tasks immediately, which would add pressure to the scheduler when they'd all just get stuck waiting anyway.

The rename to `DoActionAndRelease` sounds reasonable.

src/LibraryManager/Utilities/ParallelUtility.cs
jimmylewis: Task.Run is blocking, so I don't want it here.  It would lead to starving up to `degreeOfParallelism` threads from the thread pool.

I don't care about creating new threads, the thread pool will manage that as the number of queued work items (Tasks) grows.  Since the long pull on the current usage of this is network I/O, being able to reuse threads while they are in an awaiting state is optimal anyways.

test/LibraryManager.Test/Utilities/ParallelUtilityTests.cs
jimmylewis: This test uses a `degreeOfParallelism` of 1, so "1" must completely execute before "2", which must in turn completely execute before "3".  Any different execution order would mean that the foreach iterator yielded a different input order.

If the test fails, then it would catch a change in the implementation that breaks this behavior (which seems desirable to me).

src/LibraryManager/Utilities/ParallelUtility.cs
alexgav: You are right, we can't move Release up. Renaming the method sounds great.

src/LibraryManager/Utilities/ParallelUtility.cs
alexgav: I mean Run/RunAsync, basically the question is - do we want to make sure we actually run things in parallel, because the current method doesn't actually insure parallelism AFAIK. Our current usage with network I/O will result in parallelism, but in general case it's not.

test/LibraryManager.Test/Utilities/ParallelUtilityTests.cs
alexgav: No, "2" doesn't not have to execute completely before "3". Tasks are not necessarily awaken/given semaphore in the order they were queued up. That's what documentation is saying.  When "1" is running, "2" and "3" are blocked and queued up to acquire the semaphore. In theory (implementation-dependent), "3" could get semaphore before "2" does after "1" completes.

jimmylewis: It does have to execute completely.  With DoP=1, the foreach will get the first item ("1").  It will acquire the semaphore.  It launch the new task, it will reach the end of the loop and iterate (now iterating on "2").  Then it blocks on the semaphore.  It can't launch the task for "2" until the task for "1" reaches the `semaphore.Release()`.  It can't even try to start "3" until it's launched "2" (at which point, it will be waiting on the semaphore again.  Remember, the foreach itself is not parallel, it's just a serial iteration that is launching a set of tasks (and it's ultimately up to the TaskScheduler to run those in parallel).

With DoP>1 your point is entirely right.  That's why I used DoP=1 in these tests, because this way I do have absolute control of order of execution this way.

src/LibraryManager/Utilities/ParallelUtility.cs
jimmylewis: Calling `DoAction(..)` above is already the equivalent of `Task.RunAsync(() => DoAction(...));`  So by this await, we're already running in a separate task from the code path that launched it (probably executing in parallel).

The foreach loop basically queues up as many Tasks as it can (let's go with 10), from a single thread, but it doesn't yield while it's doing so until it gets blocked on the `semaphore.WaitAsync()`.  If the other threads are sitting idle, they'll steal the work from this thread's job queue (that's what the default TaskScheduler does, and I only call out "default" here because it's always possible to write your own and pass it in as a parameter to [TaskFactory.StartNew](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskfactory.startnew?view=netframework-4.8)), and that's how we get them to run in parallel.  There isn't anything we can do to force parallel execution ourselves through async/await.  These language features only provide asynchronicity, not parallelism.  It's entirely up to the TaskScheduler to provide parallel execution (and even using [TaskCreationOptions.LongRunning](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcreationoptions?view=netframework-4.8) only provides a hint to the scheduler to create a new thread, but doesn't guarantee it).

src/LibraryManager/CacheService/CacheService.cs
jimmylewis: No.  the new ForEachAsync should not have `Task.Run` as I discussed above.

We used to need it here to add to `refreshTasks` and then do `Task.WhenAll(refreshTasks)` but all that task management was all moved into the `ForEachAsync` helper.

test/LibraryManager.Test/Utilities/ParallelUtilityTests.cs
alexgav: OK, looked at that loop again, and you are right. I don't think I read it right the first time. The first item over the limit blocks the loop executing, so we never have more than one item waiting on the semaphore. For some reason I thought (incorrectly) that we queued up all items, including the ones waiting on the semaphore in the allTasks, but it's obviously not the case :) 

---
In reply to: [362246750](https://github.com/aspnet/LibraryManager/pull/538#discussion_r362246750) [](ancestors = 362246750)

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -124,28 +126,20 @@ private async Task<string> GetResourceAsync(string url, string localFile, int ex
...
-            cancellationToken.ThrowIfCancellationRequested();
-
-            List<Task> refreshTasks = new List<Task>();
+            await ParallelUtility.ForEachAsync(DownloadFileIfNecessaryAsync, MaxConcurrentDownloads, librariesCacheMetadata, cancellationToken);
 
-            foreach (CacheFileMetadata metadata in librariesCacheMetadata)
+            async Task DownloadFileIfNecessaryAsync(CacheFileMetadata metadata)
             {
-                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-_libraryExpiresAfterDays))
+                if (!File.Exists(metadata.DestinationPath) || File.GetLastWriteTime(metadata.DestinationPath) < DateTime.Now.AddDays(-LibraryExpiresAfterDays))
```
ToddGrun: Might be nice if instead of  DateTime.Now.AddDays(-LibraryExpiresAfterDays), LibraryExpiresAfterDays was rather a TimeSpan and we could do DateTime.Now.Subtract instead.

ToddGrun: Could we use File.GetLastWriteTimeUtc and DateTime.UtcNow (and maybe cache the DateTime.UtcNow value)

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -124,28 +126,20 @@ private async Task<string> GetResourceAsync(string url, string localFile, int ex
...
                 {
-                    Task readFileTask = DownloadToFileAsync(metadata.Source, metadata.DestinationPath, attempts: 5, cancellationToken: cancellationToken);
-                    await readFileTask;
-                    refreshTasks.Add(readFileTask);
+                    logger.Log(string.Format(Resources.Text.DownloadingFile, metadata.Source), LogLevel.Operation);
```
ToddGrun: Just to make sure, the ILogger.Log implementation is thread safe, right?

src/LibraryManager/CacheService/CacheService.cs
```diff
@@ -4,10 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Web.LibraryManager.Contracts;
+using Microsoft.Web.LibraryManager.Utilities;
 
 namespace Microsoft.Web.LibraryManager
```
ToddGrun: is the convention to not use the folder name for public APIs?

src/LibraryManager/Resources/Text.resx
```diff
@@ -274,4 +274,7 @@
   <data name="JsDelivrProviderHintText" xml:space="preserve">
     <value>Example: &lt;library&gt;@&lt;version&gt;</value>
   </data>
+  <data name="DownloadingFile" xml:space="preserve">
+    <value>Downloading file {0}...</value>
```
ToddGrun: Unicode … char instead?

src/LibraryManager/Utilities/ParallelUtility.cs
```diff
@@ -0,0 +1,64 @@
...
+            action = action ?? throw new ArgumentNullException(nameof(action));
+            items = items ?? throw new ArgumentNullException(nameof(items));
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var semaphore = new SemaphoreSlim(degreeOfParallelism, degreeOfParallelism);
+            var allTasks = new List<Task>();
+
+            foreach (T item in items)
+            {
+                await semaphore.WaitAsync();
```
ToddGrun: pass in cancellationToken?

src/LibraryManager/CacheService/CacheService.cs
jimmylewis: It originally was mixed, and hasn't had a cleanup pass to make it all consistent.

src/LibraryManager/CacheService/CacheService.cs
jimmylewis: Yes, they all should be.

src/LibraryManager/CacheService/CacheService.cs
jimmylewis: Next cleanup item on my list is to remove this expiration anyway.  Package contents should be static for a given version, so it seems dumb to make you redownload them every 30 days just to get the same file contents back.

(Unless you're modifying files in the cache, in which case it's on your own head to restore or remove them.)

jimmylewis: Bah, I might as well just do this now, it'll also remove the Utc comment below. ;P

src/LibraryManager/Resources/Text.resx
jimmylewis: I'm going to resolve this for now, but will try to follow up with someone for better guidance and what we should do here.

PR #541, opened 2020-01-02
Make restore faster by copying files

jimmylewis:
Compared to the original implementation of reading files in from the cache and writing them back out to the destination, a simple file copy seems to be several times faster on average.

### DIFF
--- a/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
+++ b/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
@@ -113,6 +113,11 @@ namespace Microsoft.Web.LibraryManager.Vsix
             cancellationToken.ThrowIfCancellationRequested();
 
             string absoluteDestinationPath = Path.Combine(WorkingDirectory, destinationPath);
+            if (!FileHelpers.IsUnderRootDirectory(absoluteDestinationPath, WorkingDirectory))
+            {
+                throw new UnauthorizedAccessException();
+            }
+
             await VsHelpers.CheckFileOutOfSourceControlAsync(absoluteDestinationPath);
             bool result = await FileHelpers.CopyFileAsync(sourcePath, absoluteDestinationPath, cancellationToken);
 
--- a/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
+++ b/src/LibraryManager.Vsix/Contracts/HostInteraction.cs
@@ -107,14 +107,21 @@ namespace Microsoft.Web.LibraryManager.Vsix
             return FileHelpers.ReadFileAsStreamAsync(filePath, cancellationToken);
         }
 
-        public Task<bool> CopyFile(string sourcePath, string destinationPath, CancellationToken cancellationToken)
+        /// <inheritdoc />
+        public async Task<bool> CopyFileAsync(string sourcePath, string destinationPath, CancellationToken cancellationToken)
         {
-            return Task.Run(() =>
-            {
-                cancellationToken.ThrowIfCancellationRequested();
+            cancellationToken.ThrowIfCancellationRequested();
 
-                return FileHelpers.CopyFile(sourcePath, destinationPath);
-            }, cancellationToken);
+            string absoluteDestinationPath = Path.Combine(WorkingDirectory, destinationPath);
+            await VsHelpers.CheckFileOutOfSourceControlAsync(absoluteDestinationPath);
+            bool result = await FileHelpers.CopyFileAsync(sourcePath, absoluteDestinationPath, cancellationToken);
+
+            if (result)
+            {
+                Logger.Log(string.Format(LibraryManager.Resources.Text.FileWrittenToDisk, destinationPath.Replace(Path.DirectorySeparatorChar, '/')), LogLevel.Operation);
+            }
+
+            return result;
         }
 
         private Task<bool> DeleteFilesFromDisk(IEnumerable<string> filePaths, CancellationToken cancellationToken)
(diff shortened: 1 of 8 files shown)


jimmylewis:
Using similar measurements as in #538, but without clearing the cache (i.e. test measures just copying files from cache to project folder).  The test libraries:

Test Size | Provider | Library used | # files | Size
-- | -- | -- | -- | --
Medium | JSDelivr | jquery@3.4.1 | 234 | 1,416,263
Large | JSDelivr | @fortawesome/fontawesome-free@5.11.2 | 1631 | 10,636,759
X-large | JSDelivr | devextreme@19.1.5 | 2018 | 144,183,992

Baseline (2.0.96 CLI tool):

Test | Trial 1 | 2 | 3 | 4 | 5 
-- | -- | -- | -- | -- | -- 
Medium | 2.78 | 2.64 | 2.72 | 2.56 | 2.67
Large | 15.94 | 15.9 | 16.83 | 23.83 | 15.94
X-Large | 51.7 | 51.27 | 50.58 | 51.72 | 50.82

Results after change (private build):

Test | Trial 1 | 2 | 3 | 4 | 5 
-- | -- | -- | -- | -- | -- 
Medium | 0.77 | 0.77 | 0.76 | 1.04 | 0.86
Large | 2.63 | 2.66 | 2.6 | 2.5 | 2.75
X-Large | 3.2 | 3.12 | 3.19 | 3.21 | 3.11

src/libman/Contracts/HostInteraction.cs
```diff
@@ -97,11 +97,14 @@ public Task<Stream> ReadFileAsync(string relativeFilePath, CancellationToken can
             return FileHelpers.ReadFileAsStreamAsync(relativeFilePath, cancellationToken);
         }
 
-        public async Task<bool> CopyFile(string sourcePath, string destinationPath, CancellationToken cancellationToken)
+        public Task<bool> CopyFile(string sourcePath, string destinationPath, CancellationToken cancellationToken)
```
ToddGrun: should this be CopyFileAsync as it returns a task?

jimmylewis: It should be, but that would require changing the interface, and I don't want to make that change in this PR.

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -190,9 +194,9 @@ protected async Task<ILibraryOperationResult> WriteToFilesAsync(ILibraryInstalla
                             return new LibraryOperationResult(state, PredefinedErrors.CouldNotWriteFile(file));
                         }
 
+                        string sourcePath = GetCachedFileLocalPath(state, file);
                         string destinationPath = Path.Combine(state.DestinationPath, file);
-                        var sourceStream = new Func<Stream>(() => GetStreamAsync(state, file, cancellationToken).Result);
```
ToddGrun: still used?

jimmylewis: You mean GetStreamAsync?  It was removed below.

src/libman/Contracts/HostInteraction.cs
ToddGrun: OK, there's no public surface area concerns then, right? Should we log a separate issue to track that?

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
```diff
@@ -109,7 +109,7 @@ public Task<Stream> ReadFileAsync(string filePath, CancellationToken cancellatio
 
         public Task<bool> CopyFile(string sourcePath, string destinationPath, CancellationToken cancellationToken)
         {
-            return System.Threading.Tasks.Task.Run(() =>
+            return Task.Run(() =>
```
ToddGrun: Just so I'm clear, the File.Copy call is faster than using the actual async IO operations? IE, something like:

public async Task<bool> CopyFile(string sourcePath, string destinationPath, CancellationToken cancellationToken)
{
	bool success = true;

	try
	{
		using (FileStream sourceStream = File.Open(sourcePath, FileMode.Open))
		{
			using (FileStream destinationStream = File.Create(destinationPath))
			{
				await sourceStream.CopyToAsync(destinationStream, cancellationToken);
			}
		}
	}
	catch
	{
		success = false;
	}

	return success;
}

is slower?

src/libman/Contracts/HostInteraction.cs
jimmylewis: At the moment, I don't believe any of the public surface area in this project are being used elsewhere.  This is something that we should probably clean up when/if we get to establishing an extensibility story.

src/LibraryManager/Providers/BaseProvider.cs
ToddGrun: Ahh, missed that as I miss codeflow

src/libman/Contracts/HostInteraction.cs
ToddGrun: OK, will resolve. Seems easy enough to just change it now, but no big deal.

src/LibraryManager.Vsix/Contracts/HostInteraction.cs
jimmylewis: That's the gist of this PR (through a few abstraction layers), yes.

ToddGrun: The old code was slightly different though. It was doing quite a bit differently from this. Could you do me a favor and verify that the code I added above is also slower than your change?

jimmylewis: Based on our conversation and further diligence, you're right they're approximately equal.  I'll rewrite `CopyFile` => `CopyFileAsync` and make it actually async to match the rest of the file operations here.

jimmylewis:
One big caveat on the performance numbers I posted above: I'm finding that performance varies by a factor of about 2x just between running the tool as a DLL (`dotnet libman.dll restore`) vs. as a CLI tool (`libman restore`).  I did mix those up during my testing, which could account for some (but not all!) of the delta.  I'll try to follow up with the SDK folks about this next week.

PR #544, opened 2020-01-14
Fixup 2 failing Cdnjs unit tests

jimmylewis:
Some of the Cdnjs unit tests were still using live web requests.  When the requests started returning different data, it changed the behavior (as far as the unit tests' expectations were concerned).  This PR refactors some of the code so that we can mock out the requests to control the data coming back. and thereby make sure that the unit tests are testing against a consistent data set.

### DIFF
--- a/src/LibraryManager.Contracts/Caching/ICacheService.cs
+++ b/src/LibraryManager.Contracts/Caching/ICacheService.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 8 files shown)


src/LibraryManager.Contracts/Caching/ICacheService.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="url">Url to the provider catalog</param>
+        /// <param name="cacheFile">Where to store the provider catalog within the cache</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns></returns>
+        Task<string> GetCatalogAsync(string url, string cacheFile, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Returns library metadata from provided Url to cacheFile
+        /// </summary>
+        /// <param name="url"></param>
```
alexgav: ></param> [](start = 30, length = 8)

Url to the library?

PR #545, opened 2020-01-22
Fix VS integration tests

jimmylewis:
Several of these tests were encountering issues where VS's auto-format would interfere with the contents being typed.

I also updated the version of the VS test framework (Apex) used to the latest available package.

### DIFF
--- a/test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
+++ b/test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
@@ -93,7 +93,8 @@ namespace Microsoft.Web.LibraryManager.IntegrationTest
             Editor.Caret.MoveDown(1);
             Editor.KeyboardCommands.Type("{");
             Editor.KeyboardCommands.Enter();
-            Editor.KeyboardCommands.Type("\"provider\": \"cdnjs\",");
+            Editor.KeyboardCommands.Type("\"provider\":");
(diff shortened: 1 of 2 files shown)


test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
```diff
@@ -189,7 +194,8 @@ public void LibmanCompletion_VersionCompletionInDescendingOrder()
             Editor.Caret.MoveDown(1);
             Editor.KeyboardCommands.Type("{");
             Editor.KeyboardCommands.Enter();
-            Editor.KeyboardCommands.Type("\"provider\": \"unpkg\",");
+            Editor.KeyboardCommands.Type("\"provider\":");
+            Editor.KeyboardCommands.Type("unpkg\","); // auto-format will put the space & quote after the colon
```
alexgav: Nit: I'd maybe create a private TypeMember helper since we do it in so many places in this class

jimmylewis: I'm going to skip this for now.  Having too many different helpers for niche operations leads to a really confusing library (Apex is already bad enough with Editor.KeyboardCommands.Type vs Editor.Edit.InsertText vs Editor.Edit.InsertTextWithNewlines vs Editor.Edit.InsertTextInBuffer ...), and if we do add a helper it should be well designed to simplify things instead of adding just one more way to do things.  https://xkcd.com/927/

Issue #530, opened 2019-10-28
NPM-based providers should support @latest version


jimmylewis:
### Functional impact
*Does the bug result in any actual functional issue, if so, what?*  
If a user types `foo@latest` in the Add Client Side Library wizard, currently auto-completion replaces that with the selected item (typically the first).  Getting it to stay at `@latest` requires moving focus not through the Tab key, or editing the manifest after completing the wizard.

### Minimal repro steps
*What is the smallest, simplest set of steps to reproduce the issue. If needed, provide a project that demonstrates the issue.*  

1. Open the Add Client Side Library wizard
2. For the library name type jquery@latest then hit Tab

### Expected result
*What would you expect to happen if there wasn't a bug*  
the library name remains `jquery@latest`

### Actual result
*What is actually happening*  
the library name becomes `jquery@1.5.1` (currently 1.5.1 is listed first)

### Further technical details
*Optional, details of the root cause if known*

PR #546, opened 2020-01-23
Add support for @latest version in jsdelivr and unpkg

jimmylewis:
This addresses #530.

### DIFF
--- a/src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
+++ b/src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
@@ -20,6 +20,8 @@ namespace Microsoft.Web.LibraryManager.Providers.jsDelivr
         public const string LatestLibraryVersionUrl = "https://data.jsdelivr.com/v1/package/npm/{0}";
         public const string LibraryFileListUrlFormatGH = "https://data.jsdelivr.com/v1/package/gh/{0}/flat";
         public const string LatestLibraryVersionUrlGH = "https://data.jsdelivr.com/v1/package/gh/{0}";
+        public const string LatestVersionTag = "latest";
+
         private readonly INpmPackageInfoFactory _packageInfoFactory;
         private readonly INpmPackageSearch _packageSearch;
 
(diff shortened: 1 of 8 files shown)


src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -72,6 +74,11 @@ public async Task<ILibrary> GetLibraryAsync(string name, string version, Cancell
             }
 
             string libraryId = _libraryNamingScheme.GetLibraryId(name, version);
+            if(string.Equals(version, LatestVersionTag, StringComparison.Ordinal))
```
alexgav: Is it case sensitive?

alexgav: Nevermind, tried it and confirmed, it is.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -72,6 +74,11 @@ public async Task<ILibrary> GetLibraryAsync(string name, string version, Cancell
...
+            {
+                string latestVersion = await GetLatestVersion(libraryId, includePreReleases: false, cancellationToken);
```
alexgav: Nit: should this be GetLatestVersionAsync?

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -231,6 +239,13 @@ public async Task<CompletionSet> GetLibraryCompletionSetAsync(string libraryName
 
                         completions.Add(completionItem);
                     }
+
+                    // support @latest version
+                    completions.Add(new CompletionItem
```
alexgav: Does it end up at the very bottom? Seems like it should be at the very top, no?

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
alexgav: I wouldn't worry about it, looks like not added as a part of this PR.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
jimmylewis: It does go at the bottom, intentionally. Putting it first is like recommending it the most.  I wouldn't recommend using the `latest` version tag very often - only in experimental/prototypal situations.  It could bring in new dependencies randomly (i.e. as soon as they're released), which could include breaking changes (e.g. bootstrap 3.x -> 4.x was a big breaking change).  It's considerably a best-practice to always explicitly manage your dependency versions.

If we get feedback requesting it higher, then we can change the ordering.

Issue #107, opened 2018-06-07
Code cleanup: namespaces do not match folder structure


jimmylewis:
### Functional impact
None

### Further technical details
We have many classes across multiple projects where the namespace is set to the assembly name, but ignores the directory structure within the project.  For example:

    Microsoft.Web.LibraryManager.Vsix
        Json
            Completion
                CompletionController.cs <- uses namespace Microsoft.Web.LibraryManager.Vsix

However:
a) This defies normal .NET convention
b) We don't do it consistently within the codebase

We should make this all consistent and follow common convention by using the AssemblyName.Folder.Path namespace pattern.

jimmylewis:
Closing this with the changes submitted by @RobJohnston (thanks!).  The libman project is a special case and will remain exempted from following this rule.

PR #551, opened 2020-02-06
Cleanup namespace-folder mismatch

RobJohnston:
Consistently follow the common .NET namespace convention by using the AssemblyName.Folder.Path pattern.

The following namespaces were **not** created because a class exists with the same name as its folder:
 * Microsoft.Web.LibraryManager.CacheService
 * Microsoft.Web.LibraryManager.SemanticVersion
 * Microsoft.Web.LibraryManager.Vsix.ErrorList

No changes done to the "libman" project in the Microsoft.Web.LibraryManager.Tools namespace.

Partially addresses #107

### DIFF
--- a/src/LibraryManager.Build/Contracts/Logger.cs
+++ b/src/LibraryManager.Build/Contracts/Logger.cs
@@ -26,12 +26,12 @@ namespace Microsoft.Web.LibraryManager.Build.Contracts
         {
             switch (level)
             {
-                case LibraryManager.Contracts.LogLevel.Error:
+                case LogLevel.Error:
                     Errors.Add(message);
                     break;
-                case LibraryManager.Contracts.LogLevel.Operation:
-                case LibraryManager.Contracts.LogLevel.Task:
-                case LibraryManager.Contracts.LogLevel.Status:
+                case LogLevel.Operation:
+                case LogLevel.Task:
+                case LogLevel.Status:
                     Messages.Add(message);
                     break;
             }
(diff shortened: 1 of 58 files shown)


src/LibraryManager.Build/Contracts/Logger.cs
```diff
@@ -21,16 +21,16 @@ public void Clear()
             Errors.Clear();
         }
 
-        public void Log(string message, Contracts.LogLevel level)
+        public void Log(string message, LibraryManager.Contracts.LogLevel level)
```
jimmylewis: It would feel a little cleaner if we added a using for the LogLevel class instead of having to always reference it 3 levels (`LibraryManager.Contracts.LogLevel` is pretty verbose)

RobJohnston: Done.

src/LibraryManager.Build/Contracts/Logger.cs
```diff
@@ -21,16 +22,16 @@ public void Clear()
             Errors.Clear();
         }
 
-        public void Log(string message, Contracts.LogLevel level)
+        public void Log(string message, LogLevel level)
         {
             switch (level)
             {
-                case Contracts.LogLevel.Error:
+                case LibraryManager.Contracts.LogLevel.Error:
```
jimmylewis: Simplify these uses of `LogLevel` too?  Thanks!

RobJohnston: Sorry for missing that.  Done.

PR #554, opened 2020-03-24
Add more entropy to fake library name

jimmylewis:
Yes, yes, someone did create a package with poiuytrewq in the name.

### DIFF
--- a/test/LibraryManager.Test/Providers/Unpkg/NpmPackageSearchTest.cs
+++ b/test/LibraryManager.Test/Providers/Unpkg/NpmPackageSearchTest.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Web.LibraryManager.Test.Providers.Unpkg
         [DataTestMethod]
         [DataRow(null, 0)]
         [DataRow("", 0)]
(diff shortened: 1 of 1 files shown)


test/LibraryManager.Test/Providers/Unpkg/NpmPackageSearchTest.cs
```diff
@@ -13,7 +13,7 @@ public class NpmPackageSearchTest
         [DataTestMethod]
         [DataRow(null, 0)]
         [DataRow("", 0)]
-        [DataRow("poiuytrewq", 0)]
+        [DataRow("plajsfdlkjsfoiuytrewq", 0)]
```
tlmii: Could you use an uppercase letter to make sure that someone can't do that to you in the future (Since uppercase letters aren't allowed in new package names anymore but are valid for existing packages)?

tlmii: After posting that, I'm not realizing this is searching not looking for exact match, so that probably wouldn't matter. Nevermind.

Issue #552, opened 2020-02-15
support for setting proxy via HTTPS_PROXY environment variable


bbodenmiller:
Nearly every tool supports reading from `HTTPS_PROXY` and `HTTP_PROXY` environment variables but this one does not seem to. Please add support to set via that rather than just via `libman config --set http_proxy=http://example`.

PR #556, opened 2020-03-25
Read settings from environment variables

jimmylewis:
Addresses #552

### DIFF
--- a/src/LibraryManager/Configuration/ProxySettings.cs
+++ b/src/LibraryManager/Configuration/ProxySettings.cs
@@ -45,7 +45,8 @@ namespace Microsoft.Web.LibraryManager.Configuration
             if (_bypassUris == null)
             {
                 _bypassUris = new List<Uri>();
-                if (_settings.TryGetValue(Constants.HttpProxyBypass, out string bypassValue))
+                if (_settings.TryGetValue(Constants.HttpsProxyBypass, out string bypassValue)
+                    || _settings.TryGetValue(Constants.HttpProxyBypass, out bypassValue))
                 {
                     string[] bypasses = bypassValue.Split(';');
                     foreach (string bypass in bypasses)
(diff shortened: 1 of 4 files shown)


src/LibraryManager/Configuration/ProxySettings.cs
```diff
@@ -45,7 +45,8 @@ private bool CheckForProxyBypass(Uri uri)
             if (_bypassUris == null)
             {
                 _bypassUris = new List<Uri>();
-                if (_settings.TryGetValue(Constants.HttpProxyBypass, out string bypassValue))
+                if (_settings.TryGetValue(Constants.HttpsProxyBypass, out string bypassValue)
+                    || _settings.TryGetValue(Constants.HttpProxyBypass, out bypassValue))
```
tlmii: This might be a dumb question, but is it possible for Constants.HttpsProxyBypass to be specified (such that it is picked up here over the Http version) but not Constants.HttpsProxy (such that the Http version is picked up in TryConfigureProxy? It looks like the settings are json, so that might not be possible, but I wanted to confirm.

test/LibraryManager.Test/Configuration/SettingsTest.cs
```diff
@@ -143,6 +143,23 @@ public void TryGetValue_ValueExistsInDifferentCase_ReturnFalse()
             Assert.IsFalse(success);
         }
 
+        [TestMethod]
+        public void TryGetValue_IfSettingPresentAmongEnvironmentVariables_UseEnvironmentValue()
+        {
+            string fileText = @"{ ""config"": { ""testKey"": ""fileValue"" } }";
+            File.WriteAllText(TestFilePath, fileText);
+
+            Environment.SetEnvironmentVariable("testKey", "envValue");
```
tlmii: No concern about using a real environment variable for proxy in the tests versus some type of fake?

src/LibraryManager/Configuration/ProxySettings.cs
jimmylewis: In both the bypass and proxy configuration code paths, we prefer https over http.  However in the bypass case they can be mixed - for example, if you define only `http_proxy` and `https_proxy.bypass`, then you'll use the https bypass value.  However, I went with the assumption that someone will use either `http_proxy` or `https_proxy` variables and not both.  Since this code path doesn't know whether the proxy settings are using `http_proxy` or `https_proxy`, I'm inclined to keep the simpler code for now until we get feedback to do otherwise.

tlmii: Fair enough. Just wanted to make sure that was intentional.

test/LibraryManager.Test/Configuration/SettingsTest.cs
jimmylewis: Mostly a matter of hope ;)  And it's not limited to this new test, if the environment variable exists, it will impact all of the other tests in this class.

I thought about abstracting something like `IEnvironment`, but that complicated how the Settings get constructed... and we already have concrete dependencies on the file system in these tests, and making it pure quickly turned into a much larger change than I wanted to make now.

PR #557, opened 2020-03-27
Create new PR/CI pipeline for Azure Pipelines

jimmylewis:
This replaces the broken appveyor build pipeline.  

Microbuild plugins and release/archival related tasks are for internal builds only, and won't light up in public builds.  There may be some kinks in those still to work out once an official build pipeline is in place.


azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
+trigger:
+  branches:
+    include: ["master", "rel/*"]
+  paths:
+    exclude: ["*.md"]
+
+variables:
+  # TreatWarningsAsErrors: true
```
alexgav: >#  [](start = 2, length = 2)

Nit: is this commented out? If so, do we need this line at all?

jimmylewis: I aspire to turn it back on someday, but I can remove it for now, since it seems like MAT warnings are impossible to suppress. :)

alexgav: No worries, as long as you meant to leave it in, it's all good.

---
In reply to: [399538253](https://github.com/aspnet/LibraryManager/pull/557#discussion_r399538253) [](ancestors = 399538253)

PR #558, opened 2020-03-27
Remove '@' from recommended installation path

jimmylewis:
In some contexts, notably ASP.NET Razor, the @@ escape sequence wasn't
supported in tag attributes (such as <script src="lib/@@foo"/>).

Removing the @ sign avoids the headache when this issue is encountered.
Scoped packages will still be pathed uniquely, just without the @
prefix.

This addresses https://developercommunity.visualstudio.com/content/problem/923113/signalr-issue-with.html.

### DIFF
--- a/src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
+++ b/src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
@@ -37,9 +37,9 @@ namespace Microsoft.Web.LibraryManager.Providers.Unpkg
         /// <returns></returns>
         public override string GetSuggestedDestination(ILibrary library)
         {
-            if (library != null && library is UnpkgLibrary unpkgLibrary)
+            if (library is UnpkgLibrary unpkgLibrary)
             {
-                return unpkgLibrary.Name.TrimStart('@');
+                return unpkgLibrary.Name?.TrimStart('@');
             }
 
             return string.Empty;
--- a/src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
+++ b/src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Web.LibraryManager.Providers.Unpkg
         {
             if (library != null && library is UnpkgLibrary unpkgLibrary)
             {
-                return unpkgLibrary.Name;
+                return unpkgLibrary.Name.TrimStart('@');
             }
 
             return string.Empty;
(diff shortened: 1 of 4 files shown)


test/LibraryManager.Test/Providers/JsDelivr/JsDelivrProviderTest.cs
```diff
@@ -222,23 +222,24 @@ public async Task InstallAsync_WithGlobPatterns_CorrectlyInstallsAllMatchingFile
...
+        [DataTestMethod]
+        [DataRow("jquery", "jquery")]
+        [DataRow("@angular/cli", "angular/cli")]
+        public void GetSuggestedDestination(string libraryName, string expected)
+        {
             var library = new JsDelivrLibrary()
             {
-                Name = "jquery",
+                Name = libraryName,
                 Version = "3.3.1",
```
alexgav: >"3.3.1 [](start = 26, length = 6)

Nit: Should this be a part of the DataRow data or removed if it's not needed/doesn't matter? Same for UnpkgProviderTest

jimmylewis: Doesn't matter, it only looks at the name.

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -39,7 +39,7 @@ public override string GetSuggestedDestination(ILibrary library)
         {
             if (library != null && library is UnpkgLibrary unpkgLibrary)
```
alexgav: >library != null && l [](start = 16, length = 20)

Nit: is this part needed?

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -39,7 +39,7 @@ public override string GetSuggestedDestination(ILibrary library)
...
             {
-                return unpkgLibrary.Name;
+                return unpkgLibrary.Name.TrimStart('@');
```
alexgav: >Name [](start = 36, length = 4)

Nit: Ever possible for the name to be null? I assume not, otherwise we could do ?.

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
jimmylewis: Hm, I guess not, I can clean these up.  They've just been there since forever.

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
jimmylewis: It is technically possible (though I don't think it would happen currently), but better safe than sorry. 👍

PR #576, opened 2020-04-20
Builds from master should be versioned for public release

jimmylewis:


jimmylewis:
This will instead be done by setting a variable in the official build pipeline, so that it's clear if a build is actually meant for release, regardless of whether it originated from master.

PR #577, opened 2020-04-21
Create VS Installer manifest and setup for pipeline insertion to VS

jimmylewis:
This PR contains the plumbing necessary to insert VSIX builds into the Visual Studio product (which will be done from an internal pipeline).  Namely:
- Creates a .vsman file
- Pushes the .vsman file and the .vsix to the VSTS Drop
- Creates a file containing variables that are easily calculated here (like build versions) but which need to be consumed in the pipeline.

### DIFF
--- a/build/MicroBuild.Plugins.props
+++ b/build/MicroBuild.Plugins.props
@@ -1,0 +1,27 @@
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <PropertyGroup>
+        <!-- MicroBuild plugins can be installed in any parent directory up the tree from the project that is building.
+        The use of a sentinel file (which will be installed automatically due to package dependencies from any plugin)
+        will help us locate the plugin packages directory while being constrained to using MSBuild's intrinsic functions,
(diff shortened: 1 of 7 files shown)


build/MicroBuild.Plugins.props
```diff
@@ -0,0 +1,27 @@
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
tlmii, line 1: This comes from the Microbuild package, right? What's the story on including it directly?

jimmylewis: The .vsmanproj doesn't really support the PackageReference to MicroBuild (at least, I couldn't cajole it into working). So, in order to support building locally (which for that project requires the Swix plugin), you'd need to download the MicroBulid Swix package, and then set MicroBuildOverridePluginDirectory to point to where that is.  It's not elegant, but it works.

I don't know if the rest of this file will ever be used aside from the last override, but I reproduced it here just in case.

jimmylewis: (I guess that's not limited to building locally, the CI needs to have this target too.  But it already knows to install the plugin and set the environment variable accordingly.)

PR #584, opened 2020-05-08
Make tests less prone to breaking when jQuery releases.

jimmylewis:
Ideally these wouldn't make any live web requests, but we don't have that level of DI built into the CLI commands yet (or their tests) yet.  Using specific versions will still test the scenarios, but we won't have tests break whenever a new jQuery release is published.

### DIFF
--- a/test/libman.Test/LibmanInstallTest.cs
+++ b/test/libman.Test/LibmanInstallTest.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Web.LibraryManager.Tools.Test
             var command = new InstallCommand(HostEnvironment);
             command.Configure(null);
 
-            _ = command.Execute("jquery@3.2.1", "--provider", "cdnjs", "--destination", "wwwroot");
+            command.Execute("jquery@3.2.1", "--provider", "cdnjs", "--destination", "wwwroot");
 
             Assert.IsTrue(File.Exists(Path.Combine(WorkingDir, "libman.json")));
 
@@ -58,7 +58,7 @@ namespace Microsoft.Web.LibraryManager.Tools.Test
             var command = new InstallCommand(HostEnvironment);
             command.Configure(null);
 
-            _ = command.Execute("jquery@3.2.1", "--destination", "wwwroot");
+            command.Execute("jquery@3.2.1", "--destination", "wwwroot");
 
             Assert.IsTrue(File.Exists(Path.Combine(WorkingDir, "libman.json")));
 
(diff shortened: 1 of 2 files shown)


test/libman.Test/LibmanInstallTest.cs
```diff
@@ -186,8 +186,7 @@ public void TestInstall_WithExistingLibmanJson_SpecificFiles()
 }";
 
             File.WriteAllText(Path.Combine(WorkingDir, "libman.json"), initialContent);
-
-            int result = command.Execute("jquery@3.2.1", "--files", "jquery.min.js");
+            _ = command.Execute("jquery@3.2.1", "--files", "jquery.min.js");
```
alexgav, line 189: >_ =  [](start = 12, length = 4)

Nit: Do we need this discard at all? It's not async, right? Also, last method doesn't use a discard, so we should be at least consistent.

test/libman.Test/LibmanUpdateTest.cs
```diff
@@ -138,20 +138,15 @@ public void TestUpdateCommand_AlreadyUpToDate()
...
             string libmanjsonPath = Path.Combine(WorkingDir, "libman.json");
             File.WriteAllText(libmanjsonPath, contents);
 
-            var restoreCommand = new RestoreCommand(HostEnvironment);
-            restoreCommand.Configure(null);
-
-            restoreCommand.Execute();
+            var installCommand = new InstallCommand(HostEnvironment);
+            installCommand.Configure(null);
+            installCommand.Execute("jquery --files jquery.min.js --files jquery.js".Split(' '));
```
alexgav, line 149: >("jquery --files jquery.min.js --files jquery.js".Split(' ')); [](start = 34, length = 62)

Looks like we are not specifying a version anywhere in the command? Should we?

jimmylewis: Not in this case.  This test is verifying that when the library is already up to date with the latest, then there should be no changes.  Because it restored using a constant version before, it would break every time a new version shipped (no longer restoring the latest version).

The change here is to remove the constant version and always install the latest, then run update and verify there's no change.

test/libman.Test/LibmanInstallTest.cs
jimmylewis: I don't think we do.  There was a squiggle and the default lightbulb was to replace the unused local with a discard, but you're right we can just remove the assignment.  I'll clean this up.

PR #599, opened 2020-09-02
Enable theme support for the Add Client Library wizard

jimmylewis:
This enables the wizard to respect the light and dark themes in the VS IDE.  

I also changed the control layouts so that all of the left margins are aligned vertically.  We had 3 different levels of indentation making the dialog look really sloppy.


src/LibraryManager.Vsix/UI/Models/PackageItem.cs
```diff
@@ -128,6 +138,28 @@ public string Name
...
+                return _moniker;
+            }
+            set
+            {
+                // doesn't matter what value is passed in, this is the logic to set it
+                ImageMoniker newMoniker = (ItemType == PackageItemType.Folder)
+                    ? (IsExpanded ? KnownMonikers.FolderOpened : KnownMonikers.FolderClosed)
+                    : WpfUtil.GetImageMonikerForFile(Name);
+                Set(ref _moniker, newMoniker);
+            }
```
tlmii, line 160: This seems awkward for a public setter. It doesn't seem to be referenced anywhere other than the one `Icon = default` above. Would this be better as a getter only property with a private ClearIcon method (or similar) instead?

tlmii (approved):
With the caveat that I have minimal exposure to theming in VS yet, this looks good other than my one suggestion.

src/LibraryManager.Vsix/UI/Models/PackageItem.cs
jimmylewis: I originally did it above, but then though that if there were ever a 2nd call path, it ought to duplicate the logic (which is not an intuitive code block).  That's why I put it here instead, along with the comment.  Until we have a case where we would need to override the value on a specific code path (but not on any other), I think having the logic here, while weird, is still better.

An alternative would be to have a `SetIconMoniker()` method for setting the moniker, but that makes it deviate more from the other properties' style.  And I dislike having a property with a getter and a separate Set* method.  If it comes up again, maybe that would be a better way to go, along with making the property setter private to ensure that the logic is kept consistent.

Issue #370, opened 2018-09-05
Providers are unreliable


spongessuck:
### Functional impact
Builds fail because required lib files don't exist in project.

### Minimal repro steps
1. Install files from a library (i.e. ladda@1.0.6, files: dist/ladda.min.js)
2. Ensure it works successfully
3. Later on, try to restore

### Expected result
Files that restore successfully once should always restore successfully

### Actual result
LibMan intermittently fails with error LIB002 or LIB018, or doesn't load the correct contents of the package (i.e., index.html instead of dist/**).

justcla:
Thanks for raising the issue. Is there a way to reliably repro this?
Are you able to post the error output?

catriname:
I had a similar issue with two libraries (bulma-checkradio and datatables-bulma) using the unpkg provider.  Intellisense would find the libraries, but not the files, leaving me with something like this:

`    {"provider": "unpkg",
      "library": "bulma-checkradio",
      "destination": "wwwroot/lib/bulma",
      "files": []
    },
`
and errors:
The "bulma-checkradio" library could not be resolved by the "unpkg" provider
The "datatables-bulma" library could not be resolved by the "unpkg" provider
Restore operation completed with errors

I was able to specify the same packages in package.json and bring them from _node_modules into my library folder using libman and provider filesystem, but not directly with libman / provider unpkg.

Mamikel:
I don't know whether this provides any assistance, but FWIW, I'm finding that the **libman restore** command works reliably with cdnjs, but referencing the provider **unpkg** which works within the IDE consistently fails via the restore command resolution errors:
[LIB002]: The "material-design-icons-font@2.2.3" library could not be resolved by the "unpkg" provider
[LIB002]: The "roboto-fontface@0.4.5
[libman.json.txt](https://github.com/aspnet/LibraryManager/files/2609363/libman.json.txt)

blakepell:
I have a similar behavior.  You add your package in via cdnjs (because unpkg almost never works, in fact I've never personally got it it to work).  Some packages you select will come down, but others fail consistently.  For instance, try to add the monaco-editor and it will consistently returns 404 for half of the files.  I'm starting to think it's a CDN problem all together.  Example:

**Failed to download resource from "https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.15.6/min-maps/vs/loader.js.map"**

[...]

PR #600, opened 2020-09-14
Add caching for provider metadata

### LINKED ISSUE #598: Request: Add metadata cache for jSDelivr and Unpkg providers
Based on https://github.com/aspnet/LibraryManager/issues/370#issuecomment-637953775 and https://github.com/aspnet/LibraryManager/issues/389#issuecomment-579920081, the library metadata for the Unpkg and JSDelivr providers is not yet cached.  This causes libman to make some requests on every restore using those providers, which should be unnecessary.  Addressing this may resolve some speed and reliability issues.

Is anyone currently working on this or is it up for grabs?

jimmylewis:
This affects the scenarios where we need to fetch the latest library versions or the file lists for a given library.  For Cdnjs, because they offer a full catalog, it also affects that.

When fetching versions (or catalog), we want to have live data if possible, as new versions (or packages) can be released.  These operations will initially attempt a web request, and if it fails, fall back to the cached contents on disk (if any).  Note that there is now a 1-day cache lifetime applied to the initial request, which should cut down on repetitive traffic.

When fetching file lists, we assume that released libraries are static.  In this case, we first check for the information on disk, and then fallback to a web request if it's missing.  There is no lifetime applied in this case.  If the library is mutated, a user will have to clear the cache to get the updated metadata for the library (but, they'd have to do the same to get updated versions of any cached library assets as well).

This resolves #598, which should in turn help with some of the reliability issues reported in #370.


src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
```diff
@@ -245,7 +245,7 @@ private async Task<bool> EnsureCatalogAsync(CancellationToken cancellationToken)
 
             try
             {
-                string json = await _cacheService.GetCatalogAsync(RemoteApiUrl, _cacheFile, cancellationToken).ConfigureAwait(false);
+                string json = await _cacheService.GetContentsFromUriWithCacheFallbackAsync(CatalogUrl, _cacheFile, cancellationToken);
```
RobJohnston, line 248: All the existing `await`ed tasks in this class have `ConfigureAwait(false)`, except the 2 calls to the new methods in `_cacheService`.  Was that an oversight?

jimmylewis: Yes, it was.  Thanks for catching it!

src/LibraryManager/Providers/Unpkg/UnpkgProvider.cs
```diff
@@ -25,7 +25,7 @@ public UnpkgProvider(IHostInteraction hostInteraction, CacheService cacheService
         public override ILibraryCatalog GetCatalog()
         {
             // TODO: sort out the WebRequestHandler dependency
```
RobJohnston, line 27: This comment seems like it no longer applies, since the `WebRequestHandler.Instance` parameter was removed, no?

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -29,16 +31,24 @@ internal class JsDelivrCatalog : ILibraryCatalog
         private readonly string _providerId;
         private readonly ILibraryNamingScheme _libraryNamingScheme;
         private readonly ILogger _logger;
-        private readonly IWebRequestHandler _webRequestHandler;
-
-        public JsDelivrCatalog(string providerId, ILibraryNamingScheme namingScheme, ILogger logger, IWebRequestHandler webRequestHandler, INpmPackageInfoFactory packageInfoFactory, INpmPackageSearch packageSearch)
+        private readonly ICacheService _cacheService;
+        private readonly string _cacheFolder;
+
+        public JsDelivrCatalog(string providerId,
```
RobJohnston, line 37: I wonder if you have considered, for consistency with the `CdnsCatalog` class, to pass the `jsDelivrProvider` in place of `providerId` and `cacheFolder`?  Same can be said for the `UnpkgCatalog` class.

src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
RobJohnston: Same applies to these method calls in the `JsDelivrCatalog` and `UnpkgCatalog` classes.

test/LibraryManager.Test/Providers/Unpkg/UnpkgCatalogTest.cs
```diff
@@ -1,12 +1,16 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
tlmii, line 8: I don't think this is needed?

tlmii (approved):
Along with the couple things Rob pointed out already, looks good to me!

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
jimmylewis: Actually, at some point I'm planning to do the opposite: make CdnjsCatalog take simple values instead of the provider.  This better decouples the two types, and would simplify construction of the CdnjsCatalog during testing as we would no longer have to supply a CdnjsProvider with all of _its_ dependencies.  But that's out of the scope of this change...

src/LibraryManager/Providers/Unpkg/UnpkgCatalog.cs
```diff
@@ -25,16 +27,18 @@ internal class UnpkgCatalog : ILibraryCatalog
         private readonly string _providerId;
         private readonly ILibraryNamingScheme _libraryNamingScheme;
         private readonly ILogger _logger;
-        private readonly IWebRequestHandler _webRequestHandler;
+        private readonly ICacheService _cacheService;
+        private readonly string _cacheFolder;
 
-        public UnpkgCatalog(string providerId, ILibraryNamingScheme namingScheme, ILogger logger, IWebRequestHandler webRequestHandler, INpmPackageInfoFactory packageInfoFactory, INpmPackageSearch packageSearch)
+        public UnpkgCatalog(string providerId, ILibraryNamingScheme namingScheme, ILogger logger, INpmPackageInfoFactory packageInfoFactory, INpmPackageSearch packageSearch, ICacheService cacheService, string cacheFolder)
```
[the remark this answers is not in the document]
jimmylewis: Yeah, there's a lot about the cache that I'd like to change, but it's just a pain to refactor and I haven't had time.  Totally agree on this though.

PR #606, opened 2020-09-17
Fix latest prerelease version for Cdnjs and JSDelivr

jimmylewis:
Cdnjs changed the ordering when they report package versions, so we need to correctly search for the latest.  (Ironically, our test mock data was already setup for this, but we didn't test it correctly.)

The JsDelivrCatalog implementation never handled prerelease versions, nor did it handle versions at all for GitHub releases.  These were both added.

Unpkg curently doesn't expose package versions via an API that we can easily consume (it's available, but as JSON content embedded in an HTML page... ew) so it was not changed as a part of this.

This change should address https://developercommunity.visualstudio.com/content/problem/1167156/check-for-updates-quick-action-broken-in-libmanjso.html

### DIFF
--- a/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
+++ b/src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
@@ -194,7 +194,10 @@ namespace Microsoft.Web.LibraryManager.Providers.Cdnjs
             }
 
             string first = includePreReleases
-                ? (await GetLibraryVersionsAsync(group.DisplayName, cancellationToken).ConfigureAwait(false)).First()
+                ? (await GetLibraryVersionsAsync(group.DisplayName, cancellationToken).ConfigureAwait(false))
+                                                                                      .Select(v => SemanticVersion.Parse(v))
+                                                                                      .Max()
+                                                                                      .ToString()
                 : group.Version;
 
             if (!string.IsNullOrEmpty(first))
--- a/src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
+++ b/src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
@@ -68,12 +68,37 @@ namespace Microsoft.Web.LibraryManager.Providers.jsDelivr
                                                                                                            cancellationToken).ConfigureAwait(false);
 
                 var packageObject = (JObject)JsonConvert.DeserializeObject(latestVersionContent);
-                if (packageObject != null)
+
+                if (includePreReleases)
                 {
-                    var versions = packageObject["tags"] as JObject;
-                    var versionValue = versions["latest"] as JValue;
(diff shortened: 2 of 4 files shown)


src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
```diff
@@ -194,7 +194,10 @@ public async Task<string> GetLatestVersion(string libraryName, bool includePreRe
             }
 
             string first = includePreReleases
-                ? (await GetLibraryVersionsAsync(group.DisplayName, cancellationToken).ConfigureAwait(false)).First()
+                ? (await GetLibraryVersionsAsync(group.DisplayName, cancellationToken).ConfigureAwait(false))
+                                                                                      .Select(v => SemanticVersion.Parse(v))
+                                                                                      .Max()
```
tlmii, line 199: Since this wasn't protected before, I assume there's no concern about the results being empty? (I think Max on an empty set throws).

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
```diff
@@ -68,12 +68,37 @@ public async Task<string> GetLatestVersion(string libraryId, bool includePreRele
...
+                        latestVersion = latestTag.Value as string;
+                    }
+                    else if (packageObject["versions"] is JArray versions
+                             && versions.Values() is IEnumerable<JToken> versionValues)
+                    {
+                        latestVersion = versionValues.Select(vv => SemanticVersion.Parse(vv.ToString()))
+                                                     .Where(semanticVersion => semanticVersion.PrereleaseVersion == null)
+                                                     .Max()
+                                                     .ToString();
+                    }
```
tlmii, line 99: Was this `else if` part added because the `latest` tag isn't reliable either (similar to your comment above about prerelease? Might be worth a comment for why the fallback is needed.

tlmii (approved):

src/LibraryManager/Providers/Cdnjs/CdnjsCatalog.cs
jimmylewis: Since Cdnjs is a curated catalog, I admit we're making an assumption here that they'll always have at least one version in the catalog.  But I think that's a safe assumption.

src/LibraryManager/Providers/jsDelivr/JsDelivrCatalog.cs
jimmylewis: It was primarily added because the GitHub versions doesn't include a "latest" tag (actually, it doesn't include any tags from my poking about).  I'm not sure whether it's also possible that they would have an NPM package that didn't have "latest" since there isn't consistency in the pre-release tags, so catching that is an extra bonus, but it's mostly driven to support the GH versions.

PR #607, opened 2020-09-17
Rename folders to fix namespace mismatch

RobJohnston:
This is a continuation of PR https://github.com/aspnet/LibraryManager/pull/551

In that PR, the following namespaces were not created because a class exists with the same name as its folder:

* Microsoft.Web.LibraryManager.CacheService
* Microsoft.Web.LibraryManager.SemanticVersion
* Microsoft.Web.LibraryManager.Vsix.ErrorList

This PR renames the "CacheService" directory to simply "Cache", removes the "SemanticVersion" folder (since it only contains one file), and renames the 'ErrorList" directory to "Errors".

Partially addresses #107.


RobJohnston:
I figured since I started working on #107, I might as well continue to get it closed.  I'm completely open to suggestions.

src/LibraryManager.Vsix/Errors/ErrorList.cs
```diff
@@ -8,7 +8,7 @@
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.Web.LibraryManager.LibraryNaming;
 
-namespace Microsoft.Web.LibraryManager.Vsix
+namespace Microsoft.Web.LibraryManager.Vsix.Errors
 {
     internal class ErrorList
```
jimmylewis, line 13: I'm inclined to go with the `ErrorList` namespace to match the original folder name as this is specifically for integrating into the Error List feature in Visual Studio.  I would actually prefer to rename the class here, maybe something like `ErrorListPropagator` or `ErrorListManager`.

jimmylewis (changes requested):

src/LibraryManager.Vsix/Errors/ErrorList.cs
RobJohnston: Done using `ErrorListPropagator`.

A bit of a nightmare dealing with the changes that VS wants to make to the VSIX project file... removing blank lines, reformatting, and inserting a strange `&#xD;&#xA;` character.

jimmylewis (approved):

PR #608, opened 2020-09-20
fix: provide error message for unknown provider.

### LINKED ISSUE #134: [CLI] "libman cache clean <provider>" command doesn't produce an error message for unknown provider id
### Functional impact
*Customer might not realize command failed since no error message is produced, e.g. if the customer misstypes the provider id*  

### Minimal repro steps

1. Execute 

```
libman cache clean foo
```

### Expected result
*Error message stating provider "foo" is not installed*  

### Actual result
*No output*  

### Further technical details
*We can get a list of the available providers, and we do give "invalid provider id" error in libman install*

RobJohnston:
If a provider is given and it is not one of the available providers, output an invalid operation message like "Provider 'foo' is not installed.

Addresses #134

### DIFF
--- a/src/libman/Commands/CacheCleanCommand.cs
+++ b/src/libman/Commands/CacheCleanCommand.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Web.LibraryManager.Contracts;
@@ -62,6 +63,13 @@ namespace Microsoft.Web.LibraryManager.Tools.Commands
                 catch (Exception ex)
                 {
                     Logger.Log(string.Format(Resources.Text.CacheCleanFailed, ex.Message), LogLevel.Error);
+                }
+            }
+            else
+            {
+                if (!ManifestDependencies.Providers.Any(p => p.Id == Provider.Value))
+                {
+                    throw new InvalidOperationException(string.Format(Resources.Text.ProviderNotInstalled, Provider.Value));
                 }
             }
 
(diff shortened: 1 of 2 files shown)


RobJohnston:
I'm thinking of adding a test like this:

```csharp
[TestMethod]
public void TestCacheClean_ThrowsIfUnknownProvider()
{
    var cleanCommand = new CacheCleanCommand(HostEnvironment);
    cleanCommand.Configure();

    try
    {
        cleanCommand.Execute("foo");
    }
    catch (AggregateException age)
    {
        Assert.AreEqual(1, age.InnerExceptions.Count);
        Assert.AreEqual(typeof(InvalidOperationException), age.InnerExceptions[0].GetType());

        return;
    }

    Assert.Fail();
}
```

Necessary?  Desired?  Correct?

src/libman/Commands/CacheCleanCommand.cs
```diff
@@ -64,6 +65,13 @@ protected override Task<int> ExecuteInternalAsync()
                     Logger.Log(string.Format(Resources.Text.CacheCleanFailed, ex.Message), LogLevel.Error);
                 }
             }
+            else
+            {
+                if (!ManifestDependencies.Providers.Any(p => p.Id == Provider.Value))
```
jimmylewis, line 70: This should just be an `else if`.

jimmylewis (changes requested):
Please add the unit test.  However, instead of using the try-catch block, use `var exception = Assert.Throws<TExceptionType>(() => cleanCommand.Execute("foo"))` (is it InvalidOperationException or AggregateException here?)

RobJohnston:
> 
> 
> Please add the unit test. However, instead of using the try-catch block, use `var exception = Assert.Throws<TExceptionType>(() => cleanCommand.Execute("foo"))` (is it InvalidOperationException or AggregateException here?)

Done.  There is no existing example like this to follow in the unit test project.  Let me know what you think.

src/libman/Commands/CacheCleanCommand.cs
RobJohnston: For future reference, I had chopped this down because there's still a case that doesn't output a message -- if the directory doesn't exist but it _is_ a valid provider.  I couldn't find a suitable message in the existing translations, so leaving this out (this case isn't mentioned in the original issue anyway).

jimmylewis (approved):

PR #629, opened 2021-03-30
Changes for consuming merged interop assemblies in Dev17

jimmylewis:
The most temporary of these is the Nuget.VisualStudio package, which is
on its own feed.  Once the NuGet team can fully publish their changes,
it should be published again on the dotnet-public feed, and we can clean
it up from our nuget.config at that time.


tlmii (approved):
Aside from whatever you decide to do with the test failure (maybe just another PackageVersion/PackageReference to MS.VS.Design.Framework?) the rest looks good.

PR #631, opened 2021-07-01
Updates to support VS2022

jimmylewis:
Summary of the changes:
 - Update VS package dependencies (this includes the interop refactoring work done earlier in Dev17)
 - Converting Nuget package version management to use CPVM
 - Fixup code analysis issues generated from updated packages
 - Fixed issue with context menu items from VSCT file not showing in IDE


src/LibraryManager.Vsix/Json/SuggestedActions/SuggestedActionProvider.cs
```diff
@@ -43,16 +43,22 @@ public IEnumerable<ISuggestedAction> GetSuggestedActions(ITextView textView, ITe
 
             yield return new UninstallSuggestedAction(this, LibraryCommandService);
 
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
tlmii, line 46: I assume you're skipping dispose because it is returned... any concern about the case where it is not returned (`!update.HasActionSets`)?

src/LibraryManager/Microsoft.Web.LibraryManager.csproj
```diff
@@ -7,9 +7,9 @@
     <PackageId>Microsoft.Web.LibraryManager</PackageId>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Minimatch" />
+    <PackageReference Include="Minimatch" NoWarn="NU1603" />
```
tlmii, line 10: Just curious, what was this for?

src/LibraryManager/Utilities/ParallelUtility.cs
```diff
@@ -34,19 +34,21 @@ public static async Task ForEachAsync<T>(Func<T, Task> action, int degreeOfParal
             items = items ?? throw new ArgumentNullException(nameof(items));
             cancellationToken.ThrowIfCancellationRequested();
 
-            var semaphore = new SemaphoreSlim(degreeOfParallelism, degreeOfParallelism);
-            var allTasks = new List<Task>();
-
-            foreach (T item in items)
+            using (var semaphore = new SemaphoreSlim(degreeOfParallelism, degreeOfParallelism))
```
tlmii, line 37: Not that it matters, much, but to simplify the edit (which I'm assuming was in response to an analyzer warning) could this have been a using declaration?

src/libman/GlobalSuppressions.cs
```diff
@@ -4,10 +4,16 @@
...
-[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Globalization", "CA1307:Specify StringComparison", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.Extensions.CommandLineUtils.CommandOption.#ctor(System.String,Microsoft.Extensions.CommandLineUtils.CommandOptionType)")]
-[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Globalization", "CA1307:Specify StringComparison", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.Extensions.CommandLineUtils.AnsiConsole.WriteLine(System.String)")]
-[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Globalization", "CA1307:Specify StringComparison", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(System.String[])~System.Int32")]
+[assembly: SuppressMessage("Design", "CA1064:Exceptions should be public", Justification = "<Pending>", Scope = "type", Target = "~T:Microsoft.Extensions.CommandLineUtils.CommandParsingException")]
```
tlmii, line 12: The `<Pending>` was in the entries before this change, so just I'm assuming you're fine with it, just making sure since it seems like the type of thing that would nag at you :)

tlmii (approved):

src/LibraryManager/Microsoft.Web.LibraryManager.csproj
jimmylewis: This assembly targets netstandard20, but the package has a dependency on some netstandard 1.x prerelease version.  It makes the restore graph super messy and gives a ton of these warnings.  I also ended up suppressing it globally above, with a comment to the issue in that repo (which is vague on details, but I don't know if this will ever get fixed...)

src/libman/GlobalSuppressions.cs
jimmylewis: Yeah, all of these are under the comment for external sources (from the Microsoft.Extensions.CommandLineUtils.Sources package)

src/LibraryManager.Vsix/Json/SuggestedActions/SuggestedActionProvider.cs
jimmylewis: Since the Dispose implementation doesn't do anything, I'm not concerned.  Yes, we are breaching the convention, but there's no effect.  If something were to change here where Disposing were important, that could be bad, but at this point it's more likely for this all to get rewritten into an LSP-based extension (so we would no longer be inheriting IDisposable from the VS Editor's ISuggestedAction) before any other major changes would happen.

src/LibraryManager/Utilities/ParallelUtility.cs
jimmylewis: It would simplify it but then I'd have to bump the compiler version.  Will do that soon (now that I've noticed haha), but when I do I want to make all the appropriate style changes at once for consistency.

PR #671, opened 2022-02-13
Generate and include SBOM in the .vsman for VS insertions

jimmylewis:


tlmii (approved):
LGTM. Only suggestion I'd say is to maybe add some comments in the .vsmanproj around what the expectation is for the MergeManifest piece (e.g. it copies **and** renames it to a not-specified-here-filename of sbom.json). Only because I can envision this process getting tweaked (again) without much notice and something being broken - being able to trace the expectations might be nice.

PR #674, opened 2022-03-25
Correct the aligment of the completion flyout

jimmylewis:
Not sure why the old offset was calculated like that, it's as though the flyout had used `Placement="Right"` in XAML instead of `Placement="Left"`.

Resolves https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1468930

### DIFF
--- a/src/LibraryManager.Vsix/UI/Controls/TextBoxWithSearch.xaml.cs
+++ b/src/LibraryManager.Vsix/UI/Controls/TextBoxWithSearch.xaml.cs
@@ -31,7 +31,6 @@ namespace Microsoft.Web.LibraryManager.Vsix.UI.Controls
         {
             InitializeComponent();
 
-            Loaded += HandleLoaded;
             DataContextChanged += HandleDataContextChanged;
         }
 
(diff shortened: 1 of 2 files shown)


src/LibraryManager.Vsix/UI/Controls/TextBoxWithSearch.xaml.cs
```diff
@@ -216,10 +216,11 @@ private void OnLostFocus(object sender, RoutedEventArgs e)
 
         private void PositionCompletionPopup()
         {
-            Flyout.VerticalOffset = SearchTextBox.ActualHeight;
-            Flyout.HorizontalOffset = -SearchTextBox.ActualWidth;
             Options.Measure(new Size(double.PositiveInfinity, double.PositiveInfinity));
             Flyout.Width = Options.DesiredSize.Width;
+            // since it's aligned with the left edge, move it by its own width to make it appear below
```
tlmii, line 221: Not sure I follow what's going on here. How does moving it by its own width make it appear below something? Or did you mean *and* make it appear below (since you're also doing the vertical offset)? Is Flyout.Width also the width of the search box (or whatever it is this is being positioned around)?

tlmii: OK. I _think_ I understand what's going on here... Flyout.Placement is left, which aligns the right edge with the left edge of the search box, and the top with the top. So you're moving it over and down to get it below the search box. Does Placement="Bottom" not work for some weird reason that the docs don't point out?

jimmylewis: Ok, so I swear when I did #650 that I tried Placement="Bottom" first and it didn't work.  But now it works great for me.  I'll use that instead.  😵‍💫

tlmii (approved):

PR #678, opened 2022-03-27
Create a package for the VS integration tests

jimmylewis:
This creates a package to be used for VS integration test automation runs.  It's a nonstandard package and only intended for use with the VS automation runs.

This package always uses a version that doesn't contain the prerelease data.  This is only because I couldn't find a way to force the nbgv CLI to respect the PublicRelease environment variable, so the NugetPackageVersion it generates is always prerelease.  This didn't match what was generated by CreateInsertionMetadata.proj which does respect the environment variable.

I also had to update some of the VS dependencies to get the tests running again.


jimmylewis:
Closing for now as this is really stale, and I don't know if it's worth it.

Issue #685, opened 2022-06-20
Libraries could not be resolved by the "cdnjs" provider


duykhiem:
### Describe the bug
We are having an issue with the "cdnjs" provider when building our project in **both Hanoi, Vietnam, and Grimstad, Norway**.
It is happening with **both local build and Azure DevOps Pipeline build**.
**Failed on some random libraries** like jquery@3.4.1, tabulator@4.4.1, tempusdominus-bootstrap-4@5.1.2...

**Error**:

```

Restore operation started
##[error]libman.json(0,0): Error LIB002: The "jquery@3.4.1" library could not be resolved by the "cdnjs" provider
     1>libman.json : error LIB002: The "jquery@3.4.1" library could not be resolved by the "cdnjs" provider 
         One or more libraries failed to restore
         
Build FAILED.

       (LibraryManagerRestore target) -> 
         **libman.json : error LIB002: The "jquery@3.4.1" library could not be resolved by the "cdnjs" provider** 
```

### To Reproduce
1. Using this version of Visual Studio 2022
2. Build the solution
4. See error:
![image](https://user-images.githubusercontent.com/29139700/174572903-9bcba493-21fe-4517-91d7-e63027800370.png)

### Expected behavior
Build and restore client libraries successfully 

### Screenshots
![image](https://user-images.githubusercontent.com/29139700/174573052-b307f452-966d-4f97-ac75-5d1517282fbd.png)
![image](https://user-images.githubusercontent.com/29139700/174573611-dc192c79-c2e4-45e0-acb8-15c591ac2d63.png)
![image](https://user-images.githubusercontent.com/29139700/174573819-d3bb69fe-e8b4-4897-b381-4a1331a27080.png)

### Additional context
- We are using cdnjs as the default provider in [LibMan](https://docs.microsoft.com/en-us/aspnet/core/client-side/libman/libman-vs?view=aspnetcore-6.0)
- It was working perfectly before but **suddenly failed today without any changes related to Libman or package references**

LibMan CLI version: 2.1.161+abc97ecc7d.RR
Visual Studio 2022
Microsoft.Web.LibraryManager.Build" version="2.1.161" targetFramework="net472"

duykhiem:
The workaround is to clear Libman cache
https://github.com/cdnjs/cdnjs/issues/14141#issuecomment-1160220546

stebueh:
[...]

PR #687, opened 2022-06-21
Adapt to Cdnjs API changes

jimmylewis:
The Cdnjs API is changing to stop including the assets for all versions of a library in the /libraries/:library API.  We were using this to determine both library versions and library files.

The changes are fairly straightforward:
- for versions, we can just use the "versions" property from the library.  It's probably easier than what we were doing before (parsing each version from an "assets" entry).
- for files, we need to invoke the /libraries/:library/:version API instead.  However, the default file (which we use for selecting the default checked box in the wizard) is on the library group, not the library-version, so we still need to get that from the first request.
- Lastly, I included fields on each API call, so that we're not downloading any data that we aren't intending to use.  This drops 10s of KB from most library group queries, since there's a lot of metadata we don't care about.  We can always add more back later.

Addresses #685 but this will also need to be backported for servicing VS2017 and VS2019.


tlmii (approved):
LGTM

PR #688, opened 2022-06-21
Port build updates to VS2019 servicing branch

jimmylewis:
This ports several changes to the build pipeline related to MicroBuild updates.  These changes are necessary to produce a build on the build pools compliant with MS policy.

Unit tests fail, due to #685, I will port that fix separately after the build pipeline is updated.  I've tested locally that the unit tests pass with these changes (which don't really affect local build) and the ported fix.


tlmii (commented):
It looks like the build failures are due to the .NET versions available (because it wants 2.1)?

jimmylewis:
I've made some fixes to the tests, now it seems the only ones failing are actually due to the expected issues with cdnjs.

PR #698, opened 2022-09-06
Switch to package refs for webtools json references

ToddGrun:
Shouldn't be using checked in binaries, and will allow us to potentially change the referenced APIs easier.


ToddGrun:
Getting a build error on the ci as the 2.4.35 MessagePack package isn't found. It doesn't look like vs-impl (or any of the other feeds consumed by this repo) publish it?

jimmylewis:
@mmitche how can we update these dependencies on the dotnet-public feed as configured by #615?

ToddGrun:
@ToddGrun is added to the review. #Closed

jimmylewis (approved):

PR #701, opened 2022-09-27
Convert ruleset files to globalconfig

jimmylewis:
Ruleset files are deprecated, being replaced by either .editorconfig or .globalconfig.  Globalconfig has the benefit of being applied to all sources within the project, whereas editorconfig only applies to files within that folder (the difference being if the project includes linked sources, only globalconfig will apply).


ToddGrun (approved):
:shipit:

PR #703, opened 2022-10-09
Read lines from file once when assigning lines to error items

### LINKED ISSUE #137: Pressing CTRL+S twice (in quick succession) on libman.json in VS causes Save As dialog to show up.
### Minimal repro steps
1. Open a libman.json file in VS
2. Press CTRL+S twice in quick succession.

### Expected result
File is saved. No additional actions taken.

### Actual result
Visual studio opens a `Save as` dialog for the libman.json file.

jimmylewis:
File.ReadLines is implemented with an iterator that opens a stream to the file.  We were iterating it several times (calling `.Count()` and `.Element(i)` would clone the iterator, which meant constantly reiterating the underlying file stream), and keeping it the iterator in scope kept the file handle open.  This caused repeated file saves in VS to be unable to execute (because the file was locked), so VS would pop up a Save As dialog.

Iterating the lines once and then realizing the result into an array allows us to more efficiently compute the results using the string array rather than relying of repetitive file IO and keeping the file handle open.

Resolves #137
Also reported internally as https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1576158

### DIFF
--- a/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
+++ b/src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
@@ -100,7 +100,9 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
 
             if (e.FileActionType == FileActionTypes.ContentSavedToDisk && textDocument != null)
             {
-                _ = Task.Run(async () =>
+                _ = Task.Run(async () => await DoRestoreOnSaveAsync());
+
+                async Task DoRestoreOnSaveAsync()
                 {
                     try
                     {
@@ -151,7 +153,7 @@ namespace Microsoft.Web.LibraryManager.Vsix.Json
                         Logger.LogEvent(ex.ToString(), LogLevel.Error);
                         Telemetry.TrackException("RestoreOnSaveFailed", ex);
                     }
-                });
+                };
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/LibraryManager.Vsix/Json/TextviewCreationListener.cs
```diff
@@ -100,7 +100,9 @@ private void OnFileSaved(object sender, TextDocumentFileActionEventArgs e)
 
             if (e.FileActionType == FileActionTypes.ContentSavedToDisk && textDocument != null)
             {
-                _ = Task.Run(async () =>
+                _ = Task.Run(async () => await DoRestoreOnSaveAsync());
```
ToddGrun, line 103: >_ = Task.Run(async () => await DoRestoreOnSaveAsync()); [](http://example.com/codeflow?start=16&length=55)

It's late, but what's the functional difference here?

jimmylewis: Functionally no difference, I just converted it to a local function while I was debugging and overall prefer a local function over lengthy anonymous delegates.  It does make the stack trace more identifiable.  I could revert this but felt it was a slight overall improvement to the code.

ToddGrun: I have no problem with it, it was late, I was a *couple* beers in and just couldn't see why it was done.

ToddGrun (approved):
:shipit:

PR #706, opened 2022-11-17
Fix warnings related to NuGet packages

jimmylewis:
Summary of the changes:
 - Sets nuget package source mappings
 - Removes duplicate PackageReference in integration test project.
 - Change to use NugetAuthenticate and dotnet restore instead of DotNetCoreCLI@2 task (the DotNetCoreCLI task breaks package source mappings)


tlmii (approved):
Seems problematic that the DotnEtCoreCLI restore task causes issues that you have to work around. You'd think they'd want that to work... But LGTM

PR #715, opened 2023-09-11
Update VS references

jimmylewis:
- Updated to latest (17.7 GA) packages from VS
 - Addressed some warnings, old and new
   - VSTHRD analyzer complaining about unawaited tasks => explicitly discard to indicate fire-and-forget
   - OutputWindowPane.OutputString is deprecated => switch to OutputWindowTextWriter, which allows free-threaded writes to an output pane
   - Couple integration tests were failing by calling removed APIs => updated to latest (17.7) and tests now pass
   - Pinned a few package versions to resolve MSB3277 warnings
   - VS JSON parser's ArrayElementNode.Elements marked deprecated => move to ElementNodes.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,1 +1,1 @@
-msbuild /r %~dp0LibraryManager.sln /clp:Verbosity=Minimal;Summary;ForceNoAlign %*
+msbuild /r %~dp0LibraryManager.sln /clp:Verbosity=Minimal;Summary;ForceNoAlign /bl:%~dp0artifacts/Build.binlog %*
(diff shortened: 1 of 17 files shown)


build.cmd
```diff
@@ -1 +1 @@
-msbuild /r %~dp0LibraryManager.sln /clp:Verbosity=Minimal;Summary;ForceNoAlign %*
+msbuild /r %~dp0LibraryManager.sln /clp:Verbosity=Minimal;Summary;ForceNoAlign /bl:%~dp0artifacts/Build.binlog %*
```
alexgav, line 1: How much time does capturing binlog add? Do we want to do this on every build, even if everything succeeded?

alexgav (approved):
Minor concern about adding binlog to every build from perf perspective, other than that LGTM.

build.cmd
jimmylewis: It's a negligible overhead, much less than having the console output turned up.  Yes, it's nice to collect it on good builds, I do that a lot when diffing a build break (before vs. after), or validating that specific steps were carried out correctly (signing, loc, etc).

PR #716, opened 2023-09-11
Switch away from deprecated NuGet API

jimmylewis:
The IVsPackageInstallerServices interface is marked deprecated for getting installed package info, due to threading problems within NuGet.  Switched to the new interface instead.

### DIFF
--- a/src/LibraryManager.Vsix/Shared/VsHelpers.cs
+++ b/src/LibraryManager.Vsix/Shared/VsHelpers.cs
@@ -545,5 +545,17 @@ namespace Microsoft.Web.LibraryManager.Vsix.Shared
                 }
             }
         }
+
+        public static Guid GetProjectGuid(Project project)
+        {
+            string uniqueName = project.UniqueName;
+            IVsSolution solution = (IVsSolution)Package.GetGlobalService(typeof(SVsSolution));
+            solution.GetProjectOfUniqueName(uniqueName, out IVsHierarchy hierarchy);
+            hierarchy.GetGuidProperty(
+                (uint)VSConstants.VSITEMID.Root,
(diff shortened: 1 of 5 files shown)


src/LibraryManager.Vsix/Shared/VsHelpers.cs
```diff
@@ -545,5 +545,17 @@ private static void DeleteEmptyFolders(HashSet<ProjectItem> folders)
                 }
             }
         }
+
+        public static Guid GetProjectGuid(Project project)
+        {
+            string uniqueName = project.UniqueName;
```
alexgav, line 551: Should we do some sort of return result checking in any of these? I'd assume there are HRs returned and object could be checked for null.

alexgav (approved):

jimmylewis:
While doing some further testing with this, it seems really flaky.  Now that the work of detecting package installation is async, it seems that the OleMenuCommand is not waiting for it to finish - it will use stale text for the menu command.  I'm seeing this in the debugger where the context menu appears between stepping through these lines:
```
_isPackageInstalled = await IsPackageInstalledAsync(projectGuid);
// context menu is shown before the following lines are executed:
if (_isPackageInstalled)
{
    button.Text = Resources.Text.DisableRestoreOnBuild;
}
else
{
    button.Text = Resources.Text.EnableRestoreOnBuild;
}
```

I tried caching the text to use, thinking I could make up for it, but that revealed that the NuGet API is actually pretty laggy.  After executing an install/uninstall operation, it takes several seconds for it to return an updated result (which different from the old implementation, which seemed to update immediately).  I'll need to follow up with someone from NuGet to understand why this API seems much flakier.

PR #725, opened 2024-02-15
#682 Add non-zero exit code on general exception and on error in Restore command

### LINKED ISSUE #682: LibMan CLI returns zero exit code even when there was an error
### Describe the bug
Even if LibMan CLI operation has finished with errors, exit code of the process is 0. This is a little inconvenient for build scripts or Dockerfiles when you don't know that some libraries haven't been restored, for example, because of library CDN provider's temporary unavailability or bugs. The only workaround is to check stderr and fail when there is any output, but it seems that returning non-zero exit code would be a more clear indication that something went wrong.

### To Reproduce
1. Create temporary Docker container with .NET 5.0 SDK and attach to its terminal:
```
sudo docker run --rm -it mcr.microsoft.com/dotnet/sdk:5.0
```
2. Install LibMan CLI tool
```
export PATH="$PATH:/root/.dotnet/tools"
dotnet tool install -g Microsoft.Web.LibraryManager.Cli
```
3. Perform any illegal operation. For example, try to restore libraries in the directory without `libman.json` file.
```
root@0f7d49cfdfcc:/# libman restore
libman.json was not found:/libman.json
```
4. Check exit code of libman:
```
root@0f7d49cfdfcc:/# echo $?
0
```
5. Exit code is zero despite libman command failure

### Expected behavior
Exit code of libman is non-zero

### Additional context
LibMan CLI version: 2.1.161+abc97ecc7d.RR

sprytnyserek:
Summary of the changes
 - Used return value from Microsoft.Extensions.CommandLineUtils.Execute() to propagate exit code, was hardcoded to 0
 - Eliminated unclear use of IEnumerable variables with multiple enumerations when evaluating 'restore' results

Resolves #682

### DIFF
--- a/src/LibraryManager.Contracts/ExitCode.cs
+++ b/src/LibraryManager.Contracts/ExitCode.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Web.LibraryManager.Contracts
+{
+    /// <summary>
+    /// Exit codes of libman tool
+    /// </summary>
+    public enum ExitCode
+    {
+        /// <summary>
+        /// Indicates successful execution
+        /// </summary>
+        Success = 0,
+
+        /// <summary>
+        /// Indicates failed execution
+        /// </summary>
+        Failure = 1
+    }
+}
--- a/src/LibraryManager/Manifest.cs
+++ b/src/LibraryManager/Manifest.cs
@@ -308,7 +308,7 @@ namespace Microsoft.Web.LibraryManager
         /// Restores all libraries in the <see cref="Libraries"/> collection.
         /// </summary>
         /// <param name="cancellationToken">A token that allows for cancellation of the operation.</param>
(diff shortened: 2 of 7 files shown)


sprytnyserek:
@dotnet-policy-service agree

src/LibraryManager.Contracts/ExitCode.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Web.LibraryManager.Contracts
+{
+    /// <summary>
+    /// Exit codes of libman tool
+    /// </summary>
+    public enum ExitCode
```
jimmylewis, line 9: This needn't be in the Contracts library.  Contracts is for establishing the interfaces that are common between each different host (MSBuild task, VSIX, CLI tool).  Since this is specific to the CLI tool, it should be in the libman project.

jimmylewis (changes requested):

jimmylewis (approved):

alexgav (approved):
:shipit:

PR #729, opened 2024-03-07
Update Library Manager build for net8

phil-allen-msft:
Update Library Manager build for net8

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -19,6 +19,8 @@
     <PackageOutputPath>$(RepoRoot)\artifacts\$(Configuration)</PackageOutputPath>
     <HighEntropyVa>true</HighEntropyVa>
     <CheckForOverflowUnderflow Condition=" '$(Configuration)' == 'Debug' ">true</CheckForOverflowUnderflow>
+    <NetFxTFM>net481</NetFxTFM>
+    <NetCoreTFM>net8</NetCoreTFM>
   </PropertyGroup>
 
   <ItemGroup>
@@ -27,7 +29,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" PrivateAssets="All" />
     <PackageReference Include="Microsoft.VisualStudio.Internal.MicroBuild" PrivateAssets="All" />
     <PackageReference Include="NerdBank.GitVersioning" PrivateAssets="All" />
   </ItemGroup>
--- a/build/CreateInsertionMetadata.proj
+++ b/build/CreateInsertionMetadata.proj
@@ -1,7 +1,7 @@
-<Project Sdk="Microsoft.Build.NoTargets/1.0.94" DefaultTargets="GenerateInsertionParameters">
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="GenerateInsertionParameters">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>$(NetCoreTFM)</TargetFramework>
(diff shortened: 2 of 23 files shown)


build/CreateInsertionMetadata.proj
```diff
@@ -1,7 +1,7 @@
-<Project Sdk="Microsoft.Build.NoTargets/1.0.94" DefaultTargets="GenerateInsertionParameters">
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="GenerateInsertionParameters">
```
jimmylewis, line 1: Why make this change?  This project didn't use any of the standard build tasks, so the NoTargets SDK was chosen specifically.

Directory.Build.props
```diff
@@ -19,6 +19,8 @@
     <PackageOutputPath>$(RepoRoot)\artifacts\$(Configuration)</PackageOutputPath>
     <HighEntropyVa>true</HighEntropyVa>
     <CheckForOverflowUnderflow Condition=" '$(Configuration)' == 'Debug' ">true</CheckForOverflowUnderflow>
+    <NetFxTFM>net481</NetFxTFM>
+    <NetCoreTFM>net8</NetCoreTFM>
```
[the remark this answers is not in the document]
jimmylewis: Possibly also applicable to net481.  It's not fair to assume this tool is only used by developers on the latest VS cadence (except for the VSIX), is there an impetus to force adopting the latest versions?

src/LibraryManager.Build/RestoreTask.cs
```diff
@@ -16,7 +16,7 @@
 namespace Microsoft.Web.LibraryManager.Build
 {
     public class RestoreTask
-#if NET472
+#if NET481
```
jimmylewis, line 19: `#if NETFRAMEWORK` would be applicable on any legacy version.

src/libman/libman.csproj
```diff
@@ -2,16 +2,15 @@
...
     <RootNamespace>Microsoft.Web.LibraryManager.Tools</RootNamespace>
     <PackAsTool>true</PackAsTool>
-    <PackAsToolShimRuntimeIdentifiers>win-x64;win-x86;osx-x64</PackAsToolShimRuntimeIdentifiers>
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <PackageId>Microsoft.Web.LibraryManager.Cli</PackageId>
     <Description>Command line tool for Library Manager</Description>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="System.Runtime" />
+    <Reference Include="System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```
jimmylewis, line 13: how does this version get resolved in the future?  I.e. when updating to dotnet 10, will this break or does it get re-resolved to the version=10.0.0.0 assembly?

jimmylewis (approved):
overall, LGTM.  Mostly concerned about communicating the updated platform requirements in a semantic way.

alexgav (approved):
:shipit:

build/CreateInsertionMetadata.proj
jimmylewis: "In the newest builds" doesn't make sense.  It's a [NuGet package](https://www.nuget.org/packages/Microsoft.Build.NoTargets), not part of the .NET SDK.  Can you share the error you were seeing?  I tried building it from your branch but with this change reverted and it seemed to build fine.

phil-allen-msft: After running commands like CI would, the existing SDK reference works great.

src/LibraryManager.Build/RestoreTask.cs
phil-allen-msft: Updated to just that.

PR #730, opened 2024-03-21
Convert official pipeline to1ES template

phil-allen-msft:
Convert official pipeline to1ES template

### DIFF
--- a/azure-pipelines/azure-pipeline.microbuild.before.yml
+++ b/azure-pipelines/azure-pipeline.microbuild.before.yml
@@ -1,10 +1,1 @@
-steps:
-
-- task: MicroBuildSigningPlugin@3
-  inputs:
-    signType: $(SignType)
-    zipSources: false
-  displayName: Install MicroBuild Signing Plugin
-
-- task: MicroBuildSwixPlugin@4
-  displayName: Install MicroBuild Swix Plugin
+steps: []
--- a/azure-pipelines/official.yml
+++ b/azure-pipelines/official.yml
@@ -11,8 +11,65 @@ variables:
   Packaging.EnableSBOMSigning: true
   Codeql.Enabled: true
 
-jobs:
-  - job: Windows
-    pool: VSEngSS-MicroBuild2022-1ES
-    steps:
-      - template: build.yml
+resources:
+  repositories:
+  - repository: MicroBuildTemplate
+    type: git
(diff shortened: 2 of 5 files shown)


azure-pipelines/azure-pipeline.microbuild.before.yml
```diff
@@ -1,10 +1 @@
-steps:
-
-- task: MicroBuildSigningPlugin@3
-  inputs:
-    signType: $(SignType)
-    zipSources: false
-  displayName: Install MicroBuild Signing Plugin
-
-- task: MicroBuildSwixPlugin@4
-  displayName: Install MicroBuild Swix Plugin
```
alexgav, line 10: I didn't see where this went - I'm guessing parent template we are extending?

alexgav (approved):
:shipit:

azure-pipelines/azure-pipeline.microbuild.before.yml
tlmii: It ended up here:

https://github.com/aspnet/LibraryManager/blob/77d22d5bcebe4b2773577a4076adc5ba01a2f4a3/azure-pipelines/official.yml#L38-L44

azure-pipelines/azure-pipeline.microbuild.before.yml
```diff
@@ -1,10 +1 @@
...
+steps: []
```
tlmii, line 1: Should we just delete this file (and the reference to it)? I guess no harm in leaving it if its needed in the future. Just seems odd having an (essentially) empty file here.

azure-pipelines/official.yml
```diff
@@ -11,8 +11,65 @@ variables:
...
+          mb:
+            signing:
+              enabled: true
+              signType: $(SignType)
+              zipSources: false
+            swix:
+              enabled: true
+          outputParentDirectory: $(Build.ArtifactStagingDirectory)
+          outputs:
+          - ${{ if eq(variables['system.collectionId'], '011b8bdf-6d56-4f87-be0d-0092136884d9') }}:
```
tlmii, line 47: Will this ever not be true for this yaml (I assume official only runs in the internal pipeline)?

tlmii (approved):
Some comments, but nothing blocking. Looks good!

Issue #728, opened 2024-03-04
Unable to build the project locally


MatthewSteeples:
### Describe the bug
Doesn't look like those of us outside of Microsoft can access the Azure Artefacts feed that this project depends on.

### To Reproduce
1. Check out the project
2. Run `dotnet restore`

### Expected behavior
Project restores dependencies successfully

### Screenshots
```
error : Failed to download package 'Microsoft.NETCore.App.Hos
t.osx-x64.6.0.27' from 'https://pkgs.dev.azure.com/azure-public/0b6653b2-b704-470b-94ed-d0e2e194a663/_packaging/5d2fa05
1-5b55-432e-849e-a04a0d44a500/nuget/v3/flat2/microsoft.netcore.app.host.osx-x64/6.0.27/microsoft.netcore.app.host.osx-x
64.6.0.27.nupkg'
```

### Additional context
Visual Studio 17.9.2
Dotnet SDK 8.0.200

jimmylewis:
I'm currently able to restore cleanly from a non-Microsoft computer.  I think it's a point in time issue: with #729 the feed was populated with all the current packages.  However, a future .NET SDK servicing release could cause one of the runtime packages (such as microsoft.netcore.app.host.osx-x64 mentioned above) to be bumped to a newer version; that then requires that an authorized user push that package into the libman feed (i.e. the point in time can and will come up again).

I tried going back to before we had the consolidated feed in #731.  It failed because some of the internal packages are no longer available on the msft_consumption feed.  It does, however, benefit from having the dotnet-public feed in the list, which would be populated with any newly serviced packages in the future.

So, one fix would be to do something like https://github.com/aspnet/LibraryManager/pull/731/commits/44c5cb32c2449a81df7ab2b2e4b03ca141eca2d6 where it uses the dotnet-public feed for runtime packages, and uses the libman feed for vs-internal packages (the dotnet-myget-legacy can also probably be dropped if using the libman feed).  Maybe revive the PR if that seems like an appropriate way to address the problem...

There's still a potential issue that any external contribution that wants to adopt a new package not present in the dotnet-public feed would be unable to have a successful PR builds without some intervention - without authorization to push new packages into the libman feed, the CI build would fail.

As a workaround, external users can add nuget.org locally to be able to restore any public packages.  However, I know the policies for MSFT CI builds is to use authenticated feeds as much as possible.

/cc @phil-allen-msft for all this context.

MatthewSteeples:
[...]

PR #731, opened 2024-03-22
Revert "Use consolidated nuget feed"

jimmylewis:
Reverts aspnet/LibraryManager#714

This change caused permissions issues when non-Microsoft users attempt to restore the solution.  I was able to reproduce this on my personal (non-MSFT) device, and the restore worked after reverting the change.

It would also be a blocker for proposing a new package or package version, as users not authorized with the ADO feed would not have rights to promote new packages into the consolidated feed.  The main security issue with multiple feeds is described at https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping.  As long as each package is only requested from a single specific feed, it avoids the risk of a supply chain attack where a package from a one feed is maliciously faked on another feed, so since we previously had packageSource configured for each package, this should not be a concern (also, because we don't use nuget.org, all of these feeds should already be secured).

I doubt that reverting this change will solve the issue of proposing a new package version, as an unauthorized user still won't have permissions to promote packages to the dotnet-public feed either.  But at least this should address the restore problem.

Addresses #728

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -2,6 +2,22 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="libman" value="https://pkgs.dev.azure.com/azure-public/vswebtools/_packaging/libman/nuget/v3/index.json" />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-myget-legacy" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/myget-legacy/nuget/v3/index.json" />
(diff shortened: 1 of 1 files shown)


jimmylewis:
I'm currently able to restore fine with the libman feed, so no need to do this.  Reverting the original change failed with some packages no longer available in the source feeds (for Apex and the Web.Editors packages), but succeeded locally because I still had those in my local cache.

I do worry that a future dotnet servicing update may cause some implicit versions (e.g. for the core SDK packages) to get bumped, and then there isn't a way for an external user to restore without being able to ingest the right versions in the feed.  For example, in #728 the reported failure is in Microsoft.NETCore.App.Host.osx-x64 version 6.0.27; the libman feed contains 6.0.22, and updated versions for 8.0.2 and 8.0.3.  When the sdk servicing bumps that to 8.0.4, what's the recourse?

PR #732, opened 2024-03-26
Add a --whatif option to CLI update command

### LINKED ISSUE #723: Expose latest version via CLI, without updating
I can use `dotnet libman update` to update all packages.

But who does that on a production system? It's important to 1) know what updates are available, 2) research the release notes, and then 3) update (and test).

So like other package managers, it would be nice to be able to simply list the latest versions; perhaps:
```
$ dotnet libman update --list
```

jimmylewis:
This flag will cause the command to print a message indicating what changes would be made, but not commit any changes to the file system. It's useful to plan in advance before making changes.

Resolves #723

### DIFF
--- a/src/libman/Commands/UpdateCommand.cs
+++ b/src/libman/Commands/UpdateCommand.cs
@@ -48,6 +48,11 @@ namespace Microsoft.Web.LibraryManager.Tools.Commands
         /// <remarks>Needs the full library id.</remarks>
         public CommandOption ToVersion { get; private set; }
 
+        /// <summary>
+        /// Option to specify if this should print the operation that would be carried out, but not make changes.
+        /// </summary>
+        public CommandOption WhatIf { get; private set; }
+
         public override BaseCommand Configure(CommandLineApplication parent = null)
         {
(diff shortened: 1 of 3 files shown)


phil-allen-msft:
Do you have a matching example where 'whatif' shows only a small amount of output?  I think I've seen other tools that will (with different command names, like 'dry run') give near identical output to the command if run without the option, albeit not actually making any changes.

phil-allen-msft:
Different question, do you have a non-libman command that has --what-if to show an analogous behavior?

jimmylewis:
Many PowerShell cmdlets use the -WhatIf convention (provided in PS via CmdletBinding), as does the Azure CLI (non-PS), and I'm sure I've seen it elsewhere before.  PS example:
```
PS D:\src> Remove-Item -Recurse c:\windows\system32 -WhatIf
What if: Performing the operation "Remove Directory" on target "C:\windows\system32".
```

Some other tools use --dry-run such as git and npm:
```
PS D:\temp> npm install --dry-run bootstrap

added 2 packages in 398ms
```

PR #739, opened 2024-04-10
Set SDK rollforward to LatestFeature

jimmylewis:
This is needed as updating VS versions often push new feature versions of the SDK (such as 8.0.200 -> 8.0.300-preview, removing 8.0.200) which breaks building locally.


alexgav (approved):
:shipit:

PR #741, opened 2024-04-24
Spacebar on the "Files" node should toggle state

phil-allen-msft:
If the "Files" node is partially checked, the user using spacebar will find nothing happens on spacebar because packageItem.IsChecked is null, so !packageItem.IsChecked is also null. Adjust spacebar behavior to match mouseclick behavior


alexgav (approved):
:shipit:

PR #742, opened 2024-04-24
Look and feel fixes for the "Install Client Libraries" dialog

phil-allen-msft:
Fixes tab order to be left to right, top to bottom Adds a border to the dialog
Reverts all styling for comboboxes to the VS common theming


alexgav (approved):
:shipit:

PR #749, opened 2024-05-02
Drop netstandard for $(NetCoreTFM)

jimmylewis:
Netstandard gives compatibility benefits for libraries to be used against netfx or netcore frameworks.  However, for libman, all of our projects that run on netcore already target that framework directly.

This change modifies the existing multitargeted projects to target netcore directly instead of netstandard.  This will allow future changes to use conditional compilation to leverage newer APIs in netcore that do not exist in the netstandard API surface.

The only breaking change here is that the Build project now uses .NET 8 for the build task.  This means consumers of that package who are not using .NET Framework will need to have a compatible version of the SDK; however, this aligns with the prior change in the CLI tool to require .NET 8 as well.


jimmylewis:
Note that this change led to a bunch of runtime package dependency changes, which are addressed via #747.  Hence I've created this PR as a draft until the other can be merged first.

phil-allen-msft (approved):

PR #750, opened 2024-05-10
Followup: Fully remove references

phil-allen-msft:
https://github.com/aspnet/LibraryManager/pull/747 removed some references to Minimatch.dll but left a few behind.

Remove those latent references.


jimmylewis (approved):

jimmylewis:
Doh, I assumed that these would've cause errors, but the glob expands to 0 items so there's no build issue.  Thanks for catching it and cleaning up.

PR #754, opened 2024-05-24
Update package references

phil-allen-msft:
Update a number of packages to be more modern
Stop shipping binaries to run in VisualStudio that VisualStudio already ships (With binding redirects/etc.)
Update the SLN file so that the Vsix project is the default startup project.

### DIFF
--- a/test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
+++ b/test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using Microsoft.Test.Apex.Editor;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NuGet;
 
 namespace Microsoft.Web.LibraryManager.IntegrationTest
 {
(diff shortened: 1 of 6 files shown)


test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using Microsoft.Test.Apex.Editor;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NuGet;
```
alexgav, line 4: >NuGet [](http://example.com/codeflow?start=6&length=5)

Is this needed? Did namespace of the used classes changed?

alexgav (approved):
:shipit:

test/LibraryManager.IntegrationTest/LibmanCompletionTests.cs
phil-allen-msft: It is not (there was an error while I was updating the .vsix project's csproj file). On reopen, I have verified that the namespace for "SemanticVersion" is Microsoft.Web.LibraryManager.  Undone.

Issue #738, opened 2024-04-10
Proposal for more granular file mappings per library


jimmylewis:
# Background
In the past, we've had numerous issues that highlight the limitations of file placement via libman: #407, #597, #702, and various others duped among those 3.

These can be summed up into a few high level issues:
1. The file structure upon installing is fixed based on how it appears within the provider's library, so users cannot exercise high fidelity control of how files are copied into their project (which is one of the primary goals of libman).
2. Users cannot configure a subset of files to one destination and another subset to a different destination.
3. Users want to copy a set of files into their project multiple times.  Libman currently only allows one copy per library ID (e.g. name@version).

# The proposal

This proposal introduces a new property in the manifest that allows more flexible file mappings, addressing all of the above issues.  I've named it "fileMappings":

```
{
  "libraries": [
    {
      "library": "jquery@3.7.1",
      "provider": "jsdelivr",
      "fileMappings": [
        {
          "root": "dist/",
          "destination": "lib/jquery/",
          "files": [
            "*.min.js"
          ]
        }
      ]
    }
  ]
}
```

The `fileMappings` property is an array of mappings.  There may be 0 or more.

Each mapping consists of 3 optional properties:
* `root`: the root of this mapping relative to the library root.  This allows truncating a portion of the relative directory structure, so users can configure their file structure more carefully (addresses issue 1 listed above).  This does not accept glob patterns.
  * If not specified, defaults to the library root (original folder structure is preserved)
* `destination`: where files identified by this mapping should be placed within the project.
  * If not specified, defaults to the library's configured installation path (i.e. respects `destination` on the library or the global `defaultDestination`). If *that* is not specified, it will be a validation error in the manifest.
* `files`: similar to the existing `files` property on the library, this is a list of file patterns (including glob patterns) relative to `root` that are to be included in this mapping.
  * If not specified, defaults to all files within the `root`.

Because there are any number of mappings allowed, users can now configure separate destinations for each set of `files` (addresses issue 2 above), and even configure the same file(s) to as many unique destinations as they desire (addresses issue 3 above).

## Compatibility across libman versions

Because this is a purely additive change to the schema, there is no break to an existing libman.json file.

Pre-existing versions of the tooling will not be able to handle this new property.  As such, a mixed environment could lead to a different state depending on which version of the tool is used.  To mitigate this, the new property will require a new manifest version.  This is to distinguish which versions of libman will support the new property.
[...]

PR #756, opened 2024-05-24
Implement support for "fileMappings"

jimmylewis:
Summary of the changes:
 - Adds support for parsing fileMappings and calculating the goal state from them.
 - Adds support for fileMappings to the schema.  Includes a schema validation error if the schema uses fileMappings but doesn't have a sufficiently high version set.
 - Enables manifest version "3.0" and sets that as the default for newly created manifests.

This implements #738.  There will be an additional PR to make the JSON completion in the editor work for fileMappings.

### DIFF
--- a/src/LibraryManager/Providers/BaseProvider.cs
+++ b/src/LibraryManager/Providers/BaseProvider.cs
@@ -247,48 +247,92 @@ namespace Microsoft.Web.LibraryManager.Providers
 
         private OperationResult<LibraryInstallationGoalState> GenerateGoalState(ILibraryInstallationState desiredState, ILibrary library)
         {
+            var mappings = new List<FileMapping>(desiredState.FileMappings ?? []);
             List<IError> errors = null;
-
-            if (string.IsNullOrEmpty(desiredState.DestinationPath))
+            if (desiredState.Files is { Count: > 0 })
             {
-                return OperationResult<LibraryInstallationGoalState>.FromError(PredefinedErrors.DestinationNotSpecified(desiredState.Name));
+                mappings.Add(new FileMapping { Destination = desiredState.DestinationPath, Files = desiredState.Files });
             }
-
-            IEnumerable<string> outFiles;
-            if (desiredState.Files == null || desiredState.Files.Count == 0)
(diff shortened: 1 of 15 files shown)


test/LibraryManager.Mocks/LibraryInstallationState.cs
```diff
@@ -46,5 +46,7 @@ public class LibraryInstallationState : ILibraryInstallationState
         /// Indicates whether the library is using the default provider.
         /// </summary>
         public bool IsUsingDefaultProvider { get; set; }
+
+        public IReadOnlyList<FileMapping> FileMappings => throw new System.NotImplementedException();
```
phil-allen-msft, line 50: Documentation header?

jimmylewis: made them all inherit from the interface since there was nothing unique to this type.

src/LibraryManager/Providers/BaseProvider.cs
```diff
@@ -247,48 +247,92 @@ public async Task<OperationResult<LibraryInstallationGoalState>> GetInstallation
...
+                // if Root is not specified, assume it's the root of the library
+                string mappingRoot = fileMapping.Root ?? string.Empty;
+                // if Destination is not specified, inherit from the library entry
+                string destination = fileMapping.Destination ?? desiredState.DestinationPath;
+
+                if (destination is null)
+                {
+                    errors ??= [];
+                    string libraryId = LibraryNamingScheme.GetLibraryId(desiredState.Name, desiredState.Version);
+                    errors.Add(PredefinedErrors.DestinationNotSpecified(libraryId));
```
phil-allen-msft, line 275: Is there a case for v1 manifests where (if they followed the schema) they could encounter this error?

jimmylewis: The v1 schema was declared such that either a defaultDestination must be set, or every library must set a destination.  But it got too complicated to be oneOf (defaultDestination, or every library sets a destination,  or (every library without fileMappings sets a destination and every library with fileMappings has a fileMapping destination)).

phil-allen-msft (approved):

PR #759, opened 2024-06-01
Add variable expansion in defaultDestination

### LINKED ISSUE #68: Library Default Destination
Instead of specifying the destination of each library, is it possible to introduce a new "**default destination**" setting that uses conventions at the top level of libman.json.  This setting would still be able to be overwritten at each library level.

For example: 
* **twitter-bootstrap version 3.3.7** library would go to wwwroot/lib/bootstrap
* **jquery version 1.12.0** library would go to wwwroot/lib/jquery/1.12.0
* **jquery version 3.3.1** library would go to wwwroot/lib/jquery/3.3.1

the libman.json would look something like this:

```
{
  "version": "1.0",
  "defaultProvider": "cdnjs",
  "defaultDestination": "wwwroot/lib/[LibraryName]/[Version]",
  "libraries": [
    {
      "library": "twitter-bootstrap@3.3.7",
      "destination": "wwwroot/lib/bootstrap"
    },
    {
      "library": "jquery@1.12.0"
    },
    {
      "library": "jquery@3.3.1"
    }
  ]
}
```

Thanks

jimmylewis:
This resolves #68, except I used [Name] instead of [LibraryName].  It felt like LibraryName should be matched with LibraryVersion and that felt verbose, so I took the shorter versions.

This expansion is applied when we expand the ManifestOnDisk (which is either read from disk or from a raw JSON) into LibraryInstallationState.  This is where we determine to use the defaultDestination or a library-specific destination, so it should be the only place this expansion needs to occur.

### DIFF
--- a/src/LibraryManager/Json/LibraryStateToFileConverter.cs
+++ b/src/LibraryManager/Json/LibraryStateToFileConverter.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
 using Microsoft.Web.LibraryManager.Contracts;
 using Microsoft.Web.LibraryManager.LibraryNaming;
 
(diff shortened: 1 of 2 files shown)


src/LibraryManager/Json/LibraryStateToFileConverter.cs
```diff
@@ -22,22 +27,49 @@ public ILibraryInstallationState ConvertToLibraryInstallationState(LibraryInstal
...
+            {
+                return destination;
+            }
+
+            // if the name contains a slash (either filesystem or scoped packages),
+            // trim that and only take the last segment.
+            int cutIndex = name.LastIndexOfAny(['/', '\\']);
+
+            StringBuilder stringBuilder = new StringBuilder(destination);
+            stringBuilder.Replace("[Name]", cutIndex == -1 ? name : name.Substring(cutIndex + 1));
```
phil-allen-msft, line 68: Possibly a non-real case, but what if name contains [Version]? That would cause the next line to do an extra/unexpected replacement.

jimmylewis: At least as far as NPM is concerned, I don't believe that is a valid name (only allows alphanumeric, underscores, and hyphens).  Since that's the source for the non-filesystem providers (jsdelivr and unpkg directly use NPM, cdnjs is indirectly curated from NPM), it would be an edge case for filesystem only.  I thought about making this logic provider-specific, but that was a lot more complicated than doing it here centrally.

I did briefly consider using a different token that wouldn't be allowed in file paths (implicitly compatible with all providers since they all come down to file names), but on Linux the only character is `/`.  I'm definitely open another delineator if you have a preference, but `[Version]` felt most natural (next on my list was something like `{{Version}}`); it should be a pretty trivial change.

phil-allen-msft (approved):

dunnodepends:
This no longer works for LibMan CLI

PR #760, opened 2024-06-18
Add NOTICE.txt to the repo

phil-allen-msft:
Add NOTICE.txt to the repo

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,0 +1,2654 @@
+NOTICES AND INFORMATION
+Do Not Translate or Localize
+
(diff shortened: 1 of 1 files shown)


jimmylewis:
Since these are all notices for specific dependency versions, what's the lifecycle for maintaining it?  I.e. when we update dependency versions, how does this file stay up to date?

jimmylewis (approved):

phil-allen-msft:
An excellent question.  There's likely a build-centric way to do so, but for now, we're getting started with the right contents here.

PR #765, opened 2024-09-06
Add monthly build

phil-allen-msft:
Verifies the ability to continuously build.

### DIFF
--- a/azure-pipelines/official.yml
+++ b/azure-pipelines/official.yml
@@ -4,6 +4,18 @@ trigger:
   paths:
     exclude: ["*.md"]
 
+schedules:
+  - cron: "0 8 23-29 * 0"
+    displayName: "Monthly build check"
+    branches:
+      include: 
+        - main
+        - rel/*
+      exclude: 
+        - ""
+    always: true # Run even if there have been no source code changes since the last successful scheduled run
+    batch: false # Do not run the pipeline if the previously scheduled run is in-progress
+
 variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
   BuildConfiguration: Release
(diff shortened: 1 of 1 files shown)


azure-pipelines/official.yml
```diff
@@ -4,6 +4,18 @@ trigger:
   paths:
     exclude: ["*.md"]
 
+schedules:
+  - cron: "0 8 23-29 * 0"
+    displayName: "Monthly build check"
+    branches:
+      include: 
+        - main
+        - rel/*
```
jimmylewis, line 13: `rel/*` is a bad idea as it exists today.  We didn't use rel branches regularly, and those that do exist are super outdated in terms of build infrastructure so the builds will almost certainly fail.

I would stick with just main until another branching strategy becomes common.

azure-pipelines/official.yml
```diff
@@ -4,6 +4,18 @@ trigger:
...
+      exclude: 
+        - ""
```
jimmylewis, line 15: What does this accomplish?

azure-pipelines/official.yml
```diff
@@ -4,6 +4,18 @@ trigger:
   paths:
     exclude: ["*.md"]
 
+schedules:
+  - cron: "0 8 23-29 * 0"
```
jimmylewis, line 8: nit: add a comment summarizing the cron syntax for readability so it doesn't require a trip to the docs each time 😉

phil-allen-msft: Added, and tweaked the dates so it is "4th Sunday" consistently

azure-pipelines/official.yml
phil-allen-msft: Denotes the location for the exclude, should we ever need one.  Will remove

jimmylewis (approved):

PR #766, opened 2024-09-06
Update package versions

phil-allen-msft:
Regular update of package versions

### DIFF
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
@@ -16,8 +16,6 @@
     <PackageReference Include="Microsoft.Build.Framework" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.IO.Redist" Condition="'$(TargetFramework)' == '$(NetFxTFM)'" />
-    <PackageReference Include="System.Formats.Asn1" />
     <PackageReference Include="System.Runtime.Loader" Condition="'$(TargetFramework)' == '$(NetCoreTFM)'" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
```diff
@@ -16,6 +16,8 @@
     <PackageReference Include="Microsoft.Build.Framework" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.IO.Redist" Condition="'$(TargetFramework)' == '$(NetFxTFM)'" />
```
jimmylewis, line 19: Are these being added as a means to pin the new versions?  Could we instead set CentralPackageTransitivePinningEnabled in Directory.Packages.props so that transitive references will respect the version specified there, but we're not adding new primary references?  (For example, if the dependency tree changes, adding this as a primary package means we'll always reference it even if it's no longer used.)

phil-allen-msft: (Trying it out)

phil-allen-msft: Double checked and all is well.

src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
```diff
@@ -184,7 +184,7 @@
     <PackageReference Include="Microsoft.WebTools.Languages.Json.VS" />
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="Nuget.VisualStudio" />
-  </ItemGroup>
+   </ItemGroup>
```
jimmylewis, line 187: nit: unnecessary change in this file (and bad indent)

jimmylewis (approved):

src/LibraryManager.Vsix/Microsoft.Web.LibraryManager.Vsix.csproj
phil-allen-msft: Correct and corrected.

PR #773, opened 2024-12-06
Fix build targets

### LINKED ISSUE #770: Microsoft.Web.LibraryManager.Build starting to show security warnings
### Describe the bug
Microsoft.Web.LibraryManager.Build starting to show security warnings 

### To Reproduce
1. Include  `<PackageReference Include="Microsoft.Web.LibraryManager.Build" Version="2.1.175" PrivateAssets="All" />` into the project
2. Build it using .NET SDK 9.0.100 or higher
3. Get   
```
(..) warning NU1903: Package 'System.Private.Uri' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-5f2m-466j-3848
(..) warning NU1902: Package 'System.Private.Uri' 4.3.0 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-x5qj-9vmx-7g6g
(..) warning NU1903: Package 'System.Private.Uri' 4.3.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-xhfc-gr8f-ffwc    
``` 

### Expected behavior
No security warnings

### Additional context
NOTE: This is not a actual vulnerability 
This is known problem in Nuget when using netstandart2 projects. https://github.com/dotnet/sdk/issues/42651
It could be worked around by moving to CPM and pin System.Private.Uri version, but users are discouraged to reference System.Private.Uri directly, and it's frustrating for users to get this warning from official Microsoft package

It's recommended than:
- Version of nuget `Microsoft.Web.LibraryManager.Cli` provided that target modern .net (net9 or net8)
- System.Private.Uri is pinned by `Microsoft.Web.LibraryManager.Cli`  to a non-vulnerable version

jimmylewis:
Fixes two issues in the recent changes to the targets file in the Build package:
 - The path to the build task assembly was incorrectly updated since we now build the project for net8.0 and net481.  Also, it was updated to use an MSBuild variable that exists in this repo, but the targets file runs in the context of the consumer (which wouldn't have it defined).  So the path was returned to a hard-coded value.  This will need to be maintained as target frameworks get updated (or we need a better way to generate the targets file).
 - The Build package was passing the dependency to System.Runtime.Loader transitively.  This was causing warnings with .NET 9's stricter security advisory due to the runtime dependencies (System.Private.Uri specifically).  This should not be a transitive dependency for consumers, so setting it to PrivateAssets fixes the issue.

Resolves #770

### DIFF
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.csproj
@@ -16,7 +16,7 @@
     <PackageReference Include="Microsoft.Build.Framework" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" PrivateAssets="all" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" PrivateAssets="all" />
-    <PackageReference Include="System.Runtime.Loader" Condition="'$(TargetFramework)' == '$(NetCoreTFM)'" />
+    <PackageReference Include="System.Runtime.Loader" Condition="'$(TargetFramework)' == '$(NetCoreTFM)'" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.targets
+++ b/src/LibraryManager.Build/Microsoft.Web.LibraryManager.Build.targets
@@ -2,8 +2,8 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
     <PropertyGroup>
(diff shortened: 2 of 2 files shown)


phil-allen-msft:
Is there a reasonable test to prevent recurrence?

jimmylewis:
Not at the moment.  Unit testing doesn't catch the "reality" of running within msbuild (which could be either netfx or netX.0).  We need some integration tests that actually use the package to build, and we need to test both runtime environments.  I can try to come up with a separate PR to add such a test later, I think it would be good to have (similarly, I'd like to add some "real" integration tests for the CLI exe as well, rather than the bootstrapping unit tests we do now).

I manually tested these changes by using dotnet build (for net8.0 runtime) and building within VS (which, AFAICT from the errors I was getting previously, uses the netfx msbuild).  Before the fix, each gave me different errors, and after the fix, both work fine.

phil-allen-msft (approved):

PR #778, opened 2025-01-10
Add service connection for NugetAuthenticate

phil-allen-msft:
Add service connection conditionally
Update "Install .NET SDK" action to use the contents of global.json to determine the version of the SDK to install

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -4,6 +4,9 @@ steps:
 
 - task: UseDotNet@2
   displayName: Install .NET Core SDK
+  inputs:
+    packageType: sdk
+    useGlobalJson: true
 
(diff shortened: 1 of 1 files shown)


jimmylewis:
how does this remove the need for #776?  That addresses the scenario where anyone who doesn't have permissions to push packages to the azure-public feed cannot build libman when either:
* adding/updating a package version
* after an SDK servicing event upgrades the runtime package versions.

jimmylewis (approved):
LGTM except for my other comment about the PR description.

phil-allen-msft:
I looked at the code change; the code change would seem to give an alternative source for packages, which the service connection fix (here) is at least related for.  But on closer look, it is indeed a separate scenario.  Modifying the description.

PR #790, opened 2025-03-24
Integration tests for CLI and Build task

jimmylewis:
Each new integration test project has steps to set up using the package from the current build.  This will allow more consistent testing in addition to the Visual Studio integration tests.

Currently each project just has some basic tests to get started.  We can add more scenarios later, this PR is to introduce the core infrastructure with some basic starters.

### DIFF
--- a/test/libman.IntegrationTest/CliBaseTest.cs
+++ b/test/libman.IntegrationTest/CliBaseTest.cs
@@ -1,0 +1,120 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Web.LibraryManager.Cli.IntegrationTest;
+
+[TestClass]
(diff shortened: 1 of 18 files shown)


test/libman.IntegrationTest/CliBaseTest.cs
```diff
@@ -0,0 +1,120 @@
...
+    private string _testDirectory;
+
+    [TestInitialize]
+    public async Task TestInitialize()
+    {
+        _testDirectory = Path.Combine(Path.GetTempPath(), "LibmanTest" + Guid.NewGuid().ToString());
+        Directory.CreateDirectory(_testDirectory);
+
+        // create an empty nuget.config with only our package source
+        await RunDotnetCommandLineAsync("new nuget.config");
```
phil-allen-msft, line 27: Would prefer `new nugetconfig`.  Both nuget.config and nugetconfig work as the template name, but all the docs point to nugetconfig (one word, no dot)

jimmylewis: easily done :)

phil-allen-msft (approved):

PR #796, opened 2025-07-03
Update versions to allow integration tests to run on 17.14

phil-allen-msft:
Update versions to allow integration tests to run on 17.14

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.200",
-    "rollForward": "latestFeature"
+    "version": "9.0.100",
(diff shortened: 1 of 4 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.200",
```
jimmylewis, line 3: Is updating the SDK necessary? (feels weird to move to a STS instead of LTS if there isn't a pressing need for the new SDK features)

jimmylewis (approved):

global.json
phil-allen-msft: Switched back to LTS.

PR #816, opened 2025-11-21
ComponentGovernance update to 17.14.28 versions Microsoft.Build.* packages and related fallout.

ToddGrun:
Pretty big jump from 17.11.4 to 17.14.28, so quite a bit of other packages needed pinned version updates

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.415",
+    "version": "9.0.308",
     "rollForward": "latestMajor"
   }
 }


Directory.Packages.props
```diff
@@ -10,12 +10,12 @@
 
   <ItemGroup>
     <PackageVersion Include="MessagePack" Version="2.5.187" />
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.Build.Framework" Version="17.11.4" />
-    <PackageVersion Include="Microsoft.Build.Tasks.Core" Version="17.11.4" />
-    <PackageVersion Include="Microsoft.Build.Utilities.Core" Version="17.11.4" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="9.0.0" />
```
jimmylewis: just wonder if we should jump to 10 (LTS).  Not sure if VS has taken that package version yet though.

jimmylewis (approved):

PR #819, opened 2025-11-30
Update CLI tool shim platforms to support ARM

### LINKED ISSUE #708: Add macOS Arm64 support to the CLI tool
### Is your feature request related to a problem? Please describe.
I am trying to run the `libman` CLI tool on my Mac mini M2 Pro after installing the recommended Arm64 versions of the .NET SDK, but it fails and says that it can't find a .NET Runtime.

### Describe the solution you'd like
Correct me if I'm wrong, but I think the problem could be solved if `osx-arm64` was added to the PackAsToolShimRuntimeIdentifiers tag in https://github.com/aspnet/LibraryManager/blob/b0570cd92d4c7e8c4650a3ebbb76838eec069469/src/libman/libman.csproj#L8

I saw a lot of macOS RIDs [here](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#macos-rids) and none of them were specifically "osx-arm64", but they do have "osx.13-arm64", so you'd think they'd have a non-version-specific one.

### Describe alternatives you've considered
I solved the issue momentarily by installing the x64 version of the Runtime, but I read that running that version uses an Intel emulator and I'm wondering if there's overhead there.

jimmylewis:
- Added win-arm64 and osx-arm64
- Removed win-x86

I don't think x86 is supported anymore, according to https://github.com/dotnet/sdk/blob/f578225fe1fdd93db9efac273562b3709db9e8e9/src/Cli/dotnet/ShellShim/ShellShimTemplateFinder.cs#L27

Resolves #708

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Commit: TBD
 - Adds fileMappings support for filesystem provider
 - Fixes LIB016 error when using fileMappings
 - Fixes issue where restore-on-save in one project in VS removes files restored in a separate project
+- Add ARM platform shims to CLI tool
 
 ## 3.0.71
 Commit: 33c04f70a4f55f1cddbaddad60fc78a282b298d3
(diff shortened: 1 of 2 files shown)


jimmylewis:
Caveats: I'm not entirely sure how to test this.  I don't have an osx-arm64 machine to test, and my win-arm64 machine works just fine with the win-x64 package so it's hard to validate a difference.

Also, I ran into #728 again while making this change because the runtime packages for these platforms aren't in the libman feed.

PR #822, opened 2026-02-09
Fix expanding template expressions from defaultDestination for CLI install 

### LINKED ISSUE #821: defaultDestination's [Name] [Version] expansion no longer works for LibMan CLI
This no longer works for LibMan CLI

_Originally posted by @dunnodepends in https://github.com/aspnet/LibraryManager/issues/759#issuecomment-3721782542_

jimmylewis:
Apply the same expansion when installing libraries whether the manifest is from disk or generated in memory.  (CLI updates the manifest on disk after the install succeeds.)

Some extras:
- turned off preference for var when the type is apparent.  Since we don't use var through the majority of the code, this was distracting.
- Added a workaround for running the CLI integration tests when a workaround is in place for #728.  Apparently if a parent directory's nuget.config configures packageSourceMappings, there isn't a way to turn it off in a child directory's nuget.config.

Resolves #821

### DIFF
--- a/src/LibraryManager/Utilities/PathTemplateUtility.cs
+++ b/src/LibraryManager/Utilities/PathTemplateUtility.cs
@@ -1,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+
+namespace Microsoft.Web.LibraryManager.Utilities
+{
+    /// <summary>
+    /// Utility for path template operations.
+    /// </summary>
+    public static class PathTemplateUtility
+    {
+        /// <summary>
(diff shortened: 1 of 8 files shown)


src/LibraryManager/Utilities/PathTemplateUtility.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
```
dbreshears, line 5: Looks like this can be removed

dbreshears (approved):

src/LibraryManager/Utilities/PathTemplateUtility.cs
jimmylewis: Needed for net481 (for string.Contains(char))

